PR #10, opened 2022-10-17
Uniform exec calls

iobuhov:
## This PR contains

-   [ ] Bug fix
-   [ ] Feature
-   [x] Refactor
-   [ ] Documentation
-   [ ] Other (describe)

### DIFF
--- a/packages/tools/release-utils-internal/src/mpk.ts
+++ b/packages/tools/release-utils-internal/src/mpk.ts
@@ -16,8 +16,7 @@ async function ensureMxBuildDockerImageExists(mendixVersion: Version): Promise<v
         await exec(
             `docker build -f ${dockerfilePath} ` +
                 `--build-arg MENDIX_VERSION=${version} ` +
-                `mxbuild:${version} ${process.cwd()}`,
-            { stdio: "inherit" }
+                `mxbuild:${version} ${process.cwd()}`
         );
     }
 }
@@ -51,11 +50,11 @@ export async function createModuleMpkInDocker(
         moduleName
     ].join(" ");
 
-    await exec(`docker run -v ${sourceDir}:/source --rm mxbuild:${version} bash -c "${args}"`, { stdio: "inherit" });
+    await exec(`docker run -v ${sourceDir}:/source --rm mxbuild:${version} bash -c "${args}"`);
     console.log(`Module ${moduleName} created successfully.`);
     if (process.env.CI) {
         console.info("Changing sourceDir ownership...");
-        await exec(`sudo chown -R "$(id -u):$(id -g)" ${sourceDir}`, { stdio: "inherit" });
+        await exec(`sudo chown -R "$(id -u):$(id -g)" ${sourceDir}`);
     }
 }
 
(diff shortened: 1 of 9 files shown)


packages/tools/release-utils-internal/src/steps.ts
```diff
@@ -12,18 +12,16 @@ export async function updateChangelogsAndCreatePR(
     const releaseBranchName = `${releaseTag}-update-changelog`;
 
     console.log(`Creating branch '${releaseBranchName}'...`);
-    await execShellCommand(`git checkout -b ${releaseBranchName}`);
+    await exec(`git checkout -b ${releaseBranchName}`, { stdio: "inherit" });
```
r0b1n, line 15: `{ stdio: "inherit" }` is repeated quite some times, can we reverse it, make it a default option and use additional options in places where we need output to be returned?

packages/tools/release-utils-internal/src/steps.ts
```diff
@@ -12,18 +12,16 @@ export async function updateChangelogsAndCreatePR(
...
 
     console.log(`Committing CHANGELOG.md to '${releaseBranchName}' and pushing to remote...`);
-    await execShellCommand([
-        `git add ${changelog.changelogPath}`,
-        `git commit -m "chore(${packageInfo.packageName}): update changelog"`,
-        `git push ${remoteName} ${releaseBranchName}`
-    ]);
+    await exec(`git add ${changelog.changelogPath}`);
+    await exec(`git commit -m "chore(${packageInfo.packageName}): update changelog"`);
+    await exec(`git push ${remoteName} ${releaseBranchName}`);
```
r0b1n, line 24: `{ stdio: "inherit" }` is probably missing for those (if we don't want to change it to the default option).

iobuhov: Done

packages/tools/release-utils-internal/src/steps.ts
iobuhov: Done

packages/tools/release-utils-internal/src/mpk.ts
```diff
@@ -1,25 +1,22 @@
...
+    const existingImages = (await exec(`docker image ls -q mxbuild:${version}`)).stdout.trim();
     if (!existingImages) {
         console.log(`Creating new mxbuild:${version} docker image...`);
         const dockerfilePath = join(process.cwd(), "packages/tools/pluggable-widgets-tools/scripts/mxbuild.Dockerfile");
-        await execShellCommand(
+        await exec(
             `docker build -f ${dockerfilePath} ` +
                 `--build-arg MENDIX_VERSION=${version} ` +
-                `-t mxbuild:${version} ${process.cwd()}`
+                `mxbuild:${version} ${process.cwd()}`
```
r0b1n, line 19: `-t` (tag) is gone, is it intentional?

iobuhov: This is intentional, we don't need terminal to build image

r0b1n: This is true for `-t` in `docker run` commands. For `docker build` command this means [tag an image](https://docs.docker.com/engine/reference/commandline/build/#tag-an-image--t) 🙈

iobuhov: No waaaay 🤨 😕

iobuhov: Done

PR #11, opened 2022-10-17
[WC-1357]: rename HTML Node to HTML Element

r0b1n:
#### Feature specific

-   Comply with designs ✅ 
-   Comply with PM's requirements ✅ 

## This PR contains

-   [ ] Bug fix
-   [x] Feature
-   [ ] Refactor
-   [ ] Documentation
-   [ ] Other (describe)

## What is the purpose of this PR?

HTML Element is a better name for the widget.


packages/pluggableWidgets/html-element-web/src/HTMLElement.editorConfig.ts
```diff
@@ -25,7 +25,7 @@ const disabledElements = ["script"];
 
 function isValidHtmlTagName(name: string): boolean {
     // rather naive approach that works in most cases
-    return /^[a-z_][\w.-]*$/i.test(name);
+    return /^[a-z][\w.-]*$/i.test(name);
```
iobuhov, line 28: @r0b1n , are you sure we allow `-`?

iobuhov (commented):
Are you sure `html-element-root` is fine class for widget?
AFAIK we trying to use `widget-<widget-name>` pattern

packages/pluggableWidgets/html-element-web/src/HTMLElement.editorConfig.ts
r0b1n: Yes, while no native html tags use `-`, we want to support web components and they are named with 2 or more words separated with `-` (most of the time at least).

r0b1n:
`html-element-root` is only in test files/snapshots. But good that you pointed this out. The actual class name was `html-element-widget` which is still off. I changed it to `widget-html-element`. Thanks!

packages/pluggableWidgets/html-element-web/src/HTMLElement.editorConfig.ts
iobuhov: Ah, okay, I got it

iobuhov (approved):

PR #21, opened 2022-10-19
Automated release for all package types

iobuhov:
### Changes

- Uniform `rui-create-gh-release` script
- Add new release steps
- Update turbo.json
- Allow to write notes on `update-changelog` to `RELEASE_NOTES_FILE`
- Fix docker file path


packages/tools/release-utils-internal/bin/rui-update-changelog-module.ts
```diff
@@ -58,8 +59,14 @@ async function main(): Promise<void> {
...
             })
         )
-        .moveUnreleasedToVersion(info.version)
-        .save();
+        .moveUnreleasedToVersion(info.version);
+
+    updated.save();
+
+    if (process.env.RELEASE_NOTES_FILE) {
+        const content = updated.getLatestReleaseContent({ header: false });
```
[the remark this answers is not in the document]
iobuhov: Yep

packages/tools/release-utils-internal/bin/rui-update-changelog-module.ts
```diff
@@ -58,8 +59,14 @@ async function main(): Promise<void> {
...
+        await writeFile(process.env.RELEASE_NOTES_FILE, content);
```
r0b1n, line 68: I see usage of `echo().to()` in other place, maybe here we could use it as well.

iobuhov: W would rather remove echo().to() as it too verbose... It does what it does, it echo stuff to stdout and then write it to the file. I think it's better if we stick with `writeFile`

r0b1n: 👍

r0b1n (approved):

PR #27, opened 2022-10-20
[WC-1294] Charts v4

iobuhov:
## This PR contains

-   [ ] Bug fix
-   [x] Feature
-   [ ] Refactor
-   [ ] Documentation
-   [ ] Other (describe)

## What is the purpose of this PR?

- Create new version of charts free of legacy code

## Relevant changes

- `charts` package become `charts-web`
- Add build script for charts

## What should be covered while testing?

- New charts mpk


packages/pluggableWidgets/charts-web/src/package.xml
```diff
@@ -0,0 +1,17 @@
...
+            <widgetFile path="AreaChart/AreaChart.xml" />
+            <widgetFile path="BarChart/BarChart.xml" />
+            <widgetFile path="BubbleChart/BubbleChart.xml" />
+            <widgetFile path="ColumnChart/ColumnChart.xml" />
+            <widgetFile path="HeatMap/HeatMap.xml" />
+            <widgetFile path="LineChart/LineChart.xml" />
+            <widgetFile path="PieChart/PieChart.xml" />
+            <widgetFile path="TimeSeries/TimeSeries.xml" />
+        </widgetFiles>
+        <!-- Please leave files empty, file entries will be generated as part of build -->
```
r0b1n, line 14: Wouldn't it be easier to generate the whole file as text completely, together with `widgetFiles` entry. This way we probably don't need to mingle with XML at all

iobuhov: Yeah, good idea, I also thought about but was not sure, I'll add!

r0b1n: Another option is to fill this section manually and also don't do xml manipulations

iobuhov: Also possible... and probably simpler... I will remove generation step

PR #60, opened 2022-11-14
[WC-1415]: Update momentjs in calendar web

iobuhov:
### Description

This PR does two things:

- Fix `build` and `release` scripts for calendar-web
- Prepare `calendar-web` v1.0.14 release

But, please, keep in mind that this package is **not a part of our pnpm workspace yet**. That means automation and other cool features not work with this package. Also, it means we have to manually release this widget. This widget is candidate for pluggable conversion, so automation changes will be part of this migration.

### Pull request type

-   [x] No code changes (changes to documentation, CI, metadata, etc)

- Probably `build` and `release` commands


HedwigAR (commented):
Hey, looks like there was a bit more changed than a library update. I see the package json was changed due to a pluggable widget tools update so I get that. Why was the build.ts added?

iobuhov:
> Why was the build.ts added?

As you may noticed, `calendar-web` is located in `packages/customWidgets` and this package is still use `webpack` under the hood. `build.ts` was added because I simplified widget build and removed dependency on `gulp`.

HedwigAR (dismissed):
Alright! I didn't actually notice it was in customWidgets, thanks for the explanation!

leonardomendix (approved):

HedwigAR (approved):

PR #162, opened 2022-11-18
ci(build): keep turbo cache only on main

iobuhov:
### Description

This PR change how we store trubo cache on GH. Now cache should be saved only on `main` and updated when `main` SHA changes. That means we not storing cache for pull request branches anymore.

### Pull request type

-   [x] No code changes (changes to documentation, CI, metadata, etc)

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -49,13 +49,15 @@ jobs:
         with:
           node-version-file: ".nvmrc"
           cache: "pnpm"
+      - name: Get sha of main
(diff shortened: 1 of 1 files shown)


iobuhov:
> Why only for the main?

Because `main` is main source of "truth". Cache stored on `main` can be reused in pull requests and other branches. That means that we still get benefit from reusing cache, but from other side we don't pollute storage with cache that right now created for every branch/PR

leonardomendix (approved):

rahmanunver (approved):

PR #182, opened 2022-12-08
[WC-1537] Format dependabot config

iobuhov:
### Description

Apply formatting for dependabot config

### Pull request type

### DIFF
--- a/automation/scripts/package.json
+++ b/automation/scripts/package.json
@@ -13,9 +13,15 @@
     "url": "https://github.com/mendix/web-widgets.git"
   },
   "scripts": {
+    "lint": "eslint . --ext .mjs",
+    "format": "prettier --write .",
     "update-dependabot-config": "node generate-dependabot-config.mjs"
   },
   "dependencies": {
     "yaml": "^2.1.3"
+  },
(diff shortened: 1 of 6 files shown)


automation/scripts/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@mendix/web-widgets-scripts",
+  "name": "@web-widgets/scripts",
```
r0b1n, line 2: Maybe it is better to keep things at `@mendix`?

iobuhov: I agree, but in my opinion it's better to use scope other then `@mendix` as I feel `@mendix` is now reserved for public tools and in general probably will be used for **public** npm packages. I would like to stay away from this scope. If you don't like `@web-widgets` we can choose any other.

r0b1n: `@mendix-web-widgets` maybe?

iobuhov: @r0b1n , done

iobuhov:
cc: @leonardomendix

r0b1n (approved):

PR #185, opened 2022-12-09
[WC-1561]: Update style regex to handle zero spaces

iobuhov:
### Description

Small change in regex that we use to split CSS rules. Now we allow zero spaces in CSS props. (e.g. "width:10px")

### Pull request type

-   [x] Bug fix (non-breaking change which fixes an issue)

- Just create any html element widget and add style attribute with zero space properties.

### DIFF
--- a/packages/pluggableWidgets/html-element-web/src/utils/style-utils.ts
+++ b/packages/pluggableWidgets/html-element-web/src/utils/style-utils.ts
@@ -1,8 +1,8 @@
 import React from "react";
 
 // We need regexp to split rows in prop/value pairs
-// split by : not work for all cases, eg "background-image: url(http://localhost:8080);"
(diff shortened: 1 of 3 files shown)


packages/pluggableWidgets/html-element-web/src/utils/style-utils.ts
```diff
@@ -2,7 +2,7 @@ import React from "react";
 
 // We need regexp to split rows in prop/value pairs
 // split by : not work for all cases, eg "background-image: url(http://localhost:8080);"
-const cssPropRegex = /(?<prop>[^:]+):\s+?(?<value>[^;]+);?/m;
+const cssPropRegex = /(?<prop>[^:]+):\s*(?<value>[^;]+);?/m;
```
r0b1n, line 5: Do we need unit tests for this scenario?

rahmanunver (approved):

leonardomendix (approved):

PR #263, opened 2023-01-17
[WC-1608]: add google tag web widget

r0b1n:
### Pull request checklist

-   [x] All new and existing tests passed
-   [x] I run `lint` command locally and it doesn’t give errors
-   [x] PR title properly formatted `[XX-000]: description`
-   [x] Added record to packages' CHANGELOG.md
-   [x] Bumped package version in `package.json` and `package.xml`
-   [ ] Added a link to related project PRs (atlas, pluggable-widgets-tools, testProject, etc.) (optional)
-   [ ] Created docs PR to [mendix/docs](https://github.com/mendix/docs.git) and added a link (optional)

### Pull request type

-   [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/packages/pluggableWidgets/google-tag-web/package.json
+++ b/packages/pluggableWidgets/google-tag-web/package.json
@@ -19,7 +19,7 @@
   },
   "marketplace": {
     "minimumMXVersion": "8.18.0",
-    "appName": "Google Tag Widget",
+    "appName": "Google Tag",
     "appNumber": -1
   },
   "testProject": {
--- a/packages/pluggableWidgets/google-tag-web/package.json
+++ b/packages/pluggableWidgets/google-tag-web/package.json
@@ -41,16 +41,17 @@
     "publish-marketplace": "rui-publish-marketplace"
   },
   "dependencies": {
-    "@mendix/pluggable-widgets-commons": "workspace:*"
+    "@mendix/pluggable-widgets-commons": "workspace:*",
+    "@types/node": "^18.8.3"
   },
   "devDependencies": {
-    "@mendix/pluggable-widgets-tools": "^9.18.0",
     "@mendix-internal/automation-utils": "workspace:*",
-    "eslint": "^7.20.0",
+    "@mendix/pluggable-widgets-tools": "^9.18.0",
     "@types/big.js": "^6.0.2",
     "@types/jest": "^26.0.24",
     "@types/react": "~17.0.1",
     "@types/react-dom": "~17.0.1",
+    "eslint": "^7.20.0",
     "ts-node": "^9.0.0",
     "typescript": "4.5.4"
   }
--- a/packages/pluggableWidgets/google-tag-web/package.json
+++ b/packages/pluggableWidgets/google-tag-web/package.json
@@ -4,6 +4,7 @@
   "version": "1.0.0",
   "description": "gtag.js integration widget",
   "copyright": "© Mendix Technology BV 2022. All rights reserved.",
+  "private": true,
   "license": "Apache-2.0",
   "repository": {
     "type": "git",
@@ -47,7 +48,10 @@
     "@mendix-internal/automation-utils": "workspace:*",
     "eslint": "^7.20.0",
     "@types/big.js": "^6.0.2",
+    "@types/jest": "^26.0.24",
     "@types/react": "~17.0.1",
-    "@types/react-dom": "~17.0.1"
+    "@types/react-dom": "~17.0.1",
+    "ts-node": "^9.0.0",
+    "typescript": "4.5.4"
   }
 }
--- a/packages/pluggableWidgets/google-tag-web/package.json
+++ b/packages/pluggableWidgets/google-tag-web/package.json
@@ -1,0 +1,53 @@
+{
+  "name": "google-tag-web",
+  "widgetName": "GoogleTag",
+  "version": "1.0.0",
+  "description": "gtag.js integration widget",
+  "copyright": "© Mendix Technology BV 2022. All rights reserved.",
+  "license": "Apache-2.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/mendix/web-widgets.git"
+  },
+  "config": {
+    "mendixHost": "http://localhost:8080",
+    "developmentPort": 3000
+  },
+  "mxpackage": {
+    "name": "GoogleTag",
+    "type": "widget"
+  },
+  "marketplace": {
+    "minimumMXVersion": "8.18.0",
+    "appName": "Google Tag Widget",
+    "appNumber": -1
+  },
+  "testProject": {
+    "githubUrl": "https://github.com/mendix/testProjects",
+    "branchName": "google-tag-web"
+  },
+  "packagePath": "com.mendix.widget.web",
+  "scripts": {
+    "start": "pluggable-widgets-tools start:server",
+    "dev": "pluggable-widgets-tools start:web",
+    "build": "pluggable-widgets-tools build:web",
+    "prerelease": "npm run lint",
+    "release": "pluggable-widgets-tools release:web",
+    "lint": "eslint --config ../../../.eslintrc.js --ext .jsx,.js,.ts,.tsx src/",
+    "verify": "rui-verify-package-format",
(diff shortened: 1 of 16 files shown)


packages/pluggableWidgets/google-tag-web/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "google-tag-web",
+  "widgetName": "GoogleTag",
+  "version": "1.0.0",
+  "description": "gtag.js integration widget",
+  "copyright": "© Mendix Technology BV 2022. All rights reserved.",
```
iobuhov, line 6: We should open one big PR where we update `2022` -> `2023`

packages/pluggableWidgets/google-tag-web/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "mendixHost": "http://localhost:8080",
+    "developmentPort": 3000
+  },
+  "mxpackage": {
+    "name": "GoogleTag",
+    "type": "widget"
+  },
+  "marketplace": {
+    "minimumMXVersion": "8.18.0",
+    "appName": "Google Tag Widget",
```
iobuhov, line 22: We probably need to ask Danny, as I never saw widgets, that have "Widget" in the title

packages/pluggableWidgets/google-tag-web/src/GoogleTag.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<widget id="com.mendix.widget.web.googletag.GoogleTag" pluginWidget="true" needsEntityContext="true" offlineCapable="true" supportedPlatform="Web" xmlns="http://www.mendix.com/widget/1.0/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.mendix.com/widget/1.0/ ../node_modules/mendix/custom_widget.xsd">
+    <name>Google Tag</name>
+    <description>gtag.js integration widget</description>
+    <icon />
+    <properties>
+        <propertyGroup caption="General">
+            <property key="command" type="enumeration" required="true" defaultValue="config">
+                <caption>Command</caption>
+                <description />
```
iobuhov, line 10: Can we add short description? Maybe ask Danny? Or this will be part of feedback?

packages/pluggableWidgets/google-tag-web/src/commonGtag.ts
```diff
@@ -0,0 +1,39 @@
+interface DocumentMxGtag extends Document {
```
iobuhov, line 1: I think it's better to use `declare global`
https://github.com/mendix/web-widgets/search?q=declare+global

Then we probably don't need use `(document as DocumentMxTag)`

packages/pluggableWidgets/google-tag-web/package.json
iobuhov, line 6: We should probably open one big PR to update all 2022s to 2023s

iobuhov:
@r0b1n, you need to add `typescript@4.5.4` and `ts-node@^9.0.0` as dev dependency

iobuhov:
Also, we should add `private: true` in package.json as this package will be not publish on it's own, but as part of module.

iobuhov (approved):

packages/pluggableWidgets/google-tag-web/src/utils.ts
```diff
@@ -0,0 +1,97 @@
...
+    useEffect(() => {
+        const handle = window.dojo.connect(window.mx.ui.getContentForm(), "onNavigation", () => {
+            cb();
+        });
+
+        return () => {
+            window.dojo.disconnect(handle);
+        };
+    }, [cb]);
+}
```
iobuhov, line 84: Should we somehow handle this for coming "modern client"? I think this may not work in Mendix 10

packages/pluggableWidgets/google-tag-web/src/utils.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export function useOnAfterRenderExecution(executeImmediately: boolean, cb: () => void): () => void {
+    const [needsExecution, setNeedsExecution] = useState<boolean>(executeImmediately);
```
iobuhov, line 87: If we just want to call callback, why we need `useState`? Maybe it's better to use `useRef(executeImmediately)`?

packages/pluggableWidgets/google-tag-web/src/GoogleTag.tsx
```diff
@@ -0,0 +1,88 @@
...
+export default function GoogleTag(props: GoogleTagContainerProps): ReactElement | null {
+    if (props.widgetMode === "basic") {
+        return <GoogleTagBasicPageView {...props} />;
+    } else {
+        return <GoogleTagAdvancedMode {...props} />;
+    }
+}
+
+function GoogleTagBasicPageView(props: GoogleTagContainerProps): ReactElement | null {
+    const runCommands = useOnAfterRenderExecution(false, () => {
```
iobuhov, line 21: Do we really need `useOnAfterRenderExecution` here? What the purpose? I think we can remove it and just pass `cb` to `useDojoOnNavigation`.

packages/pluggableWidgets/google-tag-web/src/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+            const handle = window.dojo.connect(window.mx.ui.getContentForm(), "onNavigation", cb);
+
+            return () => {
+                window.dojo.disconnect(handle);
+            };
+        }, [cb]);
+    }
+}
+
+export function useOnAfterRenderExecution(cb: () => boolean): () => void {
```
iobuhov, line 97: I still can't get why we need `useOnAfterRenderExecution`, @r0b1n , can you explain it a bit?

r0b1n: It took care of re-execution when props are not ready. This is probably not the best custom hook out there, I got rid of it :)

packages/pluggableWidgets/google-tag-web/src/GoogleTag.tsx
```diff
@@ -0,0 +1,108 @@
...
+        needsExecution.current = false;
+    };
+
+    if (props.trackPageChanges) {
+        // eslint-disable-next-line react-hooks/rules-of-hooks
+        useDojoOnNavigation(() => {
+            needsExecution.current = true;
+            runCommand();
+        });
+    }
```
iobuhov, line 103: let's use `useRef` for `props.trackPageChanges` and remove if statement. Then we can check it inside callback

packages/pluggableWidgets/google-tag-web/src/utils.ts
```diff
@@ -0,0 +1,93 @@
...
+        // eslint-disable-next-line react-hooks/rules-of-hooks
+        useEffect(() => {
+            const handle = window.dojo.connect(window.mx.ui.getContentForm(), "onNavigation", cb);
+
+            return () => {
+                window.dojo.disconnect(handle);
+            };
+        }, [cb]);
+    }
+}
```
iobuhov, line 93: Same, let's move `window.dojo` check inside `useEffect` callback

iobuhov (changes requested):

packages/pluggableWidgets/google-tag-web/src/GoogleTag.tsx
r0b1n: I don't see added value in the proposed change. `props.trackPageChanges` is static and never changes during widget's lifetime, there shouldn't be a need to wrap it in `useRef`. For the same reason it is safe to do conditional call and skip unnecessary call of the hook.

packages/pluggableWidgets/google-tag-web/src/utils.ts
r0b1n: Also here. Dojo is either there or not, so the conditional call is always stable.

iobuhov (approved):

iobuhov (approved):

leonardomendix (approved):

PR #292, opened 2023-01-20
[WC-1615] Accordion:  new "Load content" group property

iobuhov:
### Description

- Add "Load content" prop to the group properties.

### Pull request type

-   [x] New feature (non-breaking change which adds functionality)

- Group rendering when "Load content" = always
- Group rendering when "Load content" = when expanded
- xas requests from group widgets


packages/pluggableWidgets/accordion-web/src/components/MountOnceReady.tsx
```diff
@@ -0,0 +1,14 @@
+import { createElement, ReactNode, useRef, Fragment } from "react";
+
+export function MountOnceReady(props: { ready: boolean; children: ReactNode }): JSX.Element | null {
+    const { ready, children } = props;
+    const isReady = useRef(false);
+
+    isReady.current = isReady.current || ready;
```
tamelbrun, line 7: 🎄 `isReady.current ||= ready;`

r0b1n (approved):

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
```diff
@@ -179,7 +189,11 @@ export function AccordionGroup(props: AccordionGroupProps): ReactElement | null
                 aria-labelledby={`${props.id}HeaderButton`}
             >
                 <div ref={contentRef} className={"widget-accordion-group-content"}>
-                    {props.content}
+                    {lazyRender ? (
+                        <MountOnceReady ready={!renderCollapsed}>{props.content}</MountOnceReady>
```
tamelbrun, line 193: Wouldn't it be easier to simply have a `const renderContent = useRef(false);` here which is `||=`-ed with `!lazyRender || !renderCollapsed`, so you can simply render `renderContent && props.content` instead of a whole separate component?

tamelbrun (changes requested):

packages/pluggableWidgets/accordion-web/src/components/MountOnceReady.tsx
iobuhov: Done 🪄

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
iobuhov: Done 🪄

iobuhov:
@tamelbrun, please have look one more time 🙏

rahmanunver (dismissed):

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
```diff
@@ -1,15 +1,15 @@
...
 import "../ui/accordion-main.scss";
 
+/* eslint-disable no-unused-vars */
 export const enum Target {
     FIRST = "first",
     LAST = "last",
     PREVIOUS = "previous",
     NEXT = "next"
 }
+/* eslint-enable no-unused-vars */
```
tamelbrun, line 12: Why do we need to suppress ESLint warnings here?

packages/pluggableWidgets/accordion-web/src/Accordion.editorPreview.tsx
```diff
@@ -58,6 +58,7 @@ export function preview(props: AccordionPreviewProps): ReactElement {
     const expandIcon = mapPreviewIconToWebIcon(props.expandIcon);
     const collapseIcon = mapPreviewIconToWebIcon(props.collapseIcon);
 
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
tamelbrun, line 61: Why do we need to suppress a warning here?

tamelbrun: On a side note, it would seem that `mapPreviewIconToWebIcon` returns either an object or `undefined`, making the three cases of `?? undefined` below a no-op. If you're touching this code anyway, feel free to clean that up :).

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
```diff
@@ -39,6 +47,10 @@ export function AccordionGroup(props: AccordionGroupProps): ReactElement | null
     const rootRef = useRef<HTMLDivElement>(null);
     const contentWrapperRef = useRef<HTMLDivElement>(null);
     const contentRef = useRef<HTMLDivElement>(null);
+    const lazyRender = loadContent === "whenExpanded";
+    const renderContent = useRef(false);
+
+    renderContent.current ||= !lazyRender || !renderCollapsed;
```
tamelbrun, line 53: I think we can do even better: initialize the ref with `loadContent === "always"` and only do `||= !renderCollapsed`. After all, the `loadContent` prop is static.

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
```diff
@@ -27,10 +27,18 @@ export interface AccordionGroupProps {
     animateContent?: boolean;
     generateHeaderIcon?: (collapsed: boolean) => ReactElement;
     showHeaderIcon?: "right" | "left" | "no";
+    loadContent?: "always" | "whenExpanded";
```
tamelbrun, line 30: Why is this prop optional? The prop of the widget isn't (and uses the generated `LoadContentEnum` type).

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
```diff
@@ -179,7 +191,7 @@ export function AccordionGroup(props: AccordionGroupProps): ReactElement | null
                 aria-labelledby={`${props.id}HeaderButton`}
             >
                 <div ref={contentRef} className={"widget-accordion-group-content"}>
-                    {props.content}
+                    {renderContent.current ? props.content : null}
```
tamelbrun, line 194: Why not simply `renderContent.current && props.content`? This is the recommended syntax for conditional rendering in React if the left side is a boolean.

tamelbrun (changes requested):

tamelbrun:
> P.S. Don't forget to resolve conversations, otherwise I'll never merge it 😂

I'd love to, but I can't :P. According to GitHub's logic, the creator of the pull request or someone with write access to the repository can resolve conversations. And I happen to be neither :).

iobuhov:
> > P.S. Don't forget to resolve conversations, otherwise I'll never merge it 😂
> 
> I'd love to, but I can't :P. According to GitHub's logic, the creator of the pull request or someone with write access to the repository can resolve conversations. And I happen to be neither :).

Don't you have resolve button 🤨 ???
(I'm asking as probably we need to adjust our repo settings)
Screen Shot 2023-01-26 at 17 11 16

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
iobuhov: This is because `eslint` and `typescript-eslint` rules clash. This issue is fixed in #305.
In short, eslint think this is `javascript` and it thinks it's unsed var, for `typescript-eslint` all good

packages/pluggableWidgets/accordion-web/src/Accordion.editorPreview.tsx
iobuhov: > Why do we need to suppress a warning here?

Because, we use `hook` in `preview` function (which from eslint point of view is not a component, which is kinda correct, but in our case this function is a functional component` We can't change function name, otherwise export will be broken

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
iobuhov: Kinda, inheritage from Chak, we were trying to separate components and `ConainerProps`. But, there not a lot of benefit from doing that. I'll do the change

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
iobuhov: This my habit from old react days, kinda "best practice". This approach is better then  `&&` in that that it handles `0`. If you say `x` ether `0` or `1`, and you do `{x && "hello"}` user will see `0` on the screen when `x=0`. But in this case we have only `true|false` and react will take care of it. I'll make a change.

iobuhov:
@tamelbrun, added you as a maintainer 😉 , now you should be able to make resolutions 👍

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
tamelbrun: Alright, just be sure to remove it again when it's no longer necessary ;).

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
tamelbrun: Then the `loadContent = "always"` fallback in the destructuring of the props can also be removed :).

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
tamelbrun: I would not hesitate to advocate for corporal punishment of anyone who would intentionally use this construct with numbers instead of booleans ;).
✨

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
tamelbrun: If this was a generic React component, then yes. But the props of a pluggable widget are defined by us, and we _know_ that `loadContent` is a static prop that can only be changed by reconfiguring the widget in Studio Pro. Treating it as "any prop can change" would give the impression that it actualy _could_ change, which is false. But now it's totally fine ✨

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
iobuhov: Done 🪄

packages/pluggableWidgets/accordion-web/src/Accordion.editorPreview.tsx
tamelbrun: Hmm. Maybe you could use a `PreviewComponent` function internally to make it clear that it actually is a functional component. (I don't know if you can safely substitute `export const preview = PreviewComponent;` for `export function preview(...)`, otherwise you could just `return PreviewComponent(props)`.)

iobuhov: Yes, it will solve the issue, and I think it's more clean way, but then we should fix it in every widget... which is far away from scope of this PR

iobuhov: Side note, I think once I've tried to do `export const preview = PreviewComponent;`, but AFAIR, it didn't helped, it looks like Studio Pro expects to see `function preview` as statement.

tamelbrun: Yeah, I wasn't sure how strict Studio Pro would be. In any case, I'm already happy if we can avoid the ESLint warning :).

tamelbrun:
@iobuhov One last remark: it looks like there is no test coverage for `loadContent: "whenExpanded"`, all the existing tests were updated to use `"always"`. I would not recommend that ;).

packages/pluggableWidgets/accordion-web/src/components/__tests__/AccordionGroup.spec.tsx
```diff
@@ -295,4 +295,62 @@ describe("AccordionGroup", () => {
...
+                collapsible
+                content={<div>Widgets</div>}
+                loadContent={"whenExpanded"}
+                collapsed={false}
+            />
+        );
+        expect(accordionGroup.find({ children: "Widgets" })).toHaveLength(1);
+    });
+
+    it("keep widgets rendered when 'loadContent' is 'whenExpanded' and 'collapsed' prop became false", () => {
```
tamelbrun, line 338: I think you mean `and 'collapsed' prop becomes true`?

tamelbrun: ✨

tamelbrun (approved):

PR #301, opened 2023-01-23
chore: fix and cleanup package.json files

iobuhov:
### Description

This PR is about removing all "global" deps from root package and moving them to appropriate packages to make sure package includes it's dependencies explicitly.

NOTE: Also in this PR I added explicitly `jest`, `prettier`, `rollup` `ts-node` and `typescript` as devDependency for each widget that use automation scripts and `@mendix/pluggable-widgets-tools`. This is needed because of [pnpm isolation](https://pnpm.io/motivation#creating-a-non-flat-node_modules-directory) which is good way to prevent messing up with dependencies. So, reason why we need to add all above as dev dependency is that when they ship as dependency of `@mendix/pluggable-widgets-tools`, the binaries not installed in consumer `node_modules/.bin`. PWT has some trick for fixing that, but it seems changing `PATH` only works with `npm` and `lerna`. Well, I think adding packages explicitly it's good thing as now we more clean on which deps we are using in our widgets.

### Pull request type

-   [x] Dependency changes (any modification to dependencies in `package.json`)


leonardomendix (approved):

leonardomendix:
👏

leonardomendix (approved):

rahmanunver (approved):

HedwigAR (approved):

PR #381, opened 2023-02-20
[WC-1636] Gallery selection API

iobuhov:
### **Pull request type**

---

### **Description**
Add item selection implementation to Gallery based on new "selection" property from Pluggable Widgets API.

E2E updates:

Test for sorting actually was testing enum filtering. To fix that I added new test case and moved enum filtering to appropriate test group.

### DIFF
--- a/packages/pluggableWidgets/gallery-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/gallery-web/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Added
+
+-   With new "Selection" property it is now possible to make items of the gallery selectable. Additionally, it is possible to configure "On selection change" action that runs when the selection changes.
+
 ## [1.2.0] - 2022-06-13
 
 ### Fixed
--- a/packages/pluggableWidgets/gallery-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/gallery-web/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
--   With new "Selection" property it is now possible to make items of the gallery selectable. Additionally, it is possible to configure "On selection change" action that runs when the selection changes.
+-   With the new "Selection" property, it is now possible to make items in the gallery selectable. Additionally, it is possible to configure an "On selection change" action that runs when the selection changes.
 
 ## [1.2.0] - 2022-06-13
 
--- a/packages/pluggableWidgets/gallery-web/src/Gallery.tsx
+++ b/packages/pluggableWidgets/gallery-web/src/Gallery.tsx
@@ -12,7 +12,7 @@ import {
 import { FilterCondition } from "mendix/filters";
 import { extractFilters } from "./utils/filters";
 import { and } from "mendix/filters/builders";
-import { executeAction } from "@mendix/pluggable-widgets-commons";
+import { executeAction, useSelectionHelper } from "@mendix/pluggable-widgets-commons";
 
 export function Gallery(props: GalleryContainerProps): ReactElement {
     const viewStateFilters = useRef<FilterCondition | undefined>(undefined);
@@ -103,6 +103,8 @@ export function Gallery(props: GalleryContainerProps): ReactElement {
         [props.datasource, props.pageSize, isInfiniteLoad, currentPage]
     );
 
+    const selection = useSelectionHelper(props.itemSelection, props.datasource, props.onSelectionChange);
+
     return (
         <GalleryComponent
             className={props.class}
@@ -163,11 +165,24 @@ export function Gallery(props: GalleryContainerProps): ReactElement {
             itemRenderer={useCallback(
                 (renderWrapper, item) =>
                     renderWrapper(
+                        !!selection?.isSelected(item),
                         props.content?.get(item),
                         props.itemClass?.get(item)?.value,
-                        props.onClick ? () => executeAction(props.onClick?.get(item)) : undefined
+                        (props.onClick || selection) &&
+                            (() => {
+                                if (props.onClick) {
+                                    executeAction(props.onClick?.get(item));
+                                }
+                                if (selection) {
+                                    if (selection.isSelected(item)) {
+                                        selection.remove(item);
+                                    } else {
+                                        selection.add(item);
+                                    }
+                                }
+                            })
                     ),
-                [props.content, props.itemClass, props.onClick]
+                [props.content, props.itemClass, props.onClick, selection]
             )}
             numberOfItems={props.datasource.totalCount}
             page={currentPage}
--- a/packages/pluggableWidgets/gallery-web/typings/GalleryProps.d.ts
+++ b/packages/pluggableWidgets/gallery-web/typings/GalleryProps.d.ts
@@ -4,7 +4,7 @@
  * @author Mendix Widgets Framework Team
  */
 import { ComponentType, CSSProperties, ReactNode } from "react";
-import { DynamicValue, ListValue, ListActionValue, ListAttributeValue, ListExpressionValue, ListWidgetValue } from "mendix";
+import { ActionValue, DynamicValue, ListValue, ListActionValue, ListAttributeValue, ListExpressionValue, ListWidgetValue, SelectionSingleValue, SelectionMultiValue } from "mendix";
 import { Big } from "big.js";
 
 export type PaginationEnum = "buttons" | "virtualScrolling";
@@ -38,6 +38,7 @@ export interface GalleryContainerProps {
     tabIndex?: number;
     advanced: boolean;
     datasource: ListValue;
+    itemSelection?: SelectionSingleValue | SelectionMultiValue;
     content?: ListWidgetValue;
     desktopItems: number;
     tabletItems: number;
@@ -49,6 +50,7 @@ export interface GalleryContainerProps {
     emptyPlaceholder?: ReactNode;
     itemClass?: ListExpressionValue<string>;
     onClick?: ListActionValue;
+    onSelectionChange?: ActionValue;
     filterList: FilterListType[];
     filtersPlaceholder?: ReactNode;
     sortList: SortListType[];
@@ -67,6 +69,7 @@ export interface GalleryPreviewProps {
     readOnly: boolean;
     advanced: boolean;
     datasource: {} | { type: string } | null;
+    itemSelection: "None" | "Single" | "Multi";
     content: { widgetCount: number; renderer: ComponentType<{ caption?: string }> };
     desktopItems: number | null;
     tabletItems: number | null;
@@ -78,6 +81,7 @@ export interface GalleryPreviewProps {
     emptyPlaceholder: { widgetCount: number; renderer: ComponentType<{ caption?: string }> };
     itemClass: string;
     onClick: {} | null;
+    onSelectionChange: {} | null;
     filterList: FilterListPreviewType[];
     filtersPlaceholder: { widgetCount: number; renderer: ComponentType<{ caption?: string }> };
     sortList: SortListPreviewType[];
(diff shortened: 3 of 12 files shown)


packages/shared/pluggable-widgets-commons/src/selection/SelectionWrapper.ts
```diff
@@ -0,0 +1,112 @@
...
+    type: SelectionType;
+    selectionStatus: SelectionStatus;
+    isSelected(value: ObjectItem): boolean;
+    add(value: ObjectItem): void;
+    remove(value: ObjectItem): void;
+    selectAll(): void;
+    selectNone(): void;
+}
+
+export class SelectionWrapperImpl implements SelectionWrapper {
```
rahmanunver, line 18: `Impl` looks a bit ambiguous, could it have a more clear naming maybe?

rahmanunver (approved):

rahmanunver (dismissed):

packages/shared/pluggable-widgets-commons/src/selection/SelectionWrapper.ts
iobuhov: `Impl` is kinda "common" pattern in typed languages.
https://github.com/microsoft/TypeScript/search?q=path%3A%2Fsrc%2F+Impl

iobuhov: I'll keep it as I don't think we can find better name

packages/pluggableWidgets/gallery-web/src/Gallery.tsx
```diff
@@ -103,81 +104,105 @@ export function Gallery(props: GalleryContainerProps): ReactElement {
         [props.datasource, props.pageSize, isInfiniteLoad, currentPage]
     );
 
+    const { Provider: SelectionAPIProvider } = getGlobalSelectionContext();
+    const api = useCreateSelectionAPI(props.itemSelection, props.datasource, props.onSelectionChange);
+    const selection = api.selection;
```
tamelbrun, line 109: This whole "API" seems needlessly complex, especially within the context of this story where no communication with the headers is even necessary yet. I can imagine the extraction of a few helper methods to share code between Gallery and Data grid2, but there is no need to expose all of this through the global state. The only thing the header will need is "Select all/none" and whether all/some/none are selected, but none of that appears to be part of this story anyway.

packages/shared/pluggable-widgets-commons/src/selection/SelectionWrapper.ts
```diff
@@ -0,0 +1,112 @@
...
+        this.list = datasource;
+        this.onChange = onChange;
+    }
+
+    get type(): SelectionType {
+        return this.property.type;
+    }
+
+    get selectionStatus(): SelectionStatus {
+        if (this.property.type === "Single") {
```
tamelbrun, line 34: This single vs. multi check is peppered throughout this wrapper. If you're going put a convenience wrapper around the selection anyway, it would make more sense to create separate implementations for single and multi with an overlapping API for the part supported by both. That way, the code _has_ to check for single/multi before it can access the multi-only parts. Also, you probably don't need a class for this, just interfaces/types (and then you don't need an ugly `Impl`-name either. :))

packages/shared/pluggable-widgets-commons/src/selection/SelectionWrapper.ts
```diff
@@ -0,0 +1,112 @@
...
+            throw new Error("Status not available in single selection mode.");
+        }
+        if (this.property.selection.length === 0) {
+            return "none";
+        }
+        const selectedIds = new Set(this.property.selection.map(obj => obj.id));
+        const isAll = (this.list.items ?? []).every(obj => selectedIds.has(obj.id));
+        return isAll ? "all" : "some";
```
tamelbrun, line 42: Selection property is restricted to the items in the data source and does not allow duplicates, so a simple length check would be sufficient here. (But I'm not sure that this logic belongs here.)

packages/shared/pluggable-widgets-commons/src/selection/SelectionWrapper.ts
```diff
@@ -0,0 +1,112 @@
...
+    private setSelection(value: ObjectItem | undefined | ObjectItem[]): void {
+        if (Array.isArray(value)) {
+            if (this.property.type === "Multi") {
+                this.property.setSelection(value);
+            }
+        } else if (this.property.type === "Single") {
+            this.property.setSelection(value);
+        } else {
+            throw new Error("Cannot apply new value to current selection type");
+        }
```
tamelbrun, line 106: This is only necessary because this private method removes all the benefits that the discriminated union for the selection property value provides. Splitting the implementation would allow you to avoid this altogether.

packages/shared/pluggable-widgets-commons/src/selection/SelectionWrapper.ts
```diff
@@ -0,0 +1,112 @@
...
+
+        if (this.onChange?.canExecute && !this.onChange.isExecuting) {
+            this.onChange.execute();
+        }
```
tamelbrun, line 110: This approach is unfortunately buggy. Setting the selection may change the `canExecute` flag (e.g. when a single selection goes from `undefined` to an `ObjectItem`), which is only reflected in the props during the next render of the widget component. And simply ignoring the `canExecute` flag is also bad, because it may be `false` for a different reason (and then you'll get a warning in the log).

The most reliable approach that I can think of right now is a `useRef` initialized with `itemSelection.selection` and a `useEffect` which fires the action (and updates the ref) when `itemSelection.selection` is no longer equal to the ref. (Identify comparison should be enough, as the selection property should only emit a new value if there is an actual change.)

packages/pluggableWidgets/gallery-web/src/Gallery.tsx
```diff
@@ -103,81 +104,105 @@ export function Gallery(props: GalleryContainerProps): ReactElement {
...
+                        }
+
+                        return renderWrapper(
+                            !!selection?.isSelected(item),
+                            props.content?.get(item),
+                            props.itemClass?.get(item)?.value,
+                            onClick
+                        );
+                    },
+                    [props.content, props.itemClass, props.onClick, selection]
```
tamelbrun, line 194: This will update the `itemRenderer` whenever the selection changes, which might cause the entire gallery to re-render (rather than just the item that was selected/deselected). You may want to double-check this and apply some optimizations if necessary.

packages/pluggableWidgets/gallery-web/typings/GalleryProps.d.ts
```diff
@@ -38,6 +38,7 @@ export interface GalleryContainerProps {
     tabIndex?: number;
     advanced: boolean;
     datasource: ListValue;
+    itemSelection: any;
```
tamelbrun, line 41: The 9.23 release for `pluggable-widgets-tools` will happen soon, as the release of Mendix 9.23 is scheduled for tomorrow. Be sure to get updated tools & typings before you merge this PR :).

packages/shared/pluggable-widgets-commons/src/selection/SelectionWrapper.ts
tamelbrun: Assuming for the moment that we're going to keep this wrapper approach, you can also split the interface into single and multi (using a discriminated union) and turn it into a strategy object without the need for an `Impl` class (as the object is stateless anyway: it is constantly recreated when any of the props update). See my other comments about this :).

tamelbrun (dismissed):
I've done a first review pass, I think that's quite enough for the moment ;).

packages/shared/pluggable-widgets-commons/src/selection/SelectionWrapper.ts
tamelbrun: WTF is considering implementing `onChange` in the `selection` property itself, similar to `attribute` and `association`. If we manage to land that in 9.24, this will become unnecessary (and will have to be removed).

iobuhov: @tamelbrun, but how then we should proceed? I think then we should remove it and wait until WTF will bring this feature.

packages/pluggableWidgets/gallery-web/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Added
+
+-   With new "Selection" property it is now possible to make items of the gallery selectable. Additionally, it is possible to configure "On selection change" action that runs when the selection changes.
```
iobuhov, line 11:
```suggestion
-   With the new "Selection" property, it is now possible to make items in the gallery selectable. Additionally, it is possible to configure an "On selection change" action that runs when the selection changes.
```

iobuhov (commented):
Yeah, now looks cleaner 👍 , only few notes from my side.

packages/shared/pluggable-widgets-commons/src/selection/index.ts
```diff
@@ -0,0 +1,100 @@
...
+                (selectionHelper.current as SingleSelectionHelper).updateProps(property);
+            }
+        } else {
+            if (!selectionHelper.current) {
+                selectionHelper.current = new MultiSelectionHelper(property, datasource.items ?? []);
+            } else {
+                (selectionHelper.current as MultiSelectionHelper).updateProps(property, datasource.items ?? []);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
r0b1n: There is no need for a separate interface. The value that `useSelectionHelper` returns is an union of two (three with `undefined`) types already. And the following code is going to work as expected.
```ts
if (selection?.type === "Multi") {
    selection.selectAll();
}
```

packages/shared/pluggable-widgets-commons/src/selection/index.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    updateProps(value: SelectionMultiValue, items: ObjectItem[]): void {
+        this.selectionValue = value;
+        this.selectableItems = items;
+    }
+
+    get selectionStatus(): MultiSelectionStatus {
+        return this.selectionValue.selection.length === 0
+            ? "none"
+            : this.selectionValue.selection.length === this.selectableItems.length ?? 0
```
iobuhov, line 41: Why do we need `?? 0` ?
I think `selectableItems` can't be null or undefined

r0b1n: removed

packages/shared/pluggable-widgets-commons/src/selection/SelectionWrapper.ts
tamelbrun: @iobuhov It is not all certain that WTF will make this in 9.24. If we do, then you can always remove it when you update the widget XML (which will cause the action property to disappear from the typings). But for now, it is still needed.

iobuhov:
@r0b1n, I think we should update `mendix` and `pluggable-tools` in separate PR. I'll do it on Monday.

packages/shared/pluggable-widgets-commons/src/selection/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    onChange: ActionValue | undefined
+): SingleSelectionHelper | MultiSelectionHelper | undefined {
+    if (property && onChange) {
+        // eslint-disable-next-line react-hooks/rules-of-hooks
+        useEffect(() => {
+            if (onChange.canExecute && !onChange.isExecuting) {
+                onChange.execute();
+            }
+            // eslint-disable-next-line react-hooks/exhaustive-deps
+        }, [property.selection]);
```
tamelbrun, line 78: Discussed with Roman: we need to make sure that the onChange action only fires after the first load of the data source, to ensure compatibility with the upcoming platform implementation.

r0b1n: Addressed this point as discussed.

leonardomendix (approved):

leonardomendix (approved):

r0b1n (approved):

PR #395, opened 2023-02-24
fix(popup-menu): Onclick action of nested menu not working

wegiswes:
### **Pull request type**

---

### **Description**

**Cause:** 
When you nest two popup menu's, the onclick action of the inner menu doesn't work. This is caused by the "handleOnClickOutsideElement" event of the outer menu, which is blocking the click event of the inner menu.

**Fix:** 
By using the `onClickCapture` event on the custom item, the click event will be fired in the capturing phase, which runs before the bubbling phase (of the "onClickOutsideElement" handler of the outer menu). By also removing `e.stopPropagation` in the custom item click event, the click event is allowed to travel up the React tree to the click event of the "onClickOutsideElement" handler.

The onclick action of (nested) popup menu items.

### DIFF
--- a/packages/pluggableWidgets/popup-menu-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/popup-menu-web/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+-   We fixed an issue where the onclick action of a nested popup menu was not working
+
 ## [3.3.0] - 2023-01-04
 
 ### Changed
--- a/packages/pluggableWidgets/popup-menu-web/package.json
+++ b/packages/pluggableWidgets/popup-menu-web/package.json
@@ -1,7 +1,7 @@
 {
   "name": "popup-menu-web",
   "widgetName": "PopupMenu",
-  "version": "3.3.0",
+  "version": "3.3.1",
   "copyright": "© Mendix Technology BV 2023. All rights reserved.",
   "license": "Apache-2.0",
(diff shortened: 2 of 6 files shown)


iobuhov (changes requested):
During PR check I found that main cause is a memory leak. As `ref.current` is `null` on cleanup `listener` callback is never removed. And having multiple listeners cause menu to be closed before any "click" handler on custom menu item have chance to handle event (as component is unmounted it can't handle events, bubbled or captured). So, just by fixing memory leak we also fixing issue with "On click" action in custom items.

We need to save reference to `ownerDocument` in local scope.
```js
        const document = ref.current?.ownerDocument;
        document?.addEventListener("click", listener);
        document?.addEventListener("touchstart", listener);
        return () => {
            document?.removeEventListener("click", listener);
            document?.removeEventListener("touchstart", listener);
        };
```

wegiswes:
Nice find! However, just applying your change doesn't work for me. Neither does your branch [455](https://github.com/mendix/web-widgets/pull/455).

I have attached a small test project below which includes my widget changes. You will get a dialog when clicking on the nested menu item.
When replacing the widget with one of your versions, it will not run the nanoflow and you will not get the dialog.
[TEST_BlankApp_PopUpMenuWeb_Nested_Bug.mpk.zip](https://github.com/mendix/web-widgets/files/11041730/TEST_BlankApp_PopUpMenuWeb_Nested_Bug.mpk.zip)

I guess a combination of both is required to fix the widget properly.

Sorry for my late reply, I forgot about the PR as we don't use this widget anymore. So feel free to close it and continue in your own PR :)

iobuhov:
Closing this one as it get's outdated.

PR #401, opened 2023-02-27
feat(rich-text-web): make content templates for RichText widget (CKEditor 4) customizable

tobias-at-work:
### **Pull request type**

### **Description**

The Rich Text widget is built around the latest version of CKEditor 4. It already contains the content template plugin, if a "full" toolbar is chosen, or a "custom" tool bar with the "Templates" button configured in the "advanced toolbar groups". However, only the the default templates contained in CKEditor can be used, which is not very useful.

This change allows to configure a new template datasource to provide a list of custom content templates per widget instance 
(https://ckeditor.com/docs/ckeditor4/latest/examples/templates.html). Each content template has for attributes "title", "image", "description" and "html". The image attribute needs to be set in the form {module}${image_collection}${image_name}.{image_format} to access an image from the designated image collection.

Projects that already use the RichText widget, will have to update their widgets, because the XML file has been extended. However, configuring the new properties is optional unless you really want to use the content template feature.

The following aspects should be tested:
- RichText widgets with toolbar = "basic" and toolbar = "standard" as well as toolbar = "custom"  and toolbar group = "basic" should work as before. Even if toolbar group is set to "advanced", the new properties should only matter if the "Templates" button is explicitly added to the toolbar.
- RIchText widgets with toolbar = "full" or toolbar = "custom" and toolbar group = "advanced" and the button "Templates" added to the toolbar, have to configure the six attributes. However, sensible defaults have been defined, and if the template name is set to "default" the default content templates from CKEditor should be available as before.
- As opposed to CKEditor, the widget supports only one template name per editor instance. Typically, the template name should match the widget name.
- Any kind of datasource with four columns of type string should be usable as template source, be it database, association, nanoflow, microflow or XPath. 
- Images from an image collection should be used if referred to correctly. No image should be shown otherwise.

### DIFF
--- a/packages/pluggableWidgets/rich-text-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/rich-text-web/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+## [2.2.0] - 2023-02-27
+
+### Added
+
+-   Enabled customization of CKEditor 4 content template plugin.
+
 ## [2.1.3] - 2023-01-31
 
 ### Fixed
--- a/packages/pluggableWidgets/rich-text-web/package.json
+++ b/packages/pluggableWidgets/rich-text-web/package.json
@@ -1,7 +1,7 @@
 {
   "name": "rich-text-web",
   "widgetName": "RichText",
-  "version": "2.1.3",
+  "version": "2.2.0",
   "description": "Rich inline or toolbar text editing",
   "copyright": "© Mendix Technology BV 2023. All rights reserved.",
   "repository": {
--- a/packages/pluggableWidgets/rich-text-web/src/RichText.editorConfig.ts
+++ b/packages/pluggableWidgets/rich-text-web/src/RichText.editorConfig.ts
@@ -38,11 +38,39 @@ export function getProperties(
     if (values.preset !== "custom") {
         hidePropertiesIn(defaultProperties, values, toolbarGroups.concat(["toolbarConfig", "advancedConfig"]));
     }
+    if (values.preset === "basic" || values.preset === "standard") {
+        hidePropertiesIn(defaultProperties, values, [
+            "templates",
+            "templateDatasource",
+            "templateTitleAttribute",
+            "templateImageAttribute",
+            "templateDescriptionAttribute",
+            "templateHtmlAttribute"
+        ]);
+    }
     if (values.toolbarConfig === "basic") {
-        hidePropertiesIn(defaultProperties, values, ["advancedConfig"]);
+        hidePropertiesIn(defaultProperties, values, [
+            "advancedConfig",
+            "templates",
+            "templateDatasource",
+            "templateTitleAttribute",
+            "templateImageAttribute",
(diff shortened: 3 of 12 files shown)


iobuhov:
Hi @tobias-at-work, thank you for your contribution 🤝 to web-widgets!

Well, this looks like a big PR. What I want to make sure is that you understand that bringing new feature to the widget is a small journey. But don't be scared, we just need time to triage, test and analyze this feature. We may request xml, code or other changes that may be needed to match our desired look and feel for new feature (probably not much in this case).
In general -- you PR looks promising, but we first need make sure it don't brake existing behavior.

Here what we will do next:
1. We will discuss this new feature within our team.
2. If we decide to add it, we will do code review and testing.
3. If all goes well, we will publish this new feature in next minor release.

⚠️ Important: As I see one of our snyk check is failed, this is unfortunate, but to be able to merge your PR all checks should succeed. As of now, only way to "recheck" your PR with snyk is to close/open it. Please close and repone your PR to trigger snyk checks again.

tobias-at-work:
Unfortunately, I cannot use the link to see the nature of the snyk failure (not enabled for my organization?). I will close and reopen the feature as suggested.

tobias-at-work:
Snyk appears to be happy now.

takumagott:
Hi @tobias-at-work , excuses for the long wait. Thank you again for your contribution! We have not forgotten your PR and believe it would generally be a nice addition to the rich text widget. However, we have two minor usability concerns.

1. Having the new content template tab in the main properties dialog

Currently setting it up requires these steps:
* set toolbar to `Custom` => a new tab `Custom toolbar` appears
* switch to new tab and set toolbar group to `Advanced`
* create new item of type `Templates` => a new tab `Content templates` appears
This `Content templates` tab is part of the main properties dialog which is already very full. My suggestion would be to make the template fields part of the `advancedConfig` object. This way we could directly show it in the popup when setting the button type to `Template`, which would show it with the right context and keep the main dialog cleaner.

2. Adding images for the content templates

Currently it requires the path to the image in the following format: `{module_name}${image_collection}${image_name}.{image_format}`. That means that the data for content templates can be configured almost completely inside the running app (title, description, html), but the images for the templates need to be added as resources into image collections in Studio Pro. Also, the usage of string paths is not a common pattern that we use. It would be nice if we could tweak this to use an image property instead so that we could use an image uploader. This probably requires a bit of research into the feasibility first and might not be worth it if the risks/efforts outweigh the benefits.

Unfortunately, our team can't help with these improvements because we have limited capacity and a full roadmap. If you would like to work on these yourself, we can proceed with adding this feature. Else, we will have to put it on hold for now and pick it up further in the future.

tobias-at-work:
Hi @takumagott,

thanks for your feedback. Concerning the first point: Using the custom toolbar is not the only way to make the "Templates"-Button visible. The faster way is to set the option "General/Toolbar" to "full". Then all buttons featured by CKEditor4 become immediately visible. Does this in anyway impact you suggestion to put the template configuration in the "advancedConfig" object?

With respect to the second point: I do not strictly need images for the templates in my project. I just thought it to be nice to have a means to support this feature of content templates at all. So I could remove the images feature from this pull request and create separate one for it. It would makes things a lot easier for me, if I at least could get the rest of the initial pull request through.

PR #433, opened 2023-03-07
[WC-1637] Datagrid selection api

iobuhov:
### **Pull request type**

---

### **Description**
Add selection API for DataGrid

- Selection modes
- Actions onClick and onSelectionChange
- Design properties: striped and hover


packages/pluggableWidgets/datagrid-web/src/Datagrid.editorConfig.ts
```diff
@@ -140,6 +142,24 @@ export function getProperties(
...
+        hidePropertyIn(defaultProperties, values, "itemSelectionMethod");
+    }
+
+    if (itemSelection !== "Multi" || itemSelectionMethod !== "checkbox") {
+        hidePropertyIn(defaultProperties, values, "showSelectAllToggle");
+    }
+
+    if (itemSelection === "None") {
+        hidePropertyIn(defaultProperties, values, "onSelectionChange");
+    }
```
r0b1n, line 158: Two `if` have the same condition, could be combined in one

packages/pluggableWidgets/datagrid-web/src/Datagrid.editorConfig.ts
```diff
@@ -140,6 +142,24 @@ export function getProperties(
...
+
+    return defaultProperties;
```
r0b1n, line 160: No need to return it, I guess.

packages/shared/pluggable-widgets-commons/src/selection/index.ts
```diff
@@ -100,3 +98,8 @@ export function useSelectionHelper(
 
     return selectionHelper.current;
 }
+
+export type { SingleSelectionHelper };
+export type { MultiSelectionHelper };
+export type SelectionHelper = SingleSelectionHelper | MultiSelectionHelper | undefined;
```
r0b1n, line 104: I think it is better for readability to not include `undefined` here, but include it explicitly in places where the helper could be undefined.

packages/pluggableWidgets/datagrid-web/src/features/selection.ts
```diff
@@ -0,0 +1,87 @@
...
+    none = "none",
+    rowClick = "rowClick",
+    checkbox = "checkbox"
+}
+
+export type SelectActionProps<T extends ObjectItem = ObjectItem> = {
+    onSelect: (item: T) => void;
+    onSelectAll: () => void;
+    isSelected: (item: T) => boolean;
+};
```
r0b1n, line 16: It looks like this generic is redundant.

packages/pluggableWidgets/datagrid-web/src/features/selection.ts
```diff
@@ -0,0 +1,87 @@
...
+
+const defaultProps: SelectActionProps = {
+    onSelect: stub(),
+    onSelectAll: stub(),
+    isSelected: stub(false)
+};
```
r0b1n, line 22: Why do we need this `stub()`?
Something like this should work:
```
const defaultProps: SelectActionProps = {
    onSelect: () => undefined,
    onSelectAll: () => undefined,
    isSelected: () => false
};
```

And then `stub` could go.

packages/pluggableWidgets/datagrid-web/src/features/selection.ts
```diff
@@ -0,0 +1,87 @@
...
+    return useMemo(() => {
+        if (!selection) {
+            return defaultProps;
+        } else {
+            return {
+                onSelect: item => {
+                    if (selection?.isSelected(item)) {
+                        selection?.remove(item);
+                    } else {
+                        selection?.add(item);
```
r0b1n, line 34: Possibly redundant `?` operator.

packages/pluggableWidgets/datagrid-web/src/features/selection.ts
```diff
@@ -0,0 +1,87 @@
...
+                onSelectAll: () => {
+                    if (selection.type !== "Multi") {
+                        throw new Error("onSelectAll called when selection is 'Single'");
+                    }
+                    if (selection.selectionStatus === "all") {
+                        selection.selectNone();
+                    } else {
+                        selection.selectAll();
+                    }
+                },
```
r0b1n, line 46: We know the selection type at the stage of generating this object. So we might as well put `() => undefined` to this prop instead if selection is Single. And include the current function without the check is selection is Multi.

packages/pluggableWidgets/datagrid-web/src/features/selection.ts
```diff
@@ -0,0 +1,87 @@
...
+    onSelect: stub(),
+    onSelectAll: stub(),
+    isSelected: stub(false)
+};
+
+export function useOnSelectProps(selection: SelectionHelper): SelectActionProps {
+    return useMemo(() => {
+        if (!selection) {
+            return defaultProps;
+        } else {
```
r0b1n, line 28: This `else` could go as we have the early return.

packages/pluggableWidgets/datagrid-web/src/features/selection.ts
```diff
@@ -0,0 +1,87 @@
+import { useMemo } from "react";
+import type { ObjectItem } from "mendix";
+import { MultiSelectionStatus, SelectionHelper } from "@mendix/pluggable-widgets-commons";
+import { DatagridContainerProps, DatagridPreviewProps, ItemSelectionMethodEnum } from "typings/DatagridProps";
+
+export enum SelectionMethod {
+    none = "none",
+    rowClick = "rowClick",
+    checkbox = "checkbox"
+}
```
r0b1n, line 10: Using enums is not the common pattern in our code unfortunately, we use unions of strings instead. I think is it better to stick with this pattern, this requires less code as it is transparent with what typings say and makes it easier to reason about.

For example instead of 
```
    return {
        selectionStatus: selectAllOn ? status : undefined,
        selectionMethod: selectionOn
            ? methodCheckbox
                ? SelectionMethod.checkbox
                : SelectionMethod.rowClick
            : SelectionMethod.none
    };
```
we could say
```
    return {
        selectionStatus: selectAllOn ? status : undefined,
        selectionMethod: selectionOn ? itemSelectionMethod : "none"
    };
```

packages/pluggableWidgets/datagrid-web/src/Datagrid.editorConfig.ts
iobuhov: I did it on purpose, so we have one branch for one prop, but yeah, we can merge.

r0b1n: We could use `hidePropertiesIn` and pass an array of prop names to hide.

iobuhov: done 🪄

packages/pluggableWidgets/datagrid-web/src/Datagrid.editorConfig.ts
iobuhov: done 🪄

packages/pluggableWidgets/datagrid-web/src/features/selection.ts
iobuhov: done 🪄

packages/pluggableWidgets/datagrid-web/src/features/selection.ts
iobuhov: done 🪄

packages/pluggableWidgets/datagrid-web/src/features/selection.ts
iobuhov: done 🪄

packages/pluggableWidgets/datagrid-web/src/features/selection.ts
iobuhov: done 🪄

packages/pluggableWidgets/datagrid-web/src/features/selection.ts
iobuhov: done 🪄

packages/shared/pluggable-widgets-commons/src/selection/index.ts
iobuhov: done 🪄

packages/pluggableWidgets/datagrid-web/src/components/Table.tsx
```diff
@@ -336,28 +362,52 @@ export function Table<T extends ObjectItem>(props: TableProps<T>): ReactElement
...
+                                )}
                                 {visibleColumns.map(cell => renderCell(cell, row.item, rowIndex))}
                                 {columnsHidable && (
                                     <div
+                                        key="cell_column_dropdown"
                                         aria-hidden
-                                        className={classNames("td column-selector", { "td-borders": rowIndex === 0 })}
+                                        className={classNames("td column-selector", {
+                                            "td-borders": rowIndex === 0,
+                                            "td-selected": isSelected(row.item)
```
r0b1n, line 388: What is the reason we apply selection classes to individual cells instead of the whole row?

iobuhov: But what the benefits of having it on the row? Current table styles will not allow you to say
```css
.tr.tr-selected { background: blue };
```
It will not work.

r0b1n: The benefit is that we apply class only once. And semantically it is a row that gets selected, not cells.
Can we say the following? 
```
.tr.tr-selected .td { background: blue }
```

iobuhov: done

iobuhov: done

r0b1n (approved):

r0b1n (dismissed):

leonardomendix (approved):

r0b1n (approved):

PR #435, opened 2023-03-08
[WC-1690]: Dropdown basic UI

rahmanunver:
### Pull request type

---

### Description
The UI part f the Dropdown with single selection functionality. It is inspired from react-select as discussed.

Also split the component a bit to provide more readability, and it still can be re-organised better in the future as the development goes on.


packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+                            ...changes,
+                            // highlightedIndex: state.highlightedIndex,
+                            // isOpen: true,
+                            inputValue: ""
+                        };
+
+                    default:
+                        return changes;
+                }
+            }}
```
iobuhov, line 77: Well, this probably good case for a hook I see that we have only two "deps": `inputRef` and `selector`. 
We can move all this callbacks into hook (e.g. `useDownshiftProps(selector, inputRef)`) and inside you can use `useMemo`

But, please, keep in mind that we should try to avoid inline functions in component props. Inline functions should be small, otherwise they become hard to maintain.

packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+                getItemProps,
+                getMenuProps,
+                isOpen,
+                // inputValue,
+                highlightedIndex,
+                selectedItem,
+                getRootProps,
+                getToggleButtonProps
+            }) => (
+                <div className="widget-dropdown" {...getRootProps({}, { suppressRefError: true })}>
```
iobuhov, line 91: After checking `downshift` docs I see that you have to call [getRootProps](https://github.com/downshift-js/downshift/blob/c6c0af5ef4655edbddbfa0ae153633bfd5e7d6af/README.md#getrootprops) only if you have [custom layout](https://github.com/downshift-js/downshift/blob/c6c0af5ef4655edbddbfa0ae153633bfd5e7d6af/docusaurus/pages/combobox.js).

Which is not our case. So, please remove them as we probably don't need it.

packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+                    <div
+                        ref={dropdownRef}
+                        style={{ border: isOpen ? "1px solid #264ae5" : undefined }}
```
iobuhov, line 94: Inline styles. This should be classes `active/open` or with other name

packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+                    >
+                        <input
+                            className="flex-grow"
+                            ref={inputRef}
+                            {...getInputProps()}
+                            placeholder={selector.caption.get(selector.currentValue)}
+                            style={{
+                                border: "none"
+                            }}
+                        />
```
iobuhov, line 106: Why we have to put input inside our our `toggleButton`?
Do we need `toggleButton` at all? Maybe we can just use surrounding div for styling and we don't need to call `getToggleButtonProps`

packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+                                    e.preventDefault();
+                                    e.stopPropagation();
+                                    selectItem(null);
+                                }}
+                            >
+                                <div className="widget-dropdown-clear-button">
+                                    <svg fill={isOpen ? "black" : "gray"} height="10" width="10" viewBox="0 0 329 329">
+                                        <path d="m194.800781 164.769531 128.210938-128.214843c8.34375-8.339844 8.34375-21.824219 0-30.164063-8.339844-8.339844-21.824219-8.339844-30.164063 0l-128.214844 128.214844-128.210937-128.214844c-8.34375-8.339844-21.824219-8.339844-30.164063 0-8.34375 8.339844-8.34375 21.824219 0 30.164063l128.210938 128.214843-128.210938 128.214844c-8.34375 8.339844-8.34375 21.824219 0 30.164063 4.15625 4.160156 9.621094 6.25 15.082032 6.25 5.460937 0 10.921875-2.089844 15.082031-6.25l128.210937-128.214844 128.214844 128.214844c4.160156 4.160156 9.621094 6.25 15.082032 6.25 5.460937 0 10.921874-2.089844 15.082031-6.25 8.34375-8.339844 8.34375-21.824219 0-30.164063zm0 0" />
+                                    </svg>
```
iobuhov, line 120: 1. Please, move icons to `src/assets/icons.tsx`
2. Why we use named colors (black, gray)?  This should be part of `scss` and they probably should be in variable. As this also should be configured by theme.

packages/pluggableWidgets/dropdown-web/src/components/DropdownMenu.tsx
```diff
@@ -0,0 +1,53 @@
...
+    isOpen: boolean;
+    selector: SingleSelector;
+    highlightedIndex: number | null;
+    selectedItem: string | null;
+}
+
+export function DropdownMenu(props: DropdownMenuProps) {
+    const { dropdownSize, isOpen, selector, highlightedIndex, selectedItem } = props;
+    return (
+        <div>
```
iobuhov, line 16: Why we need this div?

packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+                                </div>
+                            </button>
+                        )}
+                        <div className="widget-dropdown-down-arrow">
+                            <svg fill={isOpen ? "black" : "gray"} height="20" width="20" viewBox="0 0 20 20">
+                                <path d="M4.516 7.548c0.436-0.446 1.043-0.481 1.576 0l3.908 3.747 3.908-3.747c0.533-0.481 1.141-0.446 1.574 0 0.436 0.445 0.408 1.197 0 1.615-0.406 0.418-4.695 4.502-4.695 4.502-0.217 0.223-0.502 0.335-0.787 0.335s-0.57-0.112-0.789-0.335c0 0-4.287-4.084-4.695-4.502s-0.436-1.17 0-1.615z"></path>
+                            </svg>
+                        </div>
+                    </div>
+                    <DropdownMenu
```
iobuhov, line 130: Should we use portal?

packages/pluggableWidgets/dropdown-web/src/components/DropdownMenu.tsx
```diff
@@ -0,0 +1,53 @@
...
+            <ul
+                className="dropdown-menu"
+                {...props.getMenuProps?.()}
+                style={{
+                    display: !isOpen ? "none" : "inline",
+                    left: "unset",
+                    width: dropdownSize,
+                    padding: 0
+                }}
```
iobuhov, line 25: Please, change inline style to classes. Use `classNames` helper.

packages/pluggableWidgets/dropdown-web/src/ui/Dropdown.scss
```diff
@@ -0,0 +1,53 @@
...
+.dropdown-item {
+    display: flex;
+    flex-wrap: wrap;
+    align-content: center;
+    cursor: pointer;
+    user-select: none;
+    padding: 3px 0 3px 10px;
+    height: fit-content;
+    overflow: hidden;
+}
```
iobuhov, line 45: Let's not do same mistake as we did with other widgets. We need to use "module name" in front of any class name (e.g. `widget-dropdown-menu` or `widget-dropdown-item`).
This will prevent css rule clashes and simplify future changes.

packages/pluggableWidgets/dropdown-web/src/ui/Dropdown.scss
```diff
@@ -0,0 +1,53 @@
...
+
+.flex-grow {
+    flex-grow: 1;
+}
```
iobuhov, line 49: Probably we need more clear class name, not just utility class.

packages/pluggableWidgets/dropdown-web/src/ui/Dropdown.scss
```diff
@@ -0,0 +1,53 @@
...
+
+.form-control {
+    display: flex;
+}
```
iobuhov, line 53: Please, don't override styles for atlas. If you need apply display flex, create separate class with bigger wight that will override `display` property.

packages/pluggableWidgets/dropdown-web/src/helpers/Association/AssociationOptionsProvider.ts
```diff
@@ -37,16 +34,6 @@ export class AssociationOptionsProvider implements OptionsProvider<ObjectItem, P
...
-        if (this.ds && this.attributes.length > 0) {
-            console.log("applyting ", this.searchTerm);
-            if (this.searchTerm !== "") {
-                const a = this.attributes.map(a => contains(attribute(a.id), literal(this.searchTerm)));
-                this.ds?.setFilter(a.length === 1 ? a[0] : or(...a));
-            } else {
-                this.ds?.setFilter(undefined);
-            }
-        }
     }
```
iobuhov, line 37: Why we removing search?

packages/pluggableWidgets/dropdown-web/src/components/DropdownMenu.tsx
```diff
@@ -0,0 +1,53 @@
...
+                                  index,
+                                  item: item,
+                                  style: {
+                                      backgroundColor:
+                                          selectedItem === item
+                                              ? "rgb(101 163 244)"
+                                              : highlightedIndex === index
+                                              ? "rgb(195 221 255)"
+                                              : "white",
+                                      color: selectedItem === item ? "white" : "black"
```
iobuhov, line 42: 1. Please, don't use CSS color names.
2. Save all colors to sass variables.
3. Give this variables meaningful name

iobuhov (changes requested):
Let's start preparing widget for publishing. 
- Take time and refactor your code. Remove commented code and inline styling.
- CSS classes should be prefixed with `widget-dropdown`
- Icons should be organaized [same way as in datagrid](https://github.com/mendix/web-widgets/tree/main/packages/pluggableWidgets/datagrid-web/src/components/icons)
- Do we need tests?
- Lst but not lease, @r0b1n, should we move selectors from `src/helpers` to `src/selectors`?
- @takumagott, should we use new icon set? Or we can just hardcode current icons?

packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+export function Dropdown(props: DropdownContainerProps): ReactElement {
+    const selectorRef = useRef<SingleSelector | undefined>(undefined);
+    const inputRef = useRef<HTMLInputElement>(null);
+    const dropdownRef = useRef<HTMLInputElement>(null);
+    if (!selectorRef.current) {
+        selectorRef.current = getSelector(props);
+    }
+    const selector = selectorRef.current!;
+
+    selector.updateProps(props);
```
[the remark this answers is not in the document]
r0b1n: @iobuhov the suggested change unnecessary calls `getSelector` every render. This effectively creates a new selector object and then throws it away.

packages/pluggableWidgets/dropdown-web/src/helpers/Association/AssociationOptionsProvider.ts
r0b1n: This was just an experiment, it it not part of the basic implementation. We don't plan to include lazy loading and/or partial fetch of objects in this initial version. This effectively means we have all objects from data source in our hands, and can filter in memory. In the next version we will address search and partial loading.

packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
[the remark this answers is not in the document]
iobuhov: Ah ,yes, you right, but then should we use `useState`?
Like:
```
const [selector] = useState(() => getSelector(props));
```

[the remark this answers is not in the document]
iobuhov: cc: @r0b1n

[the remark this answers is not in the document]
r0b1n: I don't think we need state for this, it never changes. What was wrong with the initial implementation?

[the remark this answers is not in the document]
iobuhov: well, I though 1 line with `useState` is bit more compact and easy to read. + non-null assertion (`.current!`) bit triggered me

packages/pluggableWidgets/dropdown-web/src/helpers/Association/AssociationOptionsProvider.ts
iobuhov: Ah, ok, got it

packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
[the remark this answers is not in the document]
r0b1n: We could move it out to a custom hook, and possible make it look like:
```
    const selectorRef = useRef<SingleSelector | undefined>(undefined);
    if (!selectorRef.current) {
        selectorRef.current = getSelector(props);
    } else {
        selectorRef.current.updateProps(props);
    }
    const selector = selectorRef.current!;
```

packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { useGetSelector } from "../hooks/useGetSelector";
+import { DropdownContainerProps } from "../../typings/DropdownProps";
+import { DropdownMenu } from "./DropdownMenu";
+import { Placeholder } from "./Placeholder";
+
+export function Dropdown(props: DropdownContainerProps): ReactElement {
+    const inputRef = useRef<HTMLInputElement>(null);
+    const dropdownRef = useRef<HTMLInputElement>(null);
+    const selector = useGetSelector(props);
+    selector.updateProps(props);
```
r0b1n, line 15: Can we move this to the `useGetSelector`?

packages/pluggableWidgets/dropdown-web/src/__tests__/__snapshots__/Dropdown.spec.tsx.snap
```diff
@@ -0,0 +1,187 @@
...
+          >
+            <path
+              d="M4.516 7.548c0.436-0.446 1.043-0.481 1.576 0l3.908 3.747 3.908-3.747c0.533-0.481 1.141-0.446 1.574 0 0.436 0.445 0.408 1.197 0 1.615-0.406 0.418-4.695 4.502-4.695 4.502-0.217 0.223-0.502 0.335-0.787 0.335s-0.57-0.112-0.789-0.335c0 0-4.287-4.084-4.695-4.502s-0.436-1.17 0-1.615z"
+            />
+          </svg>
+        </div>
+      </div>
+    </div>
+  </div>,
+  "debug": [Function],
```
r0b1n, line 135: This looks like something we don't want to include in the snapshot. We are probably checking too much.

packages/pluggableWidgets/dropdown-web/src/__tests__/__snapshots__/Dropdown.spec.tsx.snap
```diff
@@ -0,0 +1,187 @@
...
+            class="widget-dropdown-input"
+            id="downshift-0-input"
+            placeholder=""
+            value=""
+          />
+          <button
+            class="widget-dropdown-clear-button"
+          >
+            <div
+              class="widget-dropdown-clear-button"
```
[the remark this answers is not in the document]
luisarthurmicainunes: Aren't you on holidays, Illia? :)

packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
rahmanunver: Do you mean the `updateProps` part? We can actually.

packages/pluggableWidgets/dropdown-web/src/__tests__/__snapshots__/Dropdown.spec.tsx.snap
rahmanunver: Yes, for snapshot I should've used `component.container` rather than just `component`

packages/pluggableWidgets/dropdown-web/src/__tests__/__snapshots__/Dropdown.spec.tsx.snap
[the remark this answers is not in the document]
rahmanunver: Indeed it's a mistake. Fixed now.

packages/pluggableWidgets/dropdown-web/src/__tests__/__snapshots__/Dropdown.spec.tsx.snap
rahmanunver: Done

packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
rahmanunver: Done

packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
rahmanunver: Done

packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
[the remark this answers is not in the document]
rahmanunver: Done

packages/pluggableWidgets/dropdown-web/src/components/Dropdown.tsx
rahmanunver: The properties for toggling comes from these props, so we're keeping them in discussion with @iobuhov

packages/pluggableWidgets/dropdown-web/src/__tests__/__snapshots__/Dropdown.spec.tsx.snap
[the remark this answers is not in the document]
iobuhov: 😂 🙈 , it's not me, it's my AI guard.

PR #448, opened 2023-03-14
fix(web-widgets): add public-hoist patterns

iobuhov:
### Pull request type

---

### Description
With `pnpm` update to 7.29 our PWT jest config don't work anymore (which is expected, as we using [node-linker=isolated](https://pnpm.io/npmrc#node-linker)). To fix that, we need to mark some deps as "public" using [public-hoist-pattern](https://pnpm.io/npmrc#public-hoist-pattern). If package match pattern it will be available in `node_modules` even if it's not listed in package dependencies.


packages/customWidgets/signature-web/package.json
```diff
@@ -44,6 +44,7 @@
     "@types/react-resize-detector": "^5.0.0",
     "css-loader": "^6.7.3",
     "eslint": "^7.32.0",
+    "identity-obj-proxy": "^3.0.0",
```
leonardomendix, line 47: I didn't get why we need that.

iobuhov: This packages is part of jest config, `identity-obj-proxy` used as a mock for `CSS` and `SCSS` modules.

iobuhov: And it was missing in package deps

r0b1n (approved):

packages/customWidgets/signature-web/package.json
leonardomendix: Thanks!

leonardomendix (dismissed):

leonardomendix (approved):

PR #467, opened 2023-03-23
[WC-1763]: barcode scanner improvements

HedwigAR:
### Pull request type

---

### Description
- Adds a different scan method when the mask is enabled, this should only scan part of the screen instead of all. This improves scan accuracy for 1 dimensional codes in particular. 
- Changes implementation of Detect callback to always fire, since a datasource will always be attached but an event is not required. This causes the barcode scanner to 'fail' silently when first using it and having a datasource connected but no event set. 

---

- Non masked version of the widget should have no changes from current implementation (apart from Detect callback change), but the masked option should now have improved accuracy for 1 dimensional codes.

### DIFF
--- a/packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
+++ b/packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
@@ -4,9 +4,10 @@ import {
     BinaryBitmap,
     BrowserMultiFormatReader,
     DecodeHintType,
+    HTMLCanvasElementLuminanceSource,
     HybridBinarizer,
     NotFoundException,
-    RGBLuminanceSource
+    Result
 } from "@zxing/library/cjs";
 import { useEventCallback } from "@mendix/pluggable-widgets-commons";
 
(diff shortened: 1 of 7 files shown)


packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
```diff
@@ -23,16 +32,50 @@ const mediaStreamConstraints: MediaStreamConstraints = {
 type UseReaderHook = (args: {
     onSuccess?: (data: string) => void;
     onError?: (e: Error) => void;
+    showMask: boolean;
```
r0b1n, line 35: Maybe name it something like "useCrop" or "doCrop", or just "crop". Also I would pass crop value as prop to this hook from the component.

packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
```diff
@@ -44,10 +87,24 @@ export const useReader: UseReaderHook = args => {
...
-                    if (!stopped) {
+                    if (args.showMask) {
+                        videoRef.current.srcObject = stream;
+                        videoRef.current.autofocus = true;
+                        videoRef.current.playsInline = true; // Fix error in Safari
+                        await videoRef.current.play();
+                        const captureCanvas = reader.createCaptureCanvas(videoRef.current!);
+                        captureCanvas.width = videoRef.current.videoWidth * scale;
+                        captureCanvas.height = videoRef.current.videoWidth * scale;
+                        const result = await scanWithCropOnce(reader, captureCanvas);
```
r0b1n, line 101: It feels we should move this canvas creation part into `scanWithCropOnce`. Also it might use onSuccess directly, no need to return results to this place back.

packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
```diff
@@ -23,16 +32,52 @@ const mediaStreamConstraints: MediaStreamConstraints = {
...
     const videoRef = useRef<HTMLVideoElement>(null);
     const onSuccess = useEventCallback(args.onSuccess);
     const onError = useEventCallback(args.onError);
+    const scale = 0.5;
+
+    const scanWithCropOnce = (reader: BrowserMultiFormatReader, canvas: HTMLCanvasElement): Promise<Result> => {
+        const loop: any = (resolve: (value?: Result) => void, reject: (reason?: any) => void) => {
+            try {
+                const canvasContext = canvas.getContext("2d");
+                if (canvasContext !== null) {
```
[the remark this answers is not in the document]
HedwigAR: Hi! Thanks, this should be fixed now.

packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
HedwigAR: Changed

packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
HedwigAR: Yep, I moved it

packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
```diff
@@ -44,10 +91,21 @@ export const useReader: UseReaderHook = args => {
             let stream;
             try {
                 stream = await navigator.mediaDevices.getUserMedia(mediaStreamConstraints);
+                reader.hints = hints;
+                reader.timeBetweenDecodingAttempts = 500;
```
r0b1n, line 95: What is the reason we moved those setting here instead of passing those to the constructor?

HedwigAR: No good reason anymore, I've moved it to the constructor

iobuhov (dismissed):

r0b1n (dismissed):

iobuhov (approved):

leonardomendix (approved):

r0b1n (approved):

PR #492, opened 2023-04-06
[WC-1605]: group chart items where grouping value is not reachable as '(empty)'

r0b1n:
### Pull request type

and

---

### Description
Previously, when some items had no data for grouping the chart was never displayed, this change introduces new `(empty)` group for those items and groups those items.

Charts display data in the situation while it wasn't previously.

### DIFF
--- a/packages/shared/charts/src/hooks/usePlotChartDataSeries.ts
+++ b/packages/shared/charts/src/hooks/usePlotChartDataSeries.ts
@@ -143,38 +143,38 @@ function groupDataSourceItems(series: PlotDataSeries): DataSourceItemGroup[] | n
     for (const item of dataSource.items) {
         const groupByAttributeValue = ensure(groupByAttribute).get(item);
 
-        if (groupByAttributeValue.value === undefined) {
+        if (groupByAttributeValue.status === "loading") {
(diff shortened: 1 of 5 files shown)


packages/shared/charts/src/hooks/usePlotChartDataSeries.ts
```diff
@@ -143,38 +143,38 @@ function groupDataSourceItems(series: PlotDataSeries): DataSourceItemGroup[] | n
...
-            if (dynamicName) {
-                const dynamicSeriesNameValue = dynamicName.get(item);
+        group.items.push(item);
 
-                if (dynamicSeriesNameValue.value === undefined) {
-                    return null;
-                }
-
-                newDataSourceItemGroup.dynamicNameValue = dynamicSeriesNameValue.value;
+        if (dynamicName && (!group.dynamicNameValue || group.dynamicNameValue === "(empty)")) {
```
iobuhov, line 171: Is `"(empty)"` is a literal? I'm just curious where we get this "empty".

iobuhov (approved):
LGTM

rahmanunver (approved):

PR #493, opened 2023-04-07
[WC-1716] Issue dg inf loop

iobuhov:
### Pull request type

---

### Description
This PR change how filters react on changes to the defaultValue attribute. Now we use value from this attribute only as initial value and any subsequent change is ignored.

Filter widgets - rendering and behavior.

You need a datagrid wrapped with dataview, which is used for storing and managing default values for the filters.

Current behavior:
changing filter value or showing/hiding filter through column selector cause multiple rerenders, poor performance and rendering erros in filter widgets.

Expected:
Changing filter value or hiding/showing filter should not cause any observable flickering or performance drops.

### DIFF
--- a/packages/pluggableWidgets/datagrid-text-filter-web/src/DatagridTextFilter.tsx
+++ b/packages/pluggableWidgets/datagrid-text-filter-web/src/DatagridTextFilter.tsx
@@ -77,7 +77,7 @@ export default function DatagridTextFilter(props: DatagridTextFilterContainerPro
                     return <Alert bootstrapStyle="danger">{errorMessage}</Alert>;
                 }
 
-                if (props.defaultValue && props.defaultValue.status === "loading") {
+                if (props.defaultValue?.status === "loading") {
                     return null;
                 }
 
--- a/packages/pluggableWidgets/datagrid-text-filter-web/src/DatagridTextFilter.tsx
+++ b/packages/pluggableWidgets/datagrid-text-filter-web/src/DatagridTextFilter.tsx
@@ -77,12 +77,17 @@ export default function DatagridTextFilter(props: DatagridTextFilterContainerPro
                     return <Alert bootstrapStyle="danger">{errorMessage}</Alert>;
                 }
 
+                if (props.defaultValue && props.defaultValue.status === "loading") {
+                    return null;
+                }
+
                 return (
                     <FilterComponent
                         adjustable={props.adjustable}
                         className={props.class}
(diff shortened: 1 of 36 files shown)


samuelreichert (dismissed):
LGTM

rahmanunver (approved):

packages/pluggableWidgets/datagrid-text-filter-web/src/features/filter-state.ts
```diff
@@ -0,0 +1,55 @@
...
+    useEffect(() => {
+        const { current: prevState } = prevStateRef;
+        if (state.type !== prevState.type) {
+            stableDispatch(state.inputValue, state.type);
+            inputRef.current?.focus();
+        } else if (state.inputValue !== prevState.inputValue) {
+            stableDispatchDelayed(state.inputValue, state.type);
+        }
+        prevStateRef.current = state;
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
iobuhov: Nice question!
Well, this is my intent. 
1. The idea of this `useEffect` is to react _only_ to state changes.
2. Both, `stableDispatch` and `stableDispatchDelayed` are "stable" -- they both use same reference on each render and strictly equal. That means it not worth to add them to deps as they never change during component lifecycle.

leonardomendix (approved):

leonardomendix (dismissed):

packages/pluggableWidgets/datagrid-text-filter-web/src/DatagridTextFilter.tsx
```diff
@@ -77,24 +77,26 @@ export default function DatagridTextFilter(props: DatagridTextFilterContainerPro
                     return <Alert bootstrapStyle="danger">{errorMessage}</Alert>;
                 }
 
+                if (props.defaultValue && props.defaultValue.status === "loading") {
```
samuelreichert, line 80: I have one suggestion, maybe you could make it shorter by:
```suggestion
                if (props.defaultValue?.status === "loading") {
```

samuelreichert (dismissed):
LGTM

packages/pluggableWidgets/datagrid-text-filter-web/src/DatagridTextFilter.tsx
iobuhov: Done

leonardomendix (approved):

samuelreichert (approved):

PR #495, opened 2023-04-13
feat: upload image to mendix endpoint

MeisamMahdian:
### Pull request type

---

### Description
Enable upload in the rich-text-editor image dialog. The developer needs to provide an upload endpoint to receive the image file and then return the file URL in the same endpoint.


MeisamMahdian:
Out of date PR

PR #496, opened 2023-04-18
[WC-1778]: rich text - apply changes on losing focus

r0b1n:
### Pull request type

---

### Description

Monitor active element and apply latest seen value when user leaves Rich Text widget.

Rich text still:
 - saves the data with denounce while typing
 - saves data immediately when leaving the field
 - doesn't give any errors when leaving page with rich text
 - multiple editors work on the same page correctly and don't conflict


packages/pluggableWidgets/rich-text-web/src/components/Editor.tsx
```diff
@@ -171,19 +178,24 @@ export class Editor extends Component<EditorProps> {
...
     onChange(_event: CKEditorEventPayload<"change">): void {
         if (this.editor) {
             const editorData = this.editor.getData();
             const content = this.widgetProps.sanitizeContent ? DOMPurify.sanitize(editorData) : editorData;
             this.lastSentValue = content;
-            this.widgetProps.stringAttribute.setValue(content);
+            this.applyChangesDebounce();
         }
 
         this.widgetProps.onChange?.execute();
```
iobuhov: I think to keep consistency for `onChange` action we need move this line to `applyChangesImmediately` as `onChange` was debounced, and now it's not.

packages/pluggableWidgets/rich-text-web/src/components/Editor.tsx
```diff
@@ -253,6 +265,25 @@ export class Editor extends Component<EditorProps> {
...
+                const newHasFocus = this.divRef.current.parentElement.contains(document.activeElement);
+                if (newHasFocus !== this.hasFocus) {
+                    this.hasFocus = newHasFocus;
+                    if (!this.hasFocus) {
+                        // changed from true to false, user left the element, apply changes immediately
+                        this.applyChangesImmediately();
+                    }
+                }
+            }
+        });
```
iobuhov: I think this could be changed to use `this.element`, but would be a subject to discuss with @r0b1n in DM.

iobuhov (approved):
LGTM

rahmanunver (approved):
Tested, works as intended!

PR #503, opened 2023-05-04
[WC-1820] Fix regression filter regression introduced in data-widgets 2.7.1

iobuhov:
### Pull request type

---

### Description
As a result of refactoring and fixing filter value initialization we missed call that sets initial datasource condition. This PR bring this call(s) back.


packages/pluggableWidgets/datagrid-dropdown-filter-web/src/components/FilterComponent.tsx
```diff
@@ -301,22 +225,59 @@ export function FilterComponent(props: FilterComponentProps): ReactElement {
...
+            placeholder={props.emptyOptionCaption}
+            onContentScroll={props.onContentScroll}
+            onTriggerClick={props.onTriggerClick}
+        />
+    );
 }
 
-function isSelected(selected: FilterOption[], option: FilterOption): boolean {
-    return !!selected.find(hasSameValue(option));
+function useSetInitialConditionEffect(
```
rahmanunver, line 276: There are two methods with same name, they have different logics but naming is a bit confusing.

packages/pluggableWidgets/datagrid-dropdown-filter-web/src/features/setup.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const EMPTY_OPTION_VALUE = "__EMPTY_VALUE__";
+
+export function finalizeOptions(
+    options: Option[],
+    props: { multiSelect: boolean; emptyOptionCaption?: string }
+): Option[] {
+    return props.multiSelect
+        ? options
+        : [{ caption: props.emptyOptionCaption ?? "", value: EMPTY_OPTION_VALUE }, ...options];
```
rahmanunver, line 15: Why do we return the `...options` too in case `multiSelect` is `false`?

rahmanunver (changes requested):
Added some comments and questions with @samuelreichert , great work 🕺

packages/pluggableWidgets/datagrid-number-filter-web/cypress/integration/DataGridNumberFilter.spec.js
```diff
@@ -6,19 +6,34 @@ describe("datagrid-number-filter-web", () => {
     });
 
     describe("visual testing:", () => {
-        it("compares with a screenshot baseline and checks if all datagrid and filter elements are rendered as expected", () => {
+        it.skip("compares with a screenshot baseline and checks if all datagrid and filter elements are rendered as expected", () => {
```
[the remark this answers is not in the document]
iobuhov: Just forgot to remove

packages/pluggableWidgets/datagrid-dropdown-filter-web/src/components/FilterComponent.tsx
iobuhov: Yeah, it's bit confusing, but this is the idea - each package has it's own `usetSetInitialConditionEffect`.
For now it's how we fit the issue.

rahmanunver (dismissed):

leonardomendix (approved):

gjulivan (approved):

PR #515, opened 2023-05-15
feat: add custom caption to widgets (A-I)

samuelreichert:
### Pull request type

---

### Description

Added custom captions to all pluggable widgets from A to I.


packages/pluggableWidgets/badge-button-web/src/BadgeButton.editorConfig.ts
```diff
@@ -61,3 +61,7 @@ export function getPreview(values: BadgeButtonPreviewProps, isDarkMode: boolean)
         ]
     };
 }
+
+export function getCustomCaption(values: BadgeButtonPreviewProps): string {
+    return values.label ?? "Badge button";
```
gjulivan, line 66: please test on string empty

packages/pluggableWidgets/area-chart-web/src/AreaChart.editorConfig.ts
```diff
@@ -135,3 +135,19 @@ export function check(values: AreaChartPreviewProps): Problem[] {
...
 }
+
+export function getCustomCaption(values: AreaChartPreviewProps): string {
+    type DsProperty = { caption?: string };
+
+    const datasources = values.series.map(serie => {
+        const ds = serie.dataSet === "dynamic" ? serie.dynamicDataSource : serie.staticDataSource;
+        const dsProperty: DsProperty = datasource(ds)().property ?? {};
+        return dsProperty.caption?.replace("[", "").replace("]", "");
+    });
```
iobuhov, line 146: Is it worth to loop over all series?
We using only first caption.
On line 148 we checking the length, maybe it worth to do it first?
Any way, it's just a nit, I don't think there big performance issue.

iobuhov (dismissed):

packages/pluggableWidgets/area-chart-web/src/AreaChart.editorConfig.ts
samuelreichert: Good catch!

iobuhov: So, should we expect changes?

gjulivan (approved):

PR #516, opened 2023-05-15
feat: add custom caption to widgets (I-Z)

gjulivan:
Pull request type

Description
Added custom captions to all pluggable widgets from A to I.

Progress bar(show config/DS)
Progress circle (show config/DS)
Range Slider (Show DS)
Rating (show DS)
Rich Text (show Attribute)
Slider (Show DS)
Switch (show attribute, like with text input)
Timeline (show DS)
Tree node (show DS)
Video Player (Show DS)

image


iobuhov (approved):

packages/pluggableWidgets/video-player-web/src/VideoPlayer.editorConfig.ts
```diff
@@ -63,3 +63,7 @@ export function getPreview(values: VideoPlayerPreviewProps): StructurePreviewPro
         height: 211
     };
 }
+
+export function getCustomCaption(values: VideoPlayerPreviewProps, _platform = "desktop"): string {
+    return values.videoUrl || `Video Player`;
```
iobuhov, line 68: This is the only place that concerns me as there might be pretty long string.

rahmanunver (dismissed):

packages/pluggableWidgets/video-player-web/src/VideoPlayer.editorConfig.ts
gjulivan: regex used to retrieve host name

packages/pluggableWidgets/tree-node-web/src/TreeNode.editorConfig.ts
```diff
@@ -165,3 +166,9 @@ function getChevronIconPreview(headerType: HeaderTypeEnum, isDarkMode: boolean):
         ]
     };
 }
+
+export function getCustomCaption(values: TreeNodePreviewProps, _platform = "desktop"): string {
+    type DsProperty = { caption?: string };
+    const dsProperty: DsProperty = datasource(values.datasource)().property ?? {};
+    return dsProperty.caption || "Tree node";
+}
```
r0b1n, line 174: Using `datasource` helper is unnecessary. An equivalent of current reading of the `caption` field would be something like `(values.data as { caption?: string }).caption`. This cast is necessary for now because we don't have the `caption` field reflected in the typings, something to be fixed with PWT 9.24. After than it is going to be just `values.data.caption`.

gjulivan: fixed

iobuhov (dismissed):

leonardomendix (dismissed):

leonardomendix (approved):

samuelreichert (approved):
LGTM! 😄

PR #534, opened 2023-05-24
feat: cnc rich text editor upload image

MeisamMahdian:
### Pull request type

---

### Description
Enable upload in the rich-text-editor image dialog.
The PR for Mendix Docs changes is [here](https://github.com/mendix/docs/pull/6077)
The developer must provide an upload endpoint to receive the image file and then return the file URL in the same endpoint. We also have created a Mendix module to make it easier.

If the upload-image is disabled in the setting, then the end user should not see the upload tab in the insert image dialog.
If the upload-image is enabled and configured, the end user should be able to upload an image from their local computer to the Mendix server.
Moreover, if the user drag & drop an image to the editor, it will be uploaded to the server instead of putting a base64 format image.


MeisamMahdian:
I added a commit.
If the upload image is enabled then for drag&drop also the image will be uploaded to the server.

MeisamMahdian:
Since the underlying lib has changed the changes are not valid anymore

iobuhov:
Closing as PR not got review for a long time and new major version of the widget was released.

PR #544, opened 2023-05-31
[WC-1343]: Change z-index to solve dropdown visibility

rahmanunver:
### Pull request type

---

### Description
The z-index was set low for the `dropdown-content`, therefore multiple layers of modal/popups cause visibility issues after a while, now it is a higher value and shouldn't cause similar issues.

Need to make sure multiple layers of modal/popups doesn't break the dropdown. JIRA Story has a nice test project where issue can be reproduced.

### DIFF
--- a/packages/modules/data-widgets/CHANGELOG.md
+++ b/packages/modules/data-widgets/CHANGELOG.md
@@ -6,8 +6,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
--   We fixed visilibity issue when dropdown filter is used with layers of modals/popups.
-
 ## [2.8.0] DataWidgets - 2023-05-26
 
 ### [2.5.1] DatagridDateFilter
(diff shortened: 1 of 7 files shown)


iobuhov (approved):

leonardomendix (dismissed):

packages/modules/data-widgets/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+-   We fixed visilibity issue when dropdown filter is used with layers of modals/popups.
```
gjulivan, line 9: is this needed? i thought module will compile the changelog based on the changed widgets?

leonardomendix (dismissed):

packages/modules/data-widgets/CHANGELOG.md
rahmanunver: True, will just leave the version bump.

iobuhov (approved):

leonardomendix (approved):

PR #547, opened 2023-06-05
[WC-1789]: Update PWT to 9.24 and other dependencies

samuelreichert:
### Pull request type

---

### Description

Updating widgets to PWT 9.24


gjulivan (approved):

iobuhov (approved):

package.json
```diff
@@ -31,27 +32,17 @@
...
       "loader-utils@<1.4.1": ">=1.4.1",
       "loader-utils@>=3.0.0 <3.2.1": ">=3.2.1",
       "decode-uri-component@<0.2.1": ">=0.2.1",
+      "jest": "^29.4.1",
+      "jest-environment-jsdom": "^29.4.1",
       "json5@1.x": ">=1.0.2",
       "json5@0.x": ">=1.0.2",
-      "@mendix/pluggable-widgets-tools": "9.23.2"
+      "@mendix/pluggable-widgets-tools>typescript@4": "5.0.4",
+      "@mendix/pluggable-widgets-tools>ts-node@9": "^10.9.1"
```
iobuhov, line 45: Wow, never seen this syntax before.

leonardomendix: Can we use the version with a reference like in the example below?

{
  "dependencies": {
    "foo": "^1.0.0"
  },
  "overrides": {
    // BAD, will throw an EOVERRIDE error
    // "foo": "^2.0.0"
    // GOOD, specs match so override is allowed
    // "foo": "^1.0.0"
    // BEST, the override is defined as a reference to the dependency
    "foo": "$foo",
    // the referenced package does not need to match the overridden one
    "bar": "$foo"
  }
}

r0b1n: We figured out this is not working for PNPM as this is apparently only NPM stuff.

packages/customWidgets/any-chart-web/package.json
```diff
@@ -47,8 +47,8 @@
     "react-dom": "~17.0.2",
     "sass-loader": "^13.2.0",
     "ts-loader": "^9.4.2",
-    "ts-node": "^9.0.0",
-    "typescript": "4.5.4",
+    "ts-node": "^10.9.1",
+    "typescript": "5.0.4",
     "webpack": "^5.75.0",
     "webpack-cli": "^4.10.0"
   },
```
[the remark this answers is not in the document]
samuelreichert: Done!

leonardomendix (approved):

r0b1n (approved):

PR #591, opened 2023-06-19
chore: remove all react and react-dom from deps and change how pnpm works

iobuhov:
### Pull request type

---

### Description
~Fuck this shit.~ We adopt our repo for PWT, now we should switch every ~fucking~ dependency... Every ~fucking~ time. No more, I tried to keep our packages independent, I hoped it will work well, but I was wrong. Sorry, for emotions, but I fed up, seriously. With this PR I enable [shamefully-hoist](https://pnpm.io/npmrc#shamefully-hoist) and I don't care about any conflict it may cause.

# From Emotional to Rational

Yesterday I was a bit emotional because of the [thread](https://mendix.slack.com/archives/C8AF2EVEW/p1687176980261279), where I overreacted (which was unnecessary) for a few comments and our current state of tooling. So, after calming down, I decided that this PR would be a good moment to share my learnings about package management and our monorepo.

Also, I hope this will be kind of a history record and may help explain some *"why"* questions.

The whole story actually starts in October 2022, when, as a result of team reorganizing in Mendix, we were split into MOO and WC. In fact, after that, we decided to archive our old monorepo and move forward with new, independent places for each team. At that moment, we have to split the codebase and decide whether we stay with `Lerna` or move toward new tooling (after the split, each team has full control over the owned codebase).

 

## 1. The king is dead; long live the king!

So, as one of those who participated in this whole "code migration", I was experimenting with new ways of managing packages in our monorepo. We had to choose between two options: `PNPM` and `yarn`. And in fact, that was a key moment because we were unhappy with `Lerna` and especially "root hoisting", which was a good way to save disk space but a very bad way to manage dependencies. I can remember a few times we were stuck with releases because of `type` or react version mismatches and this long-standing issue with peer dependencies, which was always hard to trace, debug, and fix.

So, considering my old experience with Lerna, I was looking for some solution that could solve such problems out of the box, which is why I decided to propose `pnpm`. One of the key features of `pnpm` is that if you configure it right, then your packages (in the workspace) will be isolated, free of peerDeps issues, and free of "phantom" packages. This was exactly what I was looking for.

After a few experiments, I found that the actual migration from Lerna was much easier than I expected. Few commands, few packages json updates, and we were able to switch to `pnpm` with "great success".

 

## 2. Pnpm: good or bad?

At the beginning, I was very excited about the shiny new package manager. It was like a new car after a very old and, for a very long time, unsupported Lerna. But then the new (technical) issues started popping up here and there. And this was an alarm for me, but...

Before we dive deeper, I just wanted to make sure that you guys understood that I was pushing toward pnpm for a few reasons:

- isolation
- turob
- "since" filters (flags)
- hard linking

Each of these had a big value in my view, and almost none of that was at yarn. Yarn was well known for pushing us toward the PnP package system (With yarn v2), and it is still very popular.

Isolation: as I said, when enabled, packages do not overlap and can install different versions of the same package (like React), and it was working very well.

Turbo is a bit of a different tool; in fact, it is a "build" tool on steroids, but I think adopting turbo to our pipeline is a huge value as it saves us a lot of time on every PR. With Turbo, we were able to reduce the average PR time to 8–10 minutes. This is because of heavy build results caching, which is done by Turbo.

since - this flag was in Lerna, and the big advantage of it was the ability to run commands only in changed packages (in packages that change if you compare with master).

hard linking — it’s not visible very much, but pnpm hard links packages and reduces duplication, which is a good time and disk space saver.

So, at that point, I knew, "We covered, we can move forward with it". And that was true. These minor issues were here and there, but I was able to fix them quickly.

At some point, almost 90% (it felt like) was with explicit packages declared in package.json. That was a moment that I was pushing, and I really hoped that these explicit packages would solve a lot of our issues in the future.

But I was wrong. The main problem that I was not fully aware of was PWT.

The PWT has one interesting property: it’s a package that carries other packages, or, as other people on the internet call it, a metapackage. In other words, PWT made the assumption that you are probably not very advanced in JS (yet), and in fact, it tries to solve as much dependency (and tooling setup) burden as it can for you, so you can spend your time not setting up tooling but writing widget code.

That was a big problem for pnpm and, as a result, for us...

The core issue is that I (and probably most of us) were completely unaware of PWT as a "primary" tool for writing widgets (the platform-recommended way of doing it). And at some point I even tried to "eject" PWT from our repo to actually solve a few issues we had (it was before PWT 9.24.0, and at that time we used outdated Jest, slow Babel, and other minor issues). But then I was pushed back and my PR was closed; this is it.

So, basically, we end up in a state where we have a lot of packages that depend heavily on PTW, but the package manager is not playing nice with PWT. I think this happened because pnpm was the only package manager that was using an isolated approach. Other PMs (npm and yarn) actually used the "de facto" node (or npm 6-7) approach.

And PWT was built with assumptions and was expecting this (npm-ish) approach (when you have flat node_modules at the root of your repo) and even many of its dependencies relying on this (like babel or eslint before v8).

 

So, is pnpm bad? No. Is was a mistake? No. It just wrong thinking on how PWT works. Is it good? Also no. pnpm evolves; sometimes you encounter bugs, sometimes things are not very clear, and probably VSCode is not able to resolve packages correctly. But eventually it will become a big player in JS ecosystem.

## 3. So, what next?

I think we have to take a step back and disable isolation. This will "ease" the strictness and make pnpm behave more like other package managers in the ecosystem. Still, most packages rely on the old "check node_modules; if not found, search for it in the parent directory" strategy. This is the default resolution mechanism. And because PWT expects exactly this environment, we have to provide it. In other words, **PWT should be the main source of truth**. PWT will be the main governor for dependencies. Yeah, there is another big elephant in the room, and his name is ESM (Ecma Script Modules), but for now we can ignore it.

In short. It's a very raw finding, but I found that almost every tool has its own "resolution strategy". It includes rollup, typescript, node, babel, and others. The fun part is that there is no "standard" way of resolving things (yeah, there's CommonJS, but it’s not as strict as you may think). So, many tools decided to implement their own module resolution to solve platform issues and inconsistencies between packages. I think Webpack was one of these. And ESM is the first time in JS history when we have a truly well-defined and "clear" way of declaring and resolving modules.

So, if we talk about ESM, then I should say that we should switch ASAP as Node 18 is already LTS. And, very slowly, most packages migrate to ESM. And we should as well.

 

## 4. Conclusion

As I said, to play nicely with PWT, we have to "ease" the isolation. This way, we ensure that pnpm will be much closer to npm. And as a few of you already mentioned, we have to remove *all* dependencies that are already listed in PWT. This is where this "migration check" from PWT comes in handy, as it will be smashing in your face every time it finds duplicates in your dependencies. So, yes, it may be painful, but we have to remove duplicates and keep our DPEs short.

Last, but not least, all this is to separate concerns: PWT will be responsible for build, test, and dependencies, and we will be free of extra maintenance. Also, there will probably be fewer bugs, and we will have more time to fix and improve code that brings value to our customers.

And yeah, this is just a part of what I learned. There's more stuff, but I don’t think it's worth talking about it here. I just organized my thoughts with text.


gjulivan (approved):

samuelreichert:
Should we maybe also remove the other dependencies we also have on PWT? Like Jest, eslint, prettier, etc.

iobuhov:
> Should we maybe also remove the other dependencies we also have on PWT? Like Jest, eslint, prettier, etc.

I will comments soon as well as notes on what will be our next steps.

PR #592, opened 2023-06-20
[WC-1824]: Dropdown Internal Demo Review

rahmanunver:
### Pull request type

---

### Description
PM review for this widget is in progress, there are number of changes made in 3 rounds of reviews, below are the most important ones:

- Multi Selection added
- Config changes, both in terms of just placement, renaming, and also splitting `attribueEnumerationorBoolean` into `attributeEnumeration` and `attributeBoolean`, also some Review process configs are present to make it easier to test approaches,
- Typings refactored to reduce redundancy.
- Single / Multi selection bug fixes.
 , and more, see [JIRA WC-1824](https://mendix.atlassian.net/browse/WC-1824) for detailed context of the changes. 

If we can properly work with all three types (`association, enumeration, boolean`), do the value changes work properly as we change selection, and being able to navigate and use with keyboard is also important. The Multi Selection works with ReferenceSet values, if needed I'll have a test project posted here  [JIRA WC-1824](https://mendix.atlassian.net/browse/WC-1824). 

### Missing parts
- Styling stopped after functionality and UX concerns raised, so don't expect a styling overhaul in this PR, that part is split into a separate story.
- Config has a section named "Review" which was added to have as an option during the review process, it will be removed later, but feel free to switch the option there to see what it means, it is intended for UI/UX discussions.


packages/pluggableWidgets/combobox-web/package.json
```diff
@@ -49,6 +49,16 @@
...
+    "@testing-library/dom": "^9.3.0",
+    "@testing-library/jest-dom": "^5.16.5",
+    "@testing-library/react": "^14.0.0",
+    "@testing-library/user-event": "^14.4.3",
+    "@types/jest": "^29.0.0",
+    "@types/node": "^18.16.18",
+    "@types/react": "^18.2.14",
+    "@types/react-dom": "18.2.4",
+    "@types/react-test-renderer": "^18.0.0",
+    "@types/testing-library__jest-dom": "^5.14.5",
```
gjulivan, line 61: do we still need to include these?

packages/pluggableWidgets/combobox-web/src/components/Combobox.tsx
```diff
@@ -1,79 +1,98 @@
...
-                selectItem,
                 getInputProps,
                 getItemProps,
                 selectedItem,
                 getMenuProps,
+                clearSelection,
                 isOpen,
                 highlightedIndex,
                 getToggleButtonProps
             }): JSX.Element => (
```
gjulivan, line 46: is it possible to separate this component into another function (return ReactElement) in the same file?

packages/pluggableWidgets/combobox-web/src/components/Combobox.tsx
```diff
@@ -1,79 +1,98 @@
...
+                            "widget-combobox-input-container-disabled": readOnly
                         })}
                         {...getToggleButtonProps({
                             disabled: readOnly
                         })}
                     >
                         <input
-                            id="widget-dropdown-input"
-                            className="widget-dropdown-input"
+                            id="widget-combobox-input"
```
gjulivan, line 60: duplicate ID will exist with multiple combo box.

packages/pluggableWidgets/combobox-web/src/components/Combobox.tsx
```diff
@@ -1,79 +1,98 @@
...
+                        tabIndex={-1}
+                        className={classNames("form-control", "widget-combobox-input-container", {
+                            "widget-combobox-input-container-active": isOpen || focused,
+                            "widget-combobox-input-container-disabled": readOnly
                         })}
                         {...getToggleButtonProps({
                             disabled: readOnly
                         })}
                     >
                         <input
```
gjulivan, line 59: i think aria-label is needed here

packages/pluggableWidgets/combobox-web/src/components/ComboboxMenu.tsx
```diff
@@ -3,38 +3,38 @@ import { PropGetters } from "downshift/typings";
...
     selector: SingleSelector;
     highlightedIndex: number | null;
     selectedItem?: string | null;
 }
 
-export function DropdownMenu(props: DropdownMenuProps): ReactElement {
-    const { dropdownSize, isOpen, selector, highlightedIndex } = props;
+export function ComboboxMenu(props: ComboboxMenuProps): ReactElement {
+    const { comboboxSize, isOpen, selector, highlightedIndex, getMenuProps, getItemProps } = props;
     return createPortal(
```
gjulivan, line 19: avoid createPortal

packages/pluggableWidgets/combobox-web/src/components/ComboboxPreview.tsx
```diff
@@ -2,36 +2,36 @@ import classNames from "classnames";
...
                     <div
                         style={{ backgroundColor }}
-                        className={classNames("form-control", "widget-dropdown-input-container", {
-                            "widget-dropdown-input-container-active": isOpen
+                        className={classNames("form-control", "widget-combobox-input-container", {
+                            "widget-combobox-input-container-active": isOpen
                         })}
                         {...getToggleButtonProps()}
                     >
                         <input
```
gjulivan, line 22: aria-label

packages/pluggableWidgets/combobox-web/src/components/Placeholder.tsx
```diff
@@ -0,0 +1,15 @@
+import { createElement, ReactElement } from "react";
+import { DownArrow } from "../assets/icons";
+
+export function Placeholder(): ReactElement {
+    return (
+        <div className="widget-combobox">
+            <div className="form-control widget-combobox-placeholder">
+                &nbsp;
```
gjulivan, line 8: is using padding not possible?

packages/pluggableWidgets/combobox-web/src/assets/icons.tsx
```diff
@@ -1,16 +1,22 @@
 import { createElement, ReactElement } from "react";
 
-export function ClearButton(): ReactElement {
+export function ClearButton({ size = "10" }): ReactElement {
     return (
-        <svg fill="currentColor" height="10" width="10" viewBox="0 0 329 329">
+        <svg
```
gjulivan, line 5: for customization purposes, wrap the svg in span

packages/pluggableWidgets/combobox-web/src/helpers/types.ts
```diff
@@ -41,3 +41,18 @@ export interface SingleSelector {
     currentValue: string | null;
     setValue(value: string | null): void;
 }
+export interface MultiSelector {
```
gjulivan, line 44: i think singleselector and multiselector type can be merged

packages/pluggableWidgets/combobox-web/src/hooks/useGetSelector.ts
```diff
@@ -0,0 +1,21 @@
+import { useRef } from "react";
+import { getSelector, getMultiSelector } from "../helpers/getSelector";
+import { SingleSelector, MultiSelector } from "../helpers/types";
+import { ComboboxContainerProps } from "../../typings/ComboboxProps";
+
+export function useGetSelector(props: ComboboxContainerProps): SingleSelector {
```
gjulivan, line 6: useGet and useGetMulti is the same thing.

packages/pluggableWidgets/combobox-web/src/helpers/Association/utils.ts
samuelreichert, line 1: @rahmanunver  Here a suggestion to remove some duplication:

```tsx
export function extractAssociationProps(props: ComboboxContainerProps) {
    extractProps(props, "Reference");
}

export function extractMultiAssociationProps(props: ComboboxContainerProps) {
    extractProps(props, "ReferenceSet");
}

function extractProps(props: ComboboxContainerProps, type: "Reference" | "ReferenceSet") {
    const attr = props.attributeAssociation;
    if (!attr) {
        throw new Error("'optionsSourceType' type is 'association' but 'attributeAssociation' is not defined.");
    }
    if (attr.type !== type) {
        throw new Error(`'attributeAssociation' type is not '${type}'.`);
    }

    const ds = props.optionsSourceAssociationDataSource;
    if (!ds) {
        throw new Error(
            "'optionsSourceType' type is 'association' but 'optionsSourceAssociationDataSource' is not defined."
        );
    }
    const captionType = props.optionsSourceAssociationCaptionType;
    const captionAttribute = props.optionsSourceAssociationCaptionAttribute;
    const captionExpression = props.optionsSourceAssociationCaptionExpression;

    if (captionType === "attribute" && !captionAttribute) {
        throw new Error(
            "'optionsSourceAssociationCaptionType' type is 'attribute' but 'optionsSourceAssociationCaptionAttribute' is not defined."
        );
    }
    if (captionType === "expression" && !captionExpression) {
        throw new Error(
            "'optionsSourceAssociationCaptionType' type is 'expression' but 'optionsSourceAssociationCaptionExpression' is not defined."
        );
    }
    const emptyOption = props.emptyOptionText;
    const clearable = props.clearable;

    return [
        attr,
        ds,
        captionType === "attribute" ? captionAttribute! : captionExpression!,
        emptyOption,
        clearable
    ] as const;
}
```

packages/pluggableWidgets/combobox-web/src/ui/Combobox.scss
```diff
@@ -0,0 +1,169 @@
...
+    transition: color 150ms ease 0s;
+    &-menu {
+        position: absolute;
+        display: inline;
+        border-radius: 7px;
+        margin: 4px 0 0 0;
+        left: unset;
+        padding: 0;
+        left: unset;
+        padding: 0;
```
samuelreichert, line 23:
```suggestion
        left: unset;
        padding: 0;
```

packages/pluggableWidgets/combobox-web/src/ui/Combobox.scss
```diff
@@ -0,0 +1,169 @@
...
+            display: flex;
+            border-radius: 2px;
+            padding-left: 4px;
+            padding-right: 4px;
+            flex-wrap: wrap;
+            align-items: center;
+            margin: 1px 2px;
+            &-remove-icon {
+                padding: "4px";
+                cursor: "pointer";
```
samuelreichert, line 165: Is this working? Maybe you can type it without the double quotes

packages/pluggableWidgets/combobox-web/src/ui/Combobox.scss
```diff
@@ -0,0 +1,169 @@
...
+                color: $brand-primary;
+            }
+        }
+        &-listitem {
+            display: flex;
+            border-radius: 2px;
+            padding-left: 4px;
+            padding-right: 4px;
```
samuelreichert, line 159:
```suggestion
            padding: 0 4px;
```

samuelreichert (dismissed):

packages/pluggableWidgets/combobox-web/package.json
rahmanunver: I'm not sure, what do you think @iobuhov

packages/pluggableWidgets/combobox-web/src/components/Combobox.tsx
rahmanunver: Switched to useCombobox like in MultiSelection, this comment is outdated now.

packages/pluggableWidgets/combobox-web/src/components/ComboboxMenu.tsx
rahmanunver: Removed createPortal, this comment is outdated now.

packages/pluggableWidgets/combobox-web/src/components/Placeholder.tsx
rahmanunver: Removed it, we don't need to replace it at all.

packages/pluggableWidgets/combobox-web/src/assets/icons.tsx
rahmanunver: Done.

packages/pluggableWidgets/combobox-web/src/helpers/Association/utils.ts
rahmanunver: Thanks! Made the changes.

packages/pluggableWidgets/combobox-web/src/Combobox.editorConfig.ts
```diff
@@ -1,9 +1,9 @@
 import { StructurePreviewProps } from "@mendix/pluggable-widgets-commons";
 import { hidePropertiesIn, Properties } from "@mendix/pluggable-widgets-tools";
-import { DropdownPreviewProps } from "../typings/DropdownProps";
+import { ComboboxPreviewProps } from "../typings/ComboboxProps";
 
-export function getProperties(values: DropdownPreviewProps, defaultProperties: Properties): Properties {
-    if (values.optionsSourceType === "enumerationOrBoolean") {
+export function getProperties(values: ComboboxPreviewProps, defaultProperties: Properties): Properties {
+    if (values.optionsSourceType === "enumeration" || values.optionsSourceType === "boolean") {
```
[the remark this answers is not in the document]
rahmanunver: Done, thanks!

packages/pluggableWidgets/combobox-web/src/components/MultiSelection/MultiSelection.tsx
```diff
@@ -0,0 +1,141 @@
...
+    const readOnly =
+        (props.attributeBoolean?.readOnly || props.attributeEnumeration?.readOnly) ??
+        props.attributeAssociation?.readOnly;
+
+    if (selector.status === "unavailable") {
+        return <Placeholder />;
+    }
+
+    return (
+        <div className="widget-combobox" {...actionEvents}>
```
gjulivan, line 47: this part here is similar with SingleSelection.
easily moved to Combobox.tsx instead, along with useActionEvents

gjulivan: and SingeSelection + MultiSelection can be wrapped with React.Fragment

gjulivan (approved):

packages/pluggableWidgets/combobox-web/src/Combobox.editorConfig.ts
```diff
@@ -12,22 +12,29 @@ export function getProperties(values: DropdownPreviewProps, defaultProperties: P
...
         ]);
-
-        // clearable is not available in boolean, is it for Enums?
-        hidePropertiesIn(defaultProperties, values, ["clearable"]);
+        if (values.optionsSourceType === "boolean") {
+            hidePropertiesIn(defaultProperties, values, ["clearable"]);
+            hidePropertiesIn(defaultProperties, values, ["attributeEnumeration"]);
+        } else {
+            hidePropertiesIn(defaultProperties, values, ["attributeBoolean"]);
+        }
```
r0b1n, line 20: I see that we keep the clearable option for enums. Is this as intended, do we allow enum attributes to be optionally clearable? @takumagott

packages/pluggableWidgets/datagrid-dropdown-filter-web/src/components/AssociationFilter.tsx
```diff
@@ -37,12 +37,12 @@ function DropdownFooter(props: DropdownFooterProps): JSX.Element | null {
...
 }
 
-interface DropdownProps {
+interface ComboboxProps {
     dispatch: ConditionDispatch;
     widgetProps: DatagridDropdownFilterContainerProps;
     associationProps: AssociationProperties;
 }
-function Dropdown({ dispatch, widgetProps, associationProps }: DropdownProps): ReactElement {
+function Dropdown({ dispatch, widgetProps, associationProps }: ComboboxProps): ReactElement {
```
r0b1n, line 45: Doesn't look related to combo box widget.

packages/pluggableWidgets/combobox-web/src/ui/Combobox.scss
r0b1n: Looks like it is still duplicated

packages/pluggableWidgets/combobox-web/src/ui/Combobox.scss
```diff
@@ -0,0 +1,175 @@
...
+        background-color: $unselected-item;
+        list-style-type: "none";
+        &-hidden {
+            display: none;
+        }
+        &-footer {
+            display: flex;
+            justify-content: space-between;
+            color: #264ae5;
+            border-top: 1px solid #787d87;
```
r0b1n, line 36: We possibly want to use color variables here

packages/pluggableWidgets/combobox-web/src/helpers/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+        } else {
+            return "";
+        }
+    }
+    const selectedItemCaption = selectedItems.map(v => selector.caption.get(v));
+    if (selectedItemCaption.length >= 3) {
+        return selectedItemCaption.join();
+    } else {
+        return selectedItemCaption.slice(0, selectedItemCaption.length).join();
+    }
```
r0b1n, line 20: - I can't quite get what happens in those branches, looks like they are doing the same thing essentially as `selectedItemCaption.slice(0, selectedItemCaption.length)` is not changing the set of elements.
- We probably could check the length on `selectedItems` instead and don't do the unnecessary caption fetching if we gonna trim the list.

packages/pluggableWidgets/combobox-web/src/helpers/types.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    // options related
+    options: OptionsProvider;
+
+    // caption related
+    caption: CaptionsProvider;
+
+    // value related
+    clearable: boolean;
+    currentValue: string[] | undefined;
```
r0b1n, line 76: This is using `undefined` while others use `null`.

packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelection.tsx
```diff
@@ -0,0 +1,85 @@
...
+        getItemProps,
+        selectedItem,
+        getMenuProps,
+        reset,
+        isOpen,
+        highlightedIndex
+    } = useDownshiftSingleSelectProps(selector, setInputValue, inputRef.current, props.onChangeEvent);
+    const readOnly =
+        (props.attributeBoolean?.readOnly || props.attributeEnumeration?.readOnly) ??
+        props.attributeAssociation?.readOnly;
```
r0b1n, line 31: This feels like we have to move this information to the selector and read it from there as `readOnly` field.

packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelection.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { SingleSelectionMenu } from "./SingleSelectionMenu";
+import { Placeholder } from "../Placeholder";
+import { ComboboxWrapper } from "../ComboboxWrapper";
+import classNames from "classnames";
+
+export function SingleSelection(props: ComboboxContainerProps): ReactElement {
+    const comboboxRef = useRef<HTMLDivElement>(null);
+    const inputRef = useRef<HTMLInputElement>(null);
+    const [_inputValue, setInputValue] = useState<string>("");
+    const noFilter = props.filterType === "no";
```
r0b1n, line 16: Same with noFilter, we have to create a field on selector to tell the component if filtering is possible or not. Component should not dig into props to check that.

packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelection.tsx
```diff
@@ -0,0 +1,85 @@
...
+                    tabIndex={0}
+                    ref={inputRef}
+                    {...getInputProps(
+                        {
+                            disabled: readOnly,
+                            readOnly: props.filterType === "no"
+                        },
+                        { suppressRefError: true }
+                    )}
+                    placeholder={selectedItem ? selector.caption.get(selectedItem) : selector.caption.emptyCaption}
```
r0b1n, line 56: If `selectedItem` is not defined, `selector.caption.get(selectedItem)` should give you exactly the `selector.caption.emptyCaption`, so no need to do the logic here, call `selector.caption.get(selectedItem)` directly.

packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelectionMenu.tsx
```diff
@@ -0,0 +1,46 @@
...
+    return (
+        <ul
+            className={classNames("widget-combobox-menu", { "widget-combobox-menu-hidden": !isOpen })}
+            {...getMenuProps?.({}, { suppressRefError: true })}
+            style={{
+                width: comboboxSize?.width
+            }}
+        >
+            {isOpen
+                ? selector.options.getAll(props.filterType).map((item, index) => (
```
r0b1n, line 27: Why do we need to pass this `props.filterType` every time, it should be able to remember it itself in in initialization phase.

packages/pluggableWidgets/combobox-web/src/hooks/useDownshiftMultiSelectProps.ts
```diff
@@ -0,0 +1,153 @@
...
+                        return changes;
+                }
+            },
+            onStateChange({ inputValue, type, selectedItem: newSelectedItems }: UseComboboxStateChangeOptions<string>) {
+                switch (type) {
+                    case useCombobox.stateChangeTypes.InputKeyDownEnter:
+                    case useCombobox.stateChangeTypes.ItemClick:
+                        if (newSelectedItems && !selectedItems.includes(newSelectedItems)) {
+                            setSelectedItems([...selectedItems, newSelectedItems]);
+                            selector.setValue([...selectedItems, newSelectedItems]);
```
r0b1n, line 138: This looks like some redundant bookkeeping, we keep state in two places, the downshift state and in the selector state. This might lead to potential issue with synchronization. 

We should prefer keeping the state in one place. One option is the downshift state (similar to uncontrolled components) where state is synced back to the selector in one place, for example in the `onSelectedItemChange` callback.
Another option is to keep the state in the selector (controlled component), so downshift is getting selection from outside as props and the selector state leads.

packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelection.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { useGetSelector } from "../../hooks/useGetSelector";
+import { SingleSelectionMenu } from "./SingleSelectionMenu";
+import { Placeholder } from "../Placeholder";
+import { ComboboxWrapper } from "../ComboboxWrapper";
+import classNames from "classnames";
+
+export function SingleSelection(props: ComboboxContainerProps): ReactElement {
+    const comboboxRef = useRef<HTMLDivElement>(null);
+    const inputRef = useRef<HTMLInputElement>(null);
+    const [_inputValue, setInputValue] = useState<string>("");
```
r0b1n, line 15: Why is this state entry needed? It is not used anywhere as far as I see.

gjulivan (approved):

packages/pluggableWidgets/combobox-web/src/ui/Combobox.scss
gjulivan: fixed.

gjulivan (approved):

packages/pluggableWidgets/combobox-web/src/helpers/utils.ts
rahmanunver: Agreed that they essentially do the same thing so that needs to be cleaned up, but the reason we are fetching captions is we want to return them as string for the placeholder, I'd say we keep caption fetching here in util, what do you say?

packages/pluggableWidgets/combobox-web/src/helpers/types.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export interface SingleSelectionProps {
+    props: ComboboxContainerProps;
+    selector: SingleSelector;
+}
+
+export interface MultiSelectionProps {
+    props: ComboboxContainerProps;
+    selector: MultiSelector;
+}
```
r0b1n, line 41: Let's not add those props here, those strictly belong to the corresponding components.
I would also suggest trying to merge those like 
```
export interface SingleSelection extend ComboboxContainerProps {
    selector: MultiSelector;
}
```
And use the component as `<SingleSelection {...props} selector={selector} />`

packages/pluggableWidgets/combobox-web/src/components/MultiSelection/MultiSelection.tsx
```diff
@@ -0,0 +1,123 @@
...
+        inputValue,
+        getSelectedItemProps,
+        getDropdownProps,
+        removeSelectedItem,
+        setActiveIndex,
+        selectedItems,
+        items,
+        withCheckbox,
+        setSelectedItems
+    } = useDownshiftMultiSelectProps(selector, props.selectionType, props.onChangeEvent, inputRef.current);
```
r0b1n, line 30: Unless I am missing something, those are the only places actual widget `props` is used, maybe we can get rid of passing props in both, `SingleSelection` and `MultiSelection` components, there instead of passing down the `selectionType` we could use a boolean prop named `withCheckboxes`.

As for the onChangeEvent, we can try to move that one away to the selector. After we aligned the state changes, we now do 
```
            selector.setValue(selectedItems ?? []);
            executeAction(onChangeEvent);
```
in one place, so we can move this action execution to the `selector.setValue` and not do it in the component. 
Same applies for single selection.

packages/pluggableWidgets/combobox-web/src/helpers/Association/utils.ts
```diff
@@ -1,12 +1,16 @@
...
+    const filterType = props.filterType;
+    const onChangeEvent = props.onChangeEvent;
+
     if (!attr) {
         throw new Error("'optionsSourceType' type is 'association' but 'attributeAssociation' is not defined.");
     }
-    if (attr.type !== "Reference") {
-        throw new Error("'attributeAssociation' type is not 'Reference'.");
+    if (attr.type !== type) {
+        throw new Error(`'attributeAssociation' type is not '${type}'.`);
```
r0b1n, line 13: This looks a bit off. On line 4 we are reading the `type` from `props.attributeAssociation`, after that, we assign `props.attributeAssociation` to `attr` and then on line 12 we compare `type` with `attr.type` which is basically comparing with itself.

packages/pluggableWidgets/combobox-web/src/helpers/EnumBool/EnumAndBooleanSimpleCaptionsProvider.tsx
```diff
@@ -3,16 +3,16 @@ import { createElement, ReactNode } from "react";
...
 
 export class EnumAndBooleanSimpleCaptionsProvider implements CaptionsProvider {
     private attr?: EditableValue<string | boolean>;
-    private emptyCaption = "";
-    updateProps(props: EnumAndBooleanSimpleCaptionsProviderProps) {
+    emptyCaption = "";
+    updateProps(props: EnumAndBooleanSimpleCaptionsProviderProps): void {
         this.attr = props.attribute;
-        if (props.emptyOptionText.status === "unavailable") {
+        if (props.emptyOptionText?.status === "unavailable" || !props.emptyOptionText) {
```
r0b1n, line 15: Move `!props.emptyOptionText` to the beginning of the expression and get rid of `?`

packages/pluggableWidgets/combobox-web/src/helpers/types.ts
```diff
@@ -1,20 +1,24 @@
...
+    emptyCaption: string;
 }
 
 export interface OptionsProvider<T = unknown, P = {}> {
     status: Status;
+    filterType: FilterTypeEnum;
+    searchTerm: string;
 
-    getAll(sortType: TypeaheadEnum): string[];
+    getAll(searchTerm?: string): string[];
```
r0b1n, line 17: We don't pass anything to `getAll()` anymore.

packages/pluggableWidgets/combobox-web/src/helpers/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+    if (!selector.withCheckbox) {
+        if (selectedItems.length === 0) {
+            return selector.caption.emptyCaption;
+        } else {
+            return "";
+        }
+    }
+    const selectedItemCaption = selectedItems.map(v => selector.caption.get(v));
+
+    return selectedItemCaption.join();
```
r0b1n, line 13: By default it puts no space between characters, so it is `Option1,Option2,Option3`, let's do `Option1, Option2, Option3` instead by using `.join(", ")`.

packages/pluggableWidgets/combobox-web/src/hooks/useDownshiftMultiSelectProps.ts
```diff
@@ -0,0 +1,129 @@
...
+    UseComboboxProps,
+    UseComboboxState,
+    UseComboboxStateChangeOptions,
+    useCombobox,
+    useMultipleSelection
+} from "downshift";
+import { useMemo } from "react";
+import { MultiSelector } from "../helpers/types";
+
+export function useDownshiftMultiSelectProps(selector: MultiSelector, inputElement: HTMLInputElement | null) {
```
r0b1n, line 11: Looks like the `inputElement` is not used anywhere, we can omit it.

packages/pluggableWidgets/combobox-web/src/hooks/useDownshiftMultiSelectProps.ts
```diff
@@ -0,0 +1,129 @@
...
+                    setSelectedItems(newSelectedItems!);
+                    break;
+                default:
+                    break;
+            }
+        }
+    });
+    const filteredItems = selector.options?.getAll().filter(option => !selectedItems.includes(option));
+
+    const items = selector.withCheckbox ? selector.options?.getAll() : filteredItems;
```
r0b1n, line 40: Let's first do the check if we need to filter. And then do the filtering if needed. Likely the `?` after `options` is redundant.

packages/pluggableWidgets/combobox-web/src/components/MultiSelection/MultiSelection.tsx
```diff
@@ -0,0 +1,121 @@
...
+                                setSelectedItems([]);
+                            }
+                        }}
+                    >
+                        <ClearButton />
+                    </button>
+                )}
+            </ComboboxWrapper>
+            <MultiSelectionMenu
+                comboboxSize={comboboxRef.current?.getBoundingClientRect()}
```
r0b1n, line 109: As we are not rendering in a portal anymore, we might not need this sizing calculation? Worth checking if we can rely on the parent container for the size, using only CSS. Same for SingleSelectionMenu.

packages/pluggableWidgets/combobox-web/src/Combobox.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+    return (
+        <div className="widget-combobox" {...actionEvents}>
+            {selector.type === "single" ? (
+                <SingleSelection selector={selector} />
+            ) : (
+                <MultiSelection selector={selector} />
+            )}
+        </div>
+    );
```
r0b1n, line 22: We probably want to apply default props that are coming from widget as: `style`, `class`, `tabIndex` (if it is not, we have to enable it), so users in SP can apply styling, design properties, etc.

rahmanunver: Do you think we need to apply those to both components or just their container (".widget-combobox")?

packages/pluggableWidgets/combobox-web/src/helpers/Association/utils.ts
rahmanunver: We can get rid of that check completely as we are already splitting `Reference` and `ReferenceSet` types before it comes to this check, I don't see a case where it will hit an error because of association type here.

packages/pluggableWidgets/combobox-web/src/components/MultiSelection/MultiSelection.tsx
rahmanunver: Parent containers `.widget-combobox` size (width) is not specified, therefore we can't inherit it directly.

packages/pluggableWidgets/combobox-web/src/Combobox.tsx
iobuhov: No, this should go only to root (`.widget-combobox`)

packages/pluggableWidgets/combobox-web/src/components/MultiSelection/MultiSelectionMenu.tsx
```diff
@@ -0,0 +1,88 @@
...
+    selector,
+    selectableItems,
+    selectedItems,
+    setSelectedItems
+}: MultiSelectionMenuProps): ReactElement {
+    const allSelected = selector.options.getAll().length === selectedItems.length;
+    const noneSelected = selectedItems.length < 1;
+    return (
+        <div className={classNames("widget-combobox-menu", { "widget-combobox-menu-hidden": !isOpen })}>
+            <ul style={{ padding: 0 }} {...getMenuProps?.({}, { suppressRefError: true })}>
```
r0b1n, line 29: Can we move inline style it to the styling files?

packages/pluggableWidgets/combobox-web/src/components/MultiSelection/MultiSelectionMenu.tsx
```diff
@@ -0,0 +1,88 @@
...
+                {isOpen &&
+                    selectableItems.map((item, index) => (
+                        <li
+                            className={classNames("widget-combobox-item", {
+                                "widget-combobox-item-highlighted": highlightedIndex === index
+                            })}
+                            key={`${item}${index}`}
```
r0b1n, line 36: Why do we need to include index as part of the key?

packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelectionMenu.tsx
```diff
@@ -0,0 +1,45 @@
...
+            {...getMenuProps?.({}, { suppressRefError: true })}
+        >
+            {isOpen
+                ? selector.options.getAll().map((item, index) => (
+                      <li
+                          className={classNames("widget-combobox-item", {
+                              "widget-combobox-item-selected": selector.currentValue === item,
+                              "widget-combobox-item-highlighted": highlightedIndex === index
+                          })}
+                          key={JSON.stringify(item)}
```
r0b1n, line 32: Why do we need `JSON.stringify` ?

packages/pluggableWidgets/combobox-web/src/components/MultiSelection/MultiSelection.tsx
```diff
@@ -0,0 +1,112 @@
...
+    return (
+        <Fragment>
+            <ComboboxWrapper isOpen={isOpen} readOnly={selector.readOnly} getToggleButtonProps={getToggleButtonProps}>
+                <div className="widget-combobox-selected-items">
+                    {!selector.withCheckbox &&
+                        selectedItems.map((selectedItemForRender, index) => {
+                            return (
+                                <span
+                                    className="widget-combobox-selected-item"
+                                    key={`selected-item-${index}`}
```
r0b1n, line 42: We should not use index as key.

packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelection.tsx
```diff
@@ -0,0 +1,68 @@
...
+                    className={classNames("widget-combobox-input", {
+                        "widget-combobox-input-nofilter": selector.options.filterType === "no"
+                    })}
+                    tabIndex={0}
+                    {...getInputProps(
+                        {
+                            disabled: selector.readOnly,
+                            readOnly: selector.options.filterType === "no"
+                        },
+                        { suppressRefError: true }
```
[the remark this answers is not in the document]
rahmanunver: I'd say we should keep them for now as it is triggering an error in Downshift due to our structure, Downshift doesn't like that we are early returning a `<Placeholder>` element, there are ref error but also errors from unused Downshift prop methods, which can be part of other upcoming stories, we'll have to restructure the component to fix this.

packages/pluggableWidgets/combobox-web/src/components/MultiSelection/MultiSelectionMenu.tsx
rahmanunver: done

packages/pluggableWidgets/combobox-web/src/components/MultiSelection/MultiSelectionMenu.tsx
rahmanunver: done

packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelectionMenu.tsx
rahmanunver: No need, it's a leftover.

packages/pluggableWidgets/combobox-web/src/components/MultiSelection/MultiSelection.tsx
rahmanunver: Switched to use the caption of the element.

r0b1n: Using item itself should be probably enough, also no need for `selected-item-` prefix.

packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelectionMenu.tsx
r0b1n: It should go then, item as key should be enough.

rahmanunver: done

packages/pluggableWidgets/combobox-web/src/components/MultiSelection/MultiSelection.tsx
rahmanunver: done

leonardomendix (dismissed):

leonardomendix (approved):

gjulivan (approved):

PR #597, opened 2023-06-22
fix: upgrade swiper package in carousel

gjulivan:
Bug fix on Carousel by upgrading swiper package from 6 to 9.

fix a11y issue.
some DOM changes is occurs.

### DIFF
--- a/packages/pluggableWidgets/carousel-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/carousel-web/CHANGELOG.md
@@ -6,6 +6,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Changed
+
+-   We updated swiper package version to 9.4
+-   DOM changes from <div> wrapper to <ul> and <li>
+
+### Fixed
+
+-   We fixed accessibility issue with carousel
+
 ## [2.1.0] - 2023-06-05
 
 ### Changed
--- a/packages/pluggableWidgets/carousel-web/src/Carousel.editorPreview.tsx
+++ b/packages/pluggableWidgets/carousel-web/src/Carousel.editorPreview.tsx
@@ -10,6 +10,7 @@ export function getPreviewCss(): string {
 export function preview(props: CarouselPreviewProps): ReactNode {
     return (
         <Carousel
+            id={"Carousel"}
             className={props.className}
             navigation={props.navigation}
             pagination={props.showPagination}
--- a/packages/pluggableWidgets/carousel-web/src/Carousel.editorPreview.tsx
+++ b/packages/pluggableWidgets/carousel-web/src/Carousel.editorPreview.tsx
@@ -1,31 +1,33 @@
-import { createElement, ReactNode, useId } from "react";
+import { createElement, ReactElement } from "react";
 import { CarouselPreviewProps } from "../typings/CarouselProps";
 import { Carousel } from "./components/Carousel";
 import { GUID } from "mendix";
+import { generateUUID } from "@mendix/pluggable-widgets-commons/components/web";
 
 export function getPreviewCss(): string {
     return require("./ui/Carousel.scss");
 }
 
-export function preview(props: CarouselPreviewProps): ReactNode {
-    const id = useId();
+export function CarouselPreviewComponent(props: CarouselPreviewProps): ReactElement {
     return (
         <Carousel
-            id={id}
+            id={generateUUID().toString()}
             className={props.className}
             navigation={props.navigation}
             pagination={props.showPagination}
             loop={false}
-            items={[
-                {
-                    id: "1" as GUID,
-                    content: (
-                        <props.content.renderer caption="Carousel item content">
-                            <div />
-                        </props.content.renderer>
-                    )
-                }
-            ]}
+            items={["1", "2", "3"].map(item => ({
+                id: item as GUID,
+                content: (
+                    <props.content.renderer caption={`Carousel item content ${item}`}>
+                        <div>{`Carousel item content ${item}`}</div>
+                    </props.content.renderer>
+                )
+            }))}
         />
     );
+}
+
+export function preview(props: CarouselPreviewProps): ReactElement {
+    return <CarouselPreviewComponent {...props} />;
 }
--- a/packages/pluggableWidgets/carousel-web/src/Carousel.editorPreview.tsx
+++ b/packages/pluggableWidgets/carousel-web/src/Carousel.editorPreview.tsx
@@ -1,4 +1,4 @@
-import { createElement, ReactNode } from "react";
+import { createElement, ReactNode, useId } from "react";
 import { CarouselPreviewProps } from "../typings/CarouselProps";
 import { Carousel } from "./components/Carousel";
 import { GUID } from "mendix";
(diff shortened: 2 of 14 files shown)


samuelreichert (approved):

packages/pluggableWidgets/carousel-web/CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Changed
+
+-   We updated swiper package version to 9.4
```
r0b1n, line 11: We probably want to mention that there is a noticeable DOM change. Do we want to bump major version, btw, maybe it is worth it.

gjulivan: agree

gjulivan: as far as i know,
DOM changes without feature improvement can only be considered as minor version. (similar to pop up menu)

packages/pluggableWidgets/carousel-web/src/components/Carousel.tsx
```diff
@@ -20,25 +21,36 @@ export interface CarouselProps {
...
+        }),
+        modules: [A11y],
+        a11y: {
+            enabled: true
+        }
     };
     return (
         <div className={classNames(className, "widget-carousel")} tabIndex={tabIndex}>
-            <SwiperReact {...options} onClick={onClick}>
+            <ReactSwiper tag={"ul"} {...options} onClick={onClick}>
```
r0b1n, line 47: Looking at the snapshots, this gives the structure that is not quite correct, `ul` should not have `div`s as it's children. I guess it is better to render `.swiper-wrapper` as `ul` and keep `.swiper` as `div`.

```
<ul className="swiper">
    <div className="swiper-wrapper">
        <li className="swiper-slide"><div /></li>
        <li className="swiper-slide"><div /></li>
    </div>
    <div className="swiper-button-prev"/>
    <div className="swiper-button-next"/>
    <div className="swiper-pagination"/>
  </ul>
```

gjulivan: fixed

packages/pluggableWidgets/carousel-web/src/components/Carousel.tsx
```diff
@@ -20,25 +21,36 @@ export interface CarouselProps {
 export function Carousel(props: CarouselProps): ReactElement {
     const { items, pagination, loop, animation, autoplay, delay, navigation, className, tabIndex, onClick } = props;
 
+    const paginationOptions: PaginationOptions | boolean = pagination && {
+        type: "bullets",
+        clickable: true,
+        renderBullet: (index, className) => `<span aria-controls="ctl_${items[index]?.id}" class="${className}"></span>`
```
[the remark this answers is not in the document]
r0b1n: Probably a aria-label as well

packages/pluggableWidgets/carousel-web/src/components/Carousel.tsx
```diff
@@ -20,25 +21,36 @@ export interface CarouselProps {
...
+            enabled: true
+        }
     };
     return (
         <div className={classNames(className, "widget-carousel")} tabIndex={tabIndex}>
-            <SwiperReact {...options} onClick={onClick}>
+            <ReactSwiper tag={"ul"} {...options} onClick={onClick}>
                 {items?.map(item => (
-                    <SwiperSlide key={item.id}>{item.content}</SwiperSlide>
+                    <SwiperSlide tag={"li"} key={item.id} id={`ctl_${item.id}`}>
```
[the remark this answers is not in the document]
gjulivan: resolved

packages/pluggableWidgets/carousel-web/src/components/Carousel.tsx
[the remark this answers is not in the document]
gjulivan: tabIndex is automatically added by a11y module

samuelreichert (dismissed):

samuelreichert (dismissed):

packages/pluggableWidgets/carousel-web/src/components/Carousel.tsx
```diff
@@ -14,31 +21,61 @@ export interface CarouselProps {
...
-    items: Array<{ id: GUID; content?: ReactNode }>;
+    items: CarouselItem[];
 }
 
 export function Carousel(props: CarouselProps): ReactElement {
-    const { items, pagination, loop, animation, autoplay, delay, navigation, className, tabIndex, onClick } = props;
+    const { items, pagination, loop, animation, autoplay, delay, navigation, className, tabIndex, id, onClick } = props;
+    const [swiperRef, setSwiperRef] = useState<SwiperClass>();
+
+    const generateSliderId = useCallback(
```
r0b1n, line 31: I think something like `getSlideId` is a better name for this callback. I would argue that this function doesn't generate id in a sense of creating something new, also it does it for one slide, not the whole slider.

packages/pluggableWidgets/carousel-web/src/Carousel.editorPreview.tsx
```diff
@@ -10,6 +10,7 @@ export function getPreviewCss(): string {
 export function preview(props: CarouselPreviewProps): ReactNode {
     return (
         <Carousel
+            id={"Carousel"}
```
r0b1n, line 13: This possibly might cause id clashing as well in design mode, I would stay safe and use unique id here as well.

packages/pluggableWidgets/carousel-web/src/components/Carousel.tsx
```diff
@@ -14,31 +21,61 @@ export interface CarouselProps {
...
     return (
         <div className={classNames(className, "widget-carousel")} tabIndex={tabIndex}>
-            <SwiperReact {...options} onClick={onClick}>
-                {items?.map(item => (
-                    <SwiperSlide key={item.id}>{item.content}</SwiperSlide>
+            <ReactSwiper wrapperTag={"ul"} {...options} onClick={onClick} onSwiper={setSwiperRef}>
+                {items?.map((item, index) => (
+                    <SwiperSlide
+                        tag={"li"}
+                        aria-hidden={swiperRef?.activeIndex === index}
```
r0b1n, line 71: So, the slide **is hidden** (`aria-hidden={true}`) when it **is active**. Is it as intended?

gjulivan: i was surprise by this condition too, 
but the result was the other way around.
@leonardomendix might be able to confirm this.

packages/pluggableWidgets/carousel-web/src/Carousel.editorPreview.tsx
gjulivan: i think accordion was using the same way.
but nvm, this is fixed using useId

packages/pluggableWidgets/carousel-web/src/components/Carousel.tsx
gjulivan: fixed

r0b1n (approved):

leonardomendix (approved):

PR #608, opened 2023-07-06
[WC-1914]: Fix tabIndex image-web

HedwigAR:
### Pull request type

---

### Description

Adding the tabindex as a value to both ContentIcon and ContentImage, the change was not applied to BackgroundImage since that shouldn't have to be able to be tabbed to. Previously this was always set to 0.

There's a test project available which indicated the issue, with the new version of the widget, this should tab over the images correctly. By using `Image type` 
- Image -> Background image No
- Icon 
- Image -> Background image Yes

This should cover the scenario's, we're expecting the first two to be effected by the tab order, but not the last one.

### DIFF
--- a/packages/pluggableWidgets/image-web/src/components/Image/ui.tsx
+++ b/packages/pluggableWidgets/image-web/src/components/Image/ui.tsx
@@ -26,6 +26,7 @@ export interface ImageWrapperProps {
 
 export interface ImageContentProps {
     style?: CSSProperties;
+    tabIndex?: number;
     onClick?: ReactEventHandler<HTMLElement>;
     altText?: string;
 }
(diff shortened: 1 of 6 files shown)


packages/pluggableWidgets/image-web/src/components/Image/ui.tsx
```diff
@@ -104,7 +107,6 @@ function getImageContentOnClickProps(onClick: ImageContentProps["onClick"]): HTM
     return {
         onClick,
         role: "button",
-        tabIndex: 0,
```
r0b1n, line 107: Assigning tab index should not happen unconditionally, as you see here it is only happening when `onClick` is set. It think we should keep this and assign the passed tabIndex instead of zero.

HedwigAR: Thanks, I moved it. Let me know if this is right

r0b1n (approved):

leonardomendix (approved):

PR #620, opened 2023-07-13
[WC-1953] Enable a11y testing with cypress-axe plugin

leonardomendix:
### Pull request type

---

### Description

Enables the usage of the cypress-axe plugin to check a11y violations during the e2e testing phase. That's the first widget since we fixed the a11y issues recently, the idea is to spread the usage through all widgets with e2e automation enabled.

Automation should pass.


iobuhov (approved):
LGTM

gjulivan (approved):

PR #626, opened 2023-07-19
[WC-1865] Charts new param

iobuhov:
### Pull request type

---

### Description
Fix on click action for charts by adding action parameter.

Charts except heatmap and pie chart. To test this bugfix you need create chart and configure on click action with entity parameter to datasource. Clicking on chart "point" should call nanoflow/microflow with entity used to create that "point".

### DIFF
--- a/packages/pluggableWidgets/column-chart-web/src/ColumnChart.xml
+++ b/packages/pluggableWidgets/column-chart-web/src/ColumnChart.xml
@@ -131,7 +131,11 @@
                             </property>
                         </propertyGroup>
                         <propertyGroup caption="Events">
-                            <property key="onClickAction" type="action" required="false">
+                            <property key="onClickActionStatic" type="action" required="false" dataSource="staticDataSource">
+                                <caption>On click action</caption>
+                                <description />
+                            </property>
+                            <property key="onClickActionDynamic" type="action" required="false" dataSource="dynamicDataSource">
                                 <caption>On click action</caption>
                                 <description />
                             </property>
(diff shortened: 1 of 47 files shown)


packages/pluggableWidgets/column-chart-web/src/ColumnChart.xml
```diff
@@ -131,7 +131,11 @@
                             </property>
                         </propertyGroup>
                         <propertyGroup caption="Events">
-                            <property key="onClickAction" type="action" required="false">
+                            <property key="onClickActionStatic" type="action" required="false" dataSource="staticDataSource">
+                                <caption>On click action</caption>
+                                <description />
+                            </property>
+                            <property key="onClickActionDynamic" type="action" required="false" dataSource="dynamicDataSource">
```
r0b1n, line 138: Looking at names of other props, those prepend `static/dynamic`. Would `staticOnClickAction` `dynamicOnClickAction` be more consistent with those? Also we likely should conditionally show/hide one or the other field based on the `dataSet` enum in the editorConfig file.

rahmanunver (dismissed):

packages/pluggableWidgets/column-chart-web/src/ColumnChart.xml
iobuhov: when property linked to `datasource` SP take care of that property and hides it when datasource hidden as well

iobuhov: Done. Properties renamed

gjulivan (approved):

leonardomendix (approved):

PR #639, opened 2023-08-01
feat(image-web): add structure mode previews & fixes

takumagott:
### Features/fixes
* Added structure mode preview for static and dynamic images (the latter only if they have a static default image)
* Improved the captions in the Page Explorer
* fixed an issue where the size of icons from the new icon collections was not configurable
* fixed an issue where the image would take too much space when placed outside of containers

### What to test
**Structure preview**
* If you select a static image, it should show in structure mode (Note: in <10.2 it will always default to size 100*100px because of API limitations)
* If you selecta dynamic image and set a static image as a default, it should show the default in structure mode
* Any other selection should render the original fallback image
* Setting a size in pixels should now affect the rendering size of the image on the canvas (if only a width or height is set, it should set the other proportionally)

**Page explorer captions**
* Static images should show the image path
* Dynamic images should show the entity
* Icons should either show the image path (image icons) or icon class (not ideal but best we can do rn)
* ImageUrls should show Image Urls
* If unset, it should fall back to "(none)"

**Fixes**
* Ensure you can set an icon size for the new Atlas icons (and that it reflects in design mode)
* Ensure that an image (e.g. static image) does not take more height than necessary when put at the bottom of a page (outside of layouts/containers. It should behave the same as core images.

### DIFF
--- a/packages/pluggableWidgets/image-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/image-web/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+-   We improved the structure mode preview for static and dynamic images
+-   We improved the captions in the page explorer
+-   We fixed an issue where the size of icons from the new icon collections was not configurable
+-   We fixed an issue where the image would take too much space when placed outside of containers
+
 ## [1.4.2] - 2023-09-27
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


iobuhov (approved):

iobuhov (approved):

iobuhov (approved):

gjulivan:
https://docs.mendix.com/apidocs-mxsdk/apidocs/pluggable-widgets-config-api/
this is not in the image-web repo yet.
thus, causing build failing.

gjulivan:
> https://docs.mendix.com/apidocs-mxsdk/apidocs/pluggable-widgets-config-api/#71-image ` property?: object; // widget image property object from Values API` this is not in the image-web repo yet. thus, causing build failing.

gjulivan (approved):

PR #644, opened 2023-08-04
fix(tree-node): refresh when children updated

gjulivan:
### Pull request type

Bug fix: tree node not re-rendered when number of children updated

### Description

updated "isActualLeafNode" variables when number of children updated.

### DIFF
--- a/packages/pluggableWidgets/tree-node-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/tree-node-web/CHANGELOG.md
@@ -8,8 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
--   We fixed Atlas icon not displaying properly.
--   We fixed treenode children not refreshed when updated
+-   We fixed Atlas icon unable to be shown on tree node.
+-   We fixed the condition where tree child not directly refreshed after updating the data.
 
 ## [1.1.1] - 2023-05-26
 
--- a/packages/pluggableWidgets/tree-node-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/tree-node-web/CHANGELOG.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Fixed
 
 -   We fixed Atlas icon unable to be shown on tree node.
--   We fixed the condition where tree child not directly refreshed after updating the data.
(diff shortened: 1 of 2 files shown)


packages/pluggableWidgets/tree-node-web/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Fixed
 
 -   We fixed Atlas icon not displaying properly.
+-   We fixed treenode children not refreshed when updated
```
r0b1n, line 12: It would be nice to make those two changelog entries have more clarity for our users. They are a bit too technical and lack grammatical correctness. Focus more on the user experience and the functionality that was fixed, rather than just technical details. Use terms that our users see in the widget's settings and documentation. This would help our users better understand the impact of the changes we've made.

gjulivan: fixed

packages/pluggableWidgets/tree-node-web/src/components/TreeNode.tsx
```diff
@@ -176,6 +176,10 @@ function TreeNodeBranch(props: TreeNodeBranchProps): ReactElement {
     );
     const [isActualLeafNode, setIsActualLeafNode] = useState<boolean>(props.isUserDefinedLeafNode || !props.children);
 
+    useEffect(() => {
+        setIsActualLeafNode(props.isUserDefinedLeafNode || !props.children);
+    }, [props.children]);
```
r0b1n, line 181: This `useEffect` triggers an extra re-render which might result in a UI delay. Even though the `children` prop changes, the current render still uses "incorrect" state and element still renders without children. Next, this effect updates the state to reflect the presence of `children`, triggering a second re-render with the "correct" state.

Is this additional re-render intentional, perhaps to work around some specific issue? If not, consider removing this effect and checking existence of `children` directly, rather than storing this information in the state.

gjulivan: i've made checkings here.
it is actually quite complicated to update directly when children updated with the current treenode structure.

first, the current treenodebranch has to render as not-a-leaf,
then the treenodebranch now able to perform as treenode-body,
then able to use useInformParentContextOfChildNodes (which again, double render)

to change this into correct structure will require more effort and out of scope for this bug fix.

r0b1n (approved):

PR #647, opened 2023-08-07
Wc 1980 dg2 right border issue

gjulivan:
### Pull request type

Bug fix: right border not shown if hiding column capabilities disabled.

### Description

Try select "no" on hiding column in personalization tab on Datagrid 2.
previously, right border is missing.
this PR will fix it.

### DIFF
--- a/packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid-design-properties.scss
+++ b/packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid-design-properties.scss
@@ -28,35 +28,33 @@
 }
 
 .table-bordered-all {
-    .th {
-        border-top-width: 1px;
-        border-top-style: solid;
+    .th,
+    .td {
         border-left-width: 1px;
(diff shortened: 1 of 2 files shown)


packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid-design-properties.scss
```diff
@@ -58,6 +63,11 @@
             border-right-width: 1px;
             border-right-style: solid;
         }
+
+        &:last-child {
+            border-right-width: 1px;
+            border-right-style: solid;
+        }
```
r0b1n, line 70: Those styles look the same, can we get rid of the duplication?

gjulivan: fixed.

rahmanunver (approved):

leonardomendix (approved):

PR #653, opened 2023-08-10
[WC-1598]: DataGrid2 Export API

samuelreichert:
### Pull request type

---

### Description

Creating an export API to handle any kind of export on Datagrid2, be it JSON, SQL, CSV, XLS, ...

### DIFF
--- a/packages/pluggableWidgets/datagrid-web/src/Datagrid.tsx
+++ b/packages/pluggableWidgets/datagrid-web/src/Datagrid.tsx
@@ -248,7 +248,7 @@ function transformColumnProps(props: ColumnsType[]): TableColumn[] {
     }));
 }
 
-function dumpDataFromDataGridToConsole(name: DataGridName) {
+function dumpDataFromDataGridToConsole(name: DataGridName): void {
     console.info("You asked to dump data from: ", name);
     const stream = window[DATAGRID_DATA_EXPORT][name].create();
     console.info("Got data export stream instance, attaching processing function");
--- a/packages/pluggableWidgets/datagrid-web/src/Datagrid.tsx
+++ b/packages/pluggableWidgets/datagrid-web/src/Datagrid.tsx
@@ -1,5 +1,6 @@
 import { createElement, ReactElement, ReactNode, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { ColumnsType, DatagridContainerProps } from "../typings/DatagridProps";
+import { ObjectItem } from "mendix";
 import { FilterCondition } from "mendix/filters";
 import { and } from "mendix/filters/builders";
 import { Table, TableColumn, SortProperty } from "./components/Table";
@@ -21,8 +22,7 @@ import { useCellRenderer } from "./features/cell";
 import { getColumnAssociationProps, isSortable } from "./features/column";
 import { selectionSettings, useOnSelectProps } from "./features/selection";
 import "./ui/Datagrid.scss";
-import { Message, useDG2ExportApi } from "./features/export";
-import { DataGridName, DATAGRID_DATA_EXPORT } from "typings/global";
+import { DATAGRID_DATA_EXPORT, DataGridName, Message, useDG2ExportApi } from "./features/export";
 
 export default function Datagrid(props: DatagridContainerProps): ReactElement {
     const id = useRef(`DataGrid${generateUUID()}`);
@@ -38,7 +38,9 @@ export default function Datagrid(props: DatagridContainerProps): ReactElement {
     const { FilterContext } = useFilterContext();
     const SelectionContext = getGlobalSelectionContext();
     const cellRenderer = useCellRenderer({ columns: props.columns, onClick: props.onClick });
-    useDG2ExportApi({
+    const [memoizedItems, setItems] = useState<ObjectItem[]>([]);
+
+    const { started } = useDG2ExportApi({
         columns: props.columns,
         datasource: props.datasource,
         name: props.name,
@@ -46,9 +48,15 @@ export default function Datagrid(props: DatagridContainerProps): ReactElement {
     });
 
     useEffect(() => {
+        if (memoizedItems.length === 0 && props.datasource.items) {
+            setItems(props.datasource.items);
+        }
+    }, [props.datasource.items]);
+
+    useEffect(() => {
         setTimeout(() => {
             dumpDataFromDataGridToConsole(props.name);
-        }, 500);
+        }, 50);
     }, []);
 
     useEffect(() => {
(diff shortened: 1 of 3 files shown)


packages/pluggableWidgets/datagrid-web/src/Datagrid.tsx
```diff
@@ -58,6 +62,18 @@ export default function Datagrid(props: DatagridContainerProps): ReactElement {
         }
     }, [props.datasource, props.refreshInterval]);
 
+    useEffect(() => {
+        if (props.columns) {
+            exportFromDatagrid.addColumns(props.columns);
+        }
+    }, [props.columns]);
```
iobuhov, line 69: At which circumstances number of columns suppose to change?

packages/pluggableWidgets/datagrid-web/src/Datagrid.tsx
```diff
@@ -36,6 +38,26 @@ export default function Datagrid(props: DatagridContainerProps): ReactElement {
...
+        datasource: props.datasource,
+        name: props.name,
+        pageSize: props.pageSize
+    });
+
+    useEffect(() => {
+        if (memoizedItems.length === 0 && props.datasource.items) {
+            setItems(props.datasource.items);
+        }
+    }, [props.datasource.items]);
```
r0b1n, line 54: Let's hide the memoization inside of the `useDG2ExportApi` hook, let it return `digested` items (or the whole datasource), so there is not explicit state/logic in this component.

packages/pluggableWidgets/datagrid-web/src/features/export.ts
```diff
@@ -0,0 +1,139 @@
...
+        start: () => setStartProcess(true),
+        abort: () => {
+            if (callback) {
+                callback({ type: "end" });
+            }
+            if (startProcess) {
+                setStartProcess(false);
+            }
+        }
+    };
```
r0b1n, line 73: Let's create the stream only on request, we should be as lightweight as possible if export was not requested.

packages/pluggableWidgets/datagrid-web/src/features/export.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    useEffect(() => {
+        if (startProcess && callback) {
+            if (sentColumns === false) {
+                callback(exportColumns());
+                setSentColumns(true);
+            }
+
+            if (datasource.items && datasource.hasMoreItems) {
+                callback(exportData(datasource.items));
```
r0b1n, line 124: We should wait for a promise returned from the `callback` and proceed to the next iteration when it is resolved, so it implements the flow control. No next set should be sent before the receiving part is "ready for it".

packages/pluggableWidgets/datagrid-web/src/features/export.ts
samuelreichert: So, I dont need to send the  "end" message on abort?

packages/pluggableWidgets/datagrid-web/src/Datagrid.tsx
```diff
@@ -37,6 +38,17 @@ export default function Datagrid(props: DatagridContainerProps): ReactElement {
     const SelectionContext = getGlobalSelectionContext();
     const cellRenderer = useCellRenderer({ columns: props.columns, onClick: props.onClick });
 
+    const { items } = useDG2ExportApi({
+        columns: props.columns,
+        datasource: props.datasource,
+        name: props.name,
```
iobuhov, line 44: I'm afraid we can't use the `name` property this way. The `name` prop is not unique (Jelmer confirmed this one when I tried to implement the "Clear filters" feature).

iobuhov: If we need a truly unique value, that should be an expression configured through widget settings, and it will be up to the developer to configure it. This is probably a good topic to discuss at the tech support meeting.

packages/pluggableWidgets/datagrid-web/src/features/export.ts
r0b1n: No, that is not what I meant. At the moment we [re-]create `dataExportStream` even though the export was not requested (`create()` was not called). So every re-render we create another instance of `dataExportStream`. It should be created only when `create()` is called, this hook should be as lightweight as possible.

Same for `exportColumns` and `exportData` functions, they should be separated from the hook, no need for them to be created every time.

The hook should run the minimum amount of code possible if export is not requested.

packages/pluggableWidgets/datagrid-web/src/features/export.ts
```diff
@@ -0,0 +1,160 @@
...
+        if (startProcess && callback) {
+            if (sentColumns === false) {
+                runColumnsCallback();
+                setSentColumns(true);
+            }
+
+            if (datasource.items && datasource.hasMoreItems) {
+                runDataCallback();
+                const newOffset = datasource.offset + pageSize;
+                datasource.setOffset(newOffset);
```
r0b1n, line 147: I expect that this doesn't work as expected because this still doesn't wait for `runDataCallback` to finish before requesting the next set of items.

r0b1n: Same for `runColumnsCallback` above.

packages/pluggableWidgets/datagrid-web/src/Datagrid.tsx
r0b1n: Now, that I think of it, we should also remove the data grid from the global namespace when the data grid is unloaded, so that if user switched pages, the export is not available.

packages/pluggableWidgets/datagrid-web/src/features/export.ts
```diff
@@ -0,0 +1,160 @@
...
+
+        return { type: "data", payload: items };
+    };
+
+    useEffect(() => {
+        if (!window[DATAGRID_DATA_EXPORT]) {
+            window[DATAGRID_DATA_EXPORT] = {};
+        }
+
+        window[DATAGRID_DATA_EXPORT][name] = { create };
```
r0b1n, line 121: We should remove `name` from the namespace if widget is unmounted.

packages/pluggableWidgets/datagrid-web/src/Datagrid.tsx
```diff
@@ -235,3 +247,35 @@ function transformColumnProps(props: ColumnsType[]): TableColumn[] {
...
+                break;
+            case "end":
+                console.info("_____End_Table_Body_______");
+                console.info("We are done with the data dump!");
+                break;
+        }
+    });
+    console.info("let's start dumping some data");
+    stream.start();
+}
```
iobuhov, line 281: Can we move this function to utils? I don't want this to be merged to main.

r0b1n: We don't need to merge it at all, this should be removed completely

r0b1n: Also some comments like `// Roman types ideas` should go.

iobuhov (approved):

r0b1n (approved):

PR #656, opened 2023-08-10
[WC-1887] Fix: container handles only own events

iobuhov:
### Pull request type

---

### Description
This PR fixes the issue with keyboard event propagation through the tree node container. Now the tree node container handles only its own events and lets child events pass through, so "enter" and " " (space) keyboard events can pass through and be handled correctly and, as a result, can be seen in input elements.

You need to create a tree node with a custom header and custom content. Put a text box and a text area as a child.

Expected: After running the project, you should be able to type `" "` (space) and use "enter" and arrow keys in the input and textarea.


HedwigAR:
I tested this in the test project of the original ticket. I am able to use `space`, `arrow keys` and `enter`, but when I double click to select text this quickly opens and closes the node. This will be fixed in an upcoming story.

HedwigAR (approved):

PR #671, opened 2023-08-21
fix(datagird-web): fix preview issue by passing correct index and useless warning (WC-1912)

iobuhov:
### Pull request type

---

### Description
As of now, in the header part, table passed incorrect index to function that should render header wrapper. Now index is passed correctly.

You show see no errors in design mode preview for Data grid.


leonardomendix (approved):

samuelreichert (approved):
LGTM

PR #673, opened 2023-08-23
[WC-2018] fix: column hide by default not working

gjulivan:
### Pull request type
Bug Fix: this issue relates with previously introduced changed to allow non-numeric ID (fix a11y).
This changes causing "visible column" not working as expected due to the reference with "hidden column" that is still index-based.

### Description
Fix by adjusting correct attribute.

### DIFF
--- a/packages/pluggableWidgets/datagrid-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/datagrid-web/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
--   We fixed an issue with column capabilities hide option that was unable to be hidden by default.
+-   We fixed an issue where columns, which were configured to be hidden by default, remained visible despite visibility settings.
 
 ## [2.8.1] - 2023-08-22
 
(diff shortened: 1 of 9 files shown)


packages/pluggableWidgets/datagrid-web/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+-   We fixed an issue with column capabilities hide option that was unable to be hidden by default.
```
iobuhov, line 11:
```suggestion
-   We fixed an issue with the column capabilities hide option that was unable to be hidden by default.
```

iobuhov (changes requested):
LGTM, only changelog record needs update.

iobuhov (dismissed):
LGTM

packages/pluggableWidgets/datagrid-web/cypress/integration/DataGridSelection.spec.js
```diff
@@ -1,20 +1,3 @@
-function terminalLog(violations) {
```
leonardomendix, line 1: *Extra: I have removed this unnecessary function leftover.

leonardomendix (approved):

iobuhov (approved):

PR #677, opened 2023-08-28
fix: [WC-2002] datefilter not fully shown on overflow scrolls

gjulivan:
### Pull request type
Bug fix:
date filter not fully visible on container overflow : scrolls.

### Description

fixed by changing popper js mode to `position:fixed` instead of `position:absolute` .
popper js will do the calculation for positioning (tethered to input) automatically.
but overflow will works.

https://popper.js.org/docs/v2/constructors/#strategy

### Developer testing:
 - check if popup goes under topbar ✅ [ with latest atlas-core ]
 - check if popup tethered to input when scrolled up✅
 - check if popup maintain correct position when datagrid header filter goes into sticky position ✅

### DIFF
--- a/packages/modules/data-widgets/package.json
+++ b/packages/modules/data-widgets/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@mendix/data-widgets",
   "moduleName": "Data Widgets",
-  "version": "2.8.7",
+  "version": "2.8.8",
   "license": "Apache-2.0",
   "copyright": "© Mendix Technology BV 2023. All rights reserved.",
   "private": true,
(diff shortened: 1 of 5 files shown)


iobuhov (changes requested):
Well, I see the fix for DateFilter, and it works, but what about DropdownFilter? A possible solution is probably same as for DateFilter. Use `position: fixed` on dropdown menu. But this will cause other problems (like repositioning menu on scroll) as well as issues with stacking. I attached video to JIRA story.

iobuhov (approved):
Approve as this story was split. Fix for Dropdown will be delivered in separate story. Check WC-2027

leonardomendix (approved):

PR #683, opened 2023-09-06
[WC-2026] Refactor pluggable commons 4

iobuhov:
### Pull request type

---

### Description
Same as before, removing PWC from deps.

### DIFF
--- a/packages/pluggableWidgets/color-picker-web/src/components/ColorPicker.tsx
+++ b/packages/pluggableWidgets/color-picker-web/src/components/ColorPicker.tsx
@@ -1,5 +1,5 @@
 import { createElement, ReactElement, useEffect, useState, useCallback } from "react";
-import { Alert } from "@mendix/pluggable-widgets-commons/components/web";
+import { Alert } from "@mendix/widget-plugin-component-kit/Alert";
 import { ModeEnum, TypeEnum, DefaultColorsType, FormatEnum } from "../../typings/ColorPickerProps";
 import {
     ColorState,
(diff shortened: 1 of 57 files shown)


packages/pluggableWidgets/color-picker-web/src/components/ColorPicker.tsx
```diff
@@ -1,5 +1,5 @@
 import { createElement, ReactElement, useEffect, useState, useCallback } from "react";
-import { Alert } from "@mendix/pluggable-widgets-commons/components/web";
+import { Alert } from "@mendix/widget-plugin-component-kit/Alert";
```
r0b1n, line 2: NIT: Would it be nicer to make this a default export?

r0b1n (dismissed):

packages/pluggableWidgets/color-picker-web/src/components/ColorPicker.tsx
iobuhov: default export for component?
There could be other components (like IconInternal)

iobuhov: In short, I would prefer to stay with named exports, as we have many tools in our stack that may treat default exports differently. I've already seen many issues related to default exports. If you need source, I can find, but for now let's just keep it as a named import.

gjulivan (approved):

packages/pluggableWidgets/color-picker-web/src/components/ColorPicker.tsx
r0b1n: I was thinking that main component (same as filename) could be default while secondary can be named. But keeping as is is also fine.

r0b1n (approved):

PR #699, opened 2023-09-20
fix: checkbox and filter selector control alignment

gjulivan:
### Pull request type

fixing alignment for DG2 filter selection control and checkbox toggle.

### Description
The code use css :has , which is currently not directly compatible with firefox (need to be enabled in config : `layout.css.has-selector.enabled`)

basic logic:
if table has filter selector that is not empty,
then set checkbox and filter-control to bottom-aligned, but with a little bit margin on the bottom to make it looks like it is bottom-centered


Catoledo84 (commented):
besides i see that is missing like the bordered version and we might need to think on design mode as well, i know that it is because is the testing batch, the new alignment is great 👍

iobuhov (dismissed):
LGTM

iobuhov (approved):

leonardomendix (approved):

PR #700, opened 2023-09-20
fix(tabbing): tab on filter dropdown losses focus

bsgriggs:
### Pull request type
Bug Fix

---

### Description
 Pressing tab on the last item or shift & tab on the first item of the filter selector list would cause the widget to lose focus (document.activeElement went back to the body).

| Tab | Shift & Tab |
| --- | --- |
|  Screenshot 2023-09-20 at 9 31 37 AM  | Screenshot 2023-09-20 at 9 31 30 AM   |

Changed on KeyDown for the FilterSelector component to check this scenario and set focus back to the filter selector button with a 100ms timeout. 

Testing: 
1. Recompile the Text Filter, Number Filter, and Date Filter widgets
2. Use all of the widgets in Data Grid 2 with "Adjustable by User" set to Yes
3. Run the project and click on one of the filter's button  
Screenshot 2023-09-20 at 9 38 53 AM  
4. Pressing tab on the last item, shift & tab on the first item, or the escape key on any item should close the dropdown and re-focus the button from step 3 allowing you to tab into the text box.


bsgriggs:
Ideally, tabbing on the final item in the list would focus on the input element after the button. 
Screenshot 2023-09-20 at 1 30 28 PM
I couldn't make this happen without changing the props of the FilterComponent and passing a reference to the input in all the component's usages.

gjulivan (commented):
instead of using Tab to traverse on the list of options, WCAG pattern actually recommend to use up and down arrow instead.
https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/

we will change this Tab behavior to follow the correct guideline instead.

gjulivan:
closing this PR for incorrect expected behavior.
instead of Tab, 
it should be changed to up and down arrow key instead.
https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/

PR #703, opened 2023-09-27
[WC-2027] DG2 dropdown filter position fixed and virtual scroll sticky header

gjulivan:
### Pull request type

feature: 

- dropdown filter position changed from absolute to fixed.
-- allows it to be always on top of the content.
- sticky header now applies to DG with virtual scrolling

### Description
- dropdown filter position fixed implemented with usePositionObserver
- sticky header components now uses the refactored - InfiniteBody 
- FilterComponent on dropdown filter is considered too big to put on single file : this is also refactored into multiple files.


samuelreichert (dismissed):

packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid-scroll.scss
```diff
@@ -0,0 +1,24 @@
...
+            .th {
+                position: -webkit-sticky; /* Safari */
+                position: sticky;
+                z-index: 50;
+            }
+        }
+    }
+}
+
+.sticky-table-container.infinite-loading {
```
iobuhov, line 13: Can we change class here? I think it's better if we call it `widget-datagrid-grid-container` or `widget-datagrid-table-container`

packages/shared/widget-plugin-grid/src/components/StickyHeaderTable.tsx
```diff
@@ -1,11 +1,16 @@
...
+import classNames from "classnames";
+import { PropsWithChildren, ReactElement, createElement, useCallback, useEffect, useRef } from "react";
+import { InfiniteBodyProps, useInfiniteControl } from "./InfiniteBody";
 
 // this wrapper component works by adding "sticky-sentinel" div (and hides it using margin-bottom -1px)
 // in order to detect table going out of viewport
 // this allows position: sticky to only applies when container is "stuck"
 // this way we allow any absolute position to still be on top when sticky not "stuck"
-export function StickyHeaderTable(props: PropsWithChildren): ReactElement {
+export function StickyHeaderTable(props: PropsWithChildren<InfiniteBodyProps>): ReactElement {
```
iobuhov, line 9: I'm not sure that moving this component to `widget-plugin` is a right choice. This component very grid specific and it may be hard to maintain it here as we may need more changes in this component. What you can do instead, is export `useInfiniteControl` and import it from `datagrid-web`

packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid-scroll.scss
gjulivan: the idea is, to make it as a container specifically for sticky behavior, not generic datagrid table,
sticky header will and can be optional.

packages/shared/widget-plugin-grid/src/components/StickyHeaderTable.tsx
gjulivan: fixed

leonardomendix (approved):

iobuhov (approved):

PR #738, opened 2023-10-03
[WC-2055] feat(combobox): hide no option text on filter none

gjulivan:
### Pull request type
Refactoring ,
change filter type no -> none
show no options text when filter type is none.
move no options text to be under filter type


iobuhov (approved):
LGTM

leonardomendix (approved):

PR #739, opened 2023-10-03
[WC-1777]: Fix Image widget size in Design Mode

rahmanunver:
### Pull request type

---

### Description
Added height only styling for the Design mode component.

Image widget height was set to 100% in Design mode, with this change it will take the given pixel height in the configuration, or the default which is 100px (already existent default value inside the component).


samuelreichert (approved):
Hey! @rahmanunver can you add a screenshot with the fixed image widget on Design Mode? Thanks!

PR #741, opened 2023-10-04
[WC-1297]feat(combo-box): Custom content dropzone

gjulivan:
### Pull request type

New feature:
Add custom content dropzone on combo-box

### Description
adding new property:
`optionsSourceAssociationCustomContentType` -> Yes, List Item, No
`optionsSourceAssociationCustomContent` -> widgets

general implementation:
Association-Caption  has to know when to render custom content and normal text, thus the load of figuring out which one to render falls into the caption-renderer.

upon implmentation:
- caption-renderer must know where to render: `type CaptionPlacement = "label" | "options";`
- this placement then matched with `optionsSourceAssociationCustomContentType` property to help caption-renderer figuring out correct component to render
- apparently, custom content must be rendered before input element, this adds logic complexity in SingleSelection, which is solved by using different classnames.
- as for multiSelection, custom content will always rendered using "boxes" styling, with "x" to clear option available on each item.

### DIFF
--- a/packages/pluggableWidgets/combobox-web/src/Combobox.xml
+++ b/packages/pluggableWidgets/combobox-web/src/Combobox.xml
@@ -98,6 +98,19 @@
                     <caption>Clearable</caption>
                     <description />
                 </property>
+                <property key="optionsSourceAssociationCustomContentType" type="enumeration" defaultValue="no" required="true">
+                    <caption>Custom content</caption>
+                    <description />
+                    <enumerationValues>
+                        <enumerationValue key="yes">Yes</enumerationValue>
+                        <enumerationValue key="listItem">List items only</enumerationValue>
+                        <enumerationValue key="no">No</enumerationValue>
+                    </enumerationValues>
+                </property>
+                <property key="optionsSourceAssociationCustomContent" type="widgets" required="true" dataSource="optionsSourceAssociationDataSource">
+                    <caption>Custom content</caption>
+                    <description />
+                </property>
                 <property key="selectedItemsStyle" type="enumeration" defaultValue="text" required="true">
                     <caption>Show selected items as</caption>
                     <description />
--- a/packages/pluggableWidgets/combobox-web/src/components/Placeholder.tsx
+++ b/packages/pluggableWidgets/combobox-web/src/components/Placeholder.tsx
@@ -1,5 +1,5 @@
 import classNames from "classnames";
-import { PropsWithChildren, ReactElement, createElement } from "react";
+import { PropsWithChildren, ReactElement, createElement, Fragment } from "react";
 import { DownArrow } from "../assets/icons";
 
 export function Placeholder(): ReactElement {
@@ -22,8 +22,13 @@ export function NoOptionsPlaceholder(props: PropsWithChildren): ReactElement {
 
 interface InputPlaceholderProps extends PropsWithChildren {
     isEmpty: boolean;
+    useWrapper?: boolean;
 }
 export function InputPlaceholder(props: InputPlaceholderProps): ReactElement {
+    if (props.useWrapper === false) {
+        return <Fragment>{props.children}</Fragment>;
+    }
+
     return (
         <span
             className={classNames("widget-combobox-text-label", {
--- a/packages/pluggableWidgets/combobox-web/src/components/Placeholder.tsx
+++ b/packages/pluggableWidgets/combobox-web/src/components/Placeholder.tsx
@@ -27,8 +27,8 @@ interface InputPlaceholderProps extends PropsWithChildren {
 export function InputPlaceholder(props: InputPlaceholderProps): ReactElement {
     return (
         <div
-            className={classNames(props.className || "widget-combobox-label-text", {
-                "widget-combobox-label-text-placeholder": props.isEmpty
+            className={classNames(props.className || "widget-combobox-placeholder-text", {
+                "widget-combobox-placeholder-empty": props.isEmpty
             })}
         >
             {props.children}
--- a/packages/pluggableWidgets/combobox-web/src/components/Placeholder.tsx
+++ b/packages/pluggableWidgets/combobox-web/src/components/Placeholder.tsx
@@ -1,5 +1,5 @@
 import classNames from "classnames";
-import { PropsWithChildren, ReactElement, createElement, Fragment } from "react";
+import { PropsWithChildren, ReactElement, createElement } from "react";
 import { DownArrow } from "../assets/icons";
 
 export function Placeholder(): ReactElement {
@@ -22,17 +22,13 @@ export function NoOptionsPlaceholder(props: PropsWithChildren): ReactElement {
 
 interface InputPlaceholderProps extends PropsWithChildren {
     isEmpty: boolean;
-    useWrapper?: boolean;
+    className?: string;
 }
 export function InputPlaceholder(props: InputPlaceholderProps): ReactElement {
-    if (props.useWrapper === false) {
-        return <Fragment>{props.children}</Fragment>;
-    }
-
     return (
         <div
-            className={classNames("widget-combobox-text-label", {
-                "widget-combobox-text-label-placeholder": props.isEmpty
+            className={classNames(props.className || "widget-combobox-label-text", {
+                "widget-combobox-label-text-placeholder": props.isEmpty
             })}
         >
             {props.children}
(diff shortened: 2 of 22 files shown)


packages/pluggableWidgets/combobox-web/src/Combobox.xml
```diff
@@ -98,6 +98,19 @@
...
+                    <enumerationValues>
+                        <enumerationValue key="yes">Yes</enumerationValue>
+                        <enumerationValue key="listItem">List items only</enumerationValue>
+                        <enumerationValue key="no">No</enumerationValue>
+                    </enumerationValues>
+                </property>
+                <property key="contentPlaceholder" type="widgets" required="false">
+                    <caption>Options Placeholder</caption>
+                    <description />
+                </property>
```
r0b1n, line 113: 1) I believe this prop should be connected to DS: `dataSource="optionsSourceAssociationDataSource"`. 
2) We should make it required but hide it in `getProperties` of the editorConfig if set to "No". This way we get proper consistency checks when needed and not bothered.
3) Name it in a similar way we name other attributes connected to options data source: `optionsSourceAssociationCaptionAttribute` and `optionsSourceAssociationCaptionExpression`, so prefix it accordingly
4) Drop "placeholder" from the name as it is not a placeholder, it is the content itself.

iobuhov:
Checks didn't pass.

packages/pluggableWidgets/combobox-web/src/helpers/Association/AssociationSimpleCaptionsProvider.tsx
```diff
@@ -43,7 +51,27 @@ export class AssociationSimpleCaptionsProvider implements CaptionsProvider {
...
+        }
+
+        return this.customContent?.get(item);
+    }
+
+    render(value: string | null, placement: CaptionPlacement): ReactNode {
+        const { customContentType } = this;
+
+        return (
+            <span className="widget-combobox-caption">
```
r0b1n, line 70: We probably should not place custom content into a `span`. Maybe wrap only text version in span while having a `div` with a different CSS class so that option types can be targeted independently while styling.

gjulivan: it works well in the "custom options",
but will not work on custom selected item label.
because we need to display the "x" button there to de-select an item.

r0b1n: Not quite getting what you mean. You mean that if we do span AND div we have to implement X button two times? If so, we can go for div for both options with different class set, also what crossed my mind just now is that we can also use different classes depending on where the item is in the list, in the combobox input or in the list of options.

samuelreichert (approved):

samuelreichert:
Still needs rebase, and fix previous comments

packages/pluggableWidgets/combobox-web/src/helpers/Association/AssociationSimpleCaptionsProvider.tsx
gjulivan: sorry, bad reading, i thought you meant to removed the span wrapper entirely (which is not possible)

this is fix now.

packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelection.tsx
```diff
@@ -39,7 +39,14 @@ export function SingleSelection({
                 readOnly={selector.readOnly}
                 getToggleButtonProps={getToggleButtonProps}
             >
-                <div className="widget-combobox-selected-items">
+                <div
+                    className={classNames(
+                        { "widget-combobox-selected-items": selector.customContentType !== "yes" },
+                        {
+                            "widget-combobox-custom-content": selector.customContentType === "yes"
+                        }
```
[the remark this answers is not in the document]
gjulivan: fix.

packages/pluggableWidgets/combobox-web/src/components/Placeholder.tsx
```diff
@@ -22,8 +22,13 @@ export function NoOptionsPlaceholder(props: PropsWithChildren): ReactElement {
 
 interface InputPlaceholderProps extends PropsWithChildren {
     isEmpty: boolean;
+    useWrapper?: boolean;
 }
 export function InputPlaceholder(props: InputPlaceholderProps): ReactElement {
+    if (props.useWrapper === false) {
+        return <Fragment>{props.children}</Fragment>;
+    }
+
```
r0b1n, line 31: This should probably return `ReactNode`, so that we don't have to wrap into a fragment for the sake of typescript.

r0b1n: Can we convert this wrapper to a div and always render it, for both, text and custom contents. At the moment we have two spans for text and one div of custom elements. Which seems quite different. I am thinking of 
`div.placeholder -> span.text -> (text)`
`div.placeholder -> div.custom-context -> (widgets)`

So that top level is always the same and then it differs depends on the type.

packages/pluggableWidgets/combobox-web/src/helpers/Association/AssociationSimpleCaptionsProvider.tsx
```diff
@@ -43,7 +51,25 @@ export class AssociationSimpleCaptionsProvider implements CaptionsProvider {
...
+        return this.customContent?.get(item);
+    }
+
+    render(value: string | null, placement: CaptionPlacement): ReactNode {
+        const { customContentType } = this;
+
+        return customContentType === "no" || (placement === "label" && customContentType === "listItem") ? (
+            <span className="widget-combobox-caption">{this.get(value)}</span>
+        ) : (
+            <div className="widget-combobox-custom-caption">{this.getCustomContent(value)}</div>
```
r0b1n, line 72: We can also probably do either:
```
.[...]-caption-text
.[...]-caption-custom
```
or (if helps with cleaner styling)
```
.[...]-caption.[...]-caption-text
.[...]-caption.[...]-caption-custom
```

packages/pluggableWidgets/combobox-web/src/helpers/Association/AssociationSimpleCaptionsProvider.tsx
```diff
@@ -1,15 +1,20 @@
+import { DynamicValue, ListAttributeValue, ListExpressionValue, ListWidgetValue, ObjectItem } from "mendix";
 import { ReactNode, createElement } from "react";
-import { DynamicValue, ListAttributeValue, ListExpressionValue, ObjectItem } from "mendix";
-import { CaptionsProvider } from "../types";
+import { OptionsSourceAssociationCustomContentTypeEnum } from "../../../typings/ComboboxProps";
+import { CaptionPlacement, CaptionsProvider } from "../types";
 
-interface Props {
+export interface Props {
```
r0b1n, line 6: Do we use that anywhere else so that we export it? If so, we have to remove it so something more meaningful, not just Props. If not, we should not export it.

packages/pluggableWidgets/combobox-web/src/components/Placeholder.tsx
gjulivan: first, ReactNode is not JSX, so usage such as <InputPlaceholder /> will become invalid, 
but the placeholder changed into div. (fix)

packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelection.tsx
```diff
@@ -39,7 +39,12 @@ export function SingleSelection({
                 readOnly={selector.readOnly}
                 getToggleButtonProps={getToggleButtonProps}
             >
-                <div className="widget-combobox-selected-items">
+                <div
+                    className={classNames({
+                        "widget-combobox-selected-items": selector.customContentType !== "yes",
+                        "widget-combobox-custom-content": selector.customContentType === "yes"
```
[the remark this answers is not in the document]
gjulivan: "widget-combobox-selected-items" is kept. 
"widget-combobox-custom-content" is now just appended.
(fix)

packages/pluggableWidgets/combobox-web/src/helpers/Association/AssociationSimpleCaptionsProvider.tsx
gjulivan: updated to "widget-combobox-caption-custom".

packages/pluggableWidgets/combobox-web/src/helpers/Association/AssociationSimpleCaptionsProvider.tsx
gjulivan: it was exported at somepoint in development phase. but not needed anymore.
removed.

r0b1n (approved):

r0b1n (dismissed):

r0b1n (approved):

leonardomendix (approved):

PR #743, opened 2023-10-05
[WC-2051]: Make Image functionality configurable in Rich Text

rahmanunver:
### Pull request type

---

### Description
Not having _Image button_ (Insert group) in the Rich Text toolbar prevents the user from adding/using Image functionalities in the basic toolbar, the only option to deal with this is to **add the Image button from the advanced toolbar configuration**, but there was no "_Insert group_" available which covers "Image button" too. 

So, "Custom" toolbar now has the Insert group option 
Screenshot 2023-10-05 at 18 31 52

Screenshot 2023-10-05 at 18 30 35

And the "Basic" toolbar now has the Image button along with other default buttons.

Copy/paste, drag/drop the images into rich text, make all types of rich text editors can share their image content. (Unlike the customers situattion)


iobuhov:
Checks not passed.

iobuhov (approved):

PR #744, opened 2023-10-05
feat(datagrid-web): add column visible

Andries-Smit:
When Using the data grid 2, depending on the the context come columns do not need to be shown. 
By adding  'Visible' expression to the column, the developer can specify when the column need to be shown.
This PR includes an highlight for Visibility in the structured mode.

image

### DIFF
--- a/packages/pluggableWidgets/datagrid-web/src/Datagrid.editorConfig.ts
+++ b/packages/pluggableWidgets/datagrid-web/src/Datagrid.editorConfig.ts
@@ -179,6 +179,7 @@ export const getPreview = (
                   dynamicText: "Dynamic text",
                   draggable: false,
                   hidable: "no",
+                  visible: "true",
                   size: 1,
                   sortable: false,
                   alignment: "left",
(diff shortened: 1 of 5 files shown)


Andries-Smit:
The visible state should be like the core visibility visible state on the root of the widget. Ideally it would have the same kind of options: dialog: Default, always, based on attribute value, based on expression, for selective roles. (But this cannot (yet) be used for the columns objects, but only on the root level, the expression is the next best thing.)
 
image

In this case I expect that visibility “false” will overwrite all other settings, and never show the column, and will not show the column in the “selector”.  While “true” will be equal to "Default", and behave as the current setup depending on the key “hidable”, together with the visibility state of the column selector.

takumagott:
Hey @Andries-Smit, 
Thanks again for the PR! This is a feature that has been asked by more people, so I appreciate the initiative to implement it for everyone to use!
Using an expression makes sense here, so let's get this PR in.
Let's quickly update the [docs](https://docs.mendix.com/appstore/modules/data-grid-2/#4-columns) as well though.

CC @iobuhov

gjulivan:
linked to https://mendix.atlassian.net/browse/WC-2197

> Hey @Andries-Smit, Thanks again for the PR! This is a feature that has been asked by more people, so I appreciate the initiative to implement it for everyone to use! Using an expression makes sense here, so let's get this PR in. Let's quickly update the [docs](https://docs.mendix.com/appstore/modules/data-grid-2/#4-columns) as well though.
> 
> CC @iobuhov

gjulivan:
closing this PR, will be resolved in https://github.com/mendix/web-widgets/pull/799

PR #745, opened 2023-10-06
feat(datagrid-web): reflect changes from Table on Export API

samuelreichert:
### Pull request type

---


packages/pluggableWidgets/datagrid-web/src/Datagrid.tsx
```diff
@@ -15,10 +15,16 @@ import { GridHeaderWidgets } from "./components/GridHeaderWidgets";
 import { Column } from "./helpers/Column";
 import { useDG2ExportApi, UpdateDataSourceFn } from "./features/export";
 
+export type OnColumnsChangeProps = {
+    columnOrder?: string[];
+    hiddenColumns?: string[];
+};
+
```
iobuhov, line 22: I like this idea of extracting this two arrays to `ColumnProps` 👍 , but I think it should be `ColumnsState`. I think time has come to "extract" this to hook where we can utilize `useReducer` to manage column params all together.
Can we move this to some hook like `useColumnsState`?

NOTE: I'm not expert, but I think for names we should use plural `columns`. So, probably we should use `columnsOrder`

iobuhov (changes requested):

packages/pluggableWidgets/datagrid-web/src/features/export.ts
```diff
@@ -260,14 +278,25 @@ type State = WorkingState | ReadyState | InitState;
 
 type DataSourceUpdate = {
     type: "DataSourceUpdate";
-    payload: { offset: number; items: ObjectItem[]; hasMoreItems: boolean; limit: number };
+    payload: {
+        offset: number;
+        items: ObjectItem[];
+        hasMoreItems: boolean;
+        limit: number;
+        columnsProps: ColumnsProps;
```
iobuhov, line 286: From my POV, we don't need to pass `columnsProps` to the export. All this could be achieved by filtering and sorting `columns` prop. This way our code less coupled and export is more simple.

iobuhov (changes requested):
I have some problems. Changes needed heere.

iobuhov (dismissed):

leonardomendix (approved):

PR #749, opened 2023-10-11
[WC-2150]: Configure a11y for combobox arrow (Toggle Button)

rahmanunver:
### Pull request type

---

### Description
Since we are using the getToggleButtonProps method from Downshift, we need to manipulate the default behaviour of adding the aria-expanded, this one line change does that.


gjulivan (dismissed):

gjulivan:
please fix failed test

iobuhov (changes requested):
I'm not sure that setting `aria-expanded` to `undefined` is a right choice. After reading [article on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#associated_roles) I see that this attribute is applied correctly (they even have example with combobox). The reason why we had Lighthouse report about this attribute is probably the fact that his attribute was applied to element with wrong role...
So, in my opinion, to fix this issue we should add `role=button` to arrow element. I'm not sure, need testing, but I think this is main reason why lighthouse is throwing this error.

rahmanunver:
@iobuhov is right about his suggestion, indeed unsetting `aria-expanded` is not a good option, but settin `role='button'` isn't sufficient alone, we need an `aria-label` too to silence the Lighthouse warnings. Updating the PR accordingly. 
CC: @gjulivan  @takumagott

iobuhov (approved):
LGTM 👍

PR #755, opened 2023-10-12
[WC-2144][WC-2150]: Combobox a11y adjustments, keyboard navigation for Multi

rahmanunver:
### Pull request type

---

### Description
Added focus state styles, aria-label, and behavior change in Clear button, when it is cleared, the focus will go to input field as requested.
Screenshot 2023-10-12 at 18 36 18

The changes in description should behave properly without causing a11y glitches on keyboard, aria-label should be visible with "Clear selection" value, and clear button now should have a focus state style. 
-->

### DIFF
--- a/packages/pluggableWidgets/combobox-web/src/__tests__/SingleSelection.spec.tsx
+++ b/packages/pluggableWidgets/combobox-web/src/__tests__/SingleSelection.spec.tsx
@@ -46,7 +46,9 @@ describe("Combo box (Association)", () => {
             ariaRequired: true,
             clearable: true,
             filterType: "contains",
-            selectedItemsStyle: "text"
+            selectedItemsStyle: "text",
+            clearButtonAriaLabel: "Clear selection",
+            removeValueAriaLabel: "Remove value"
         };
         if (defaultProps.optionsSourceAssociationCaptionType === "expression") {
             defaultProps.optionsSourceAssociationCaptionExpression!.get = i => {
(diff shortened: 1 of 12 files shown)


gjulivan (approved):

iobuhov (dismissed):

packages/pluggableWidgets/combobox-web/src/__tests__/SingleSelection.spec.tsx
```diff
@@ -44,7 +44,9 @@ describe("Combo box (Association)", () => {
             ariaRequired: true,
             clearable: true,
```
iobuhov, line 45: NIT: Can we add return type to `getTogglebutton` fix this lint warning?

packages/pluggableWidgets/combobox-web/src/ui/Combobox.scss
```diff
@@ -112,6 +112,12 @@ $disabled-text-color: #6c7180;
             stroke-width: 0;
             transition: stroke-width 0.2s ease-in-out;
         }
+
+        &:focus .widget-combobox-icon-container {
+            border-radius: 2px !important;
+            outline: 2px solid $brand-primary !important;
```
iobuhov, line 118: Why we have to use `!important`? Can we avoid that?

iobuhov (approved):

iobuhov (approved):

PR #771, opened 2023-10-18
fix: inconsistent behavior with editor set data

gjulivan:
### Pull request type
Bug fix: Editor not displaying data sometimes.

### Description
it seems like there’s a racing condition on updating the editor’s data.
this probably because unsync DOM with React’s component.
we could try debugging more and making a proper fix, but that would require a lot of refactoring.
delaying setData for 50ms seems works for now.

### DIFF
--- a/packages/pluggableWidgets/rich-text-web/src/components/Editor.tsx
+++ b/packages/pluggableWidgets/rich-text-web/src/components/Editor.tsx
@@ -35,6 +35,7 @@ export class Editor extends Component<EditorProps> {
     element: HTMLElement;
     lastSentValue: string | undefined;
     applyChangesDebounce: () => void;
+    setDataDebounce: (data: string | undefined) => void;
     cancelRAF: (() => void) | undefined;
     hasFocus: boolean;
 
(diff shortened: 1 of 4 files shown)


packages/pluggableWidgets/rich-text-web/src/components/Editor.tsx
```diff
@@ -228,7 +228,7 @@ export class Editor extends Component<EditorProps> {
         // call addListeners immediately.
         // https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#method-setData
         if ("data" in args) {
-            this.editor?.setData(args.data, () => this.addListeners());
+            setTimeout(() => this.editor?.setData(args.data, () => this.addListeners()), 50);
```
iobuhov, line 231: What will happen if I call `updateEditorState` multiple times? I think, we need to clear timers to avoid multiple `setData` calls.

gjulivan: what about debounce? there's already a debounce function being used in this file

iobuhov: Yeah, we should use debounce

gjulivan: changed.

iobuhov (approved):

leonardomendix (approved):

PR #774, opened 2023-10-20
[WC-2151]: Combobox screenreader a11y

rahmanunver:
### Pull request type

---

### Description
Override the default screen reader `getA11yStatusMessage` method inside Downshift to include the selected items in announcement when reading the combobox. 

Now users can modify the a11yStatusMessage which is read by screen readers while using combobox, by changing the default message in the Accessibility tab in SP.

Also, since to this story initially was focusing on Multi selection, we noticed there is a mismatch in behaviour with Single and Multi selection; 

Single selection doesn't open menu when tab focused, but does when click focused. (_correct behaviour_)
Multi selection did open menu both when tab and click focused, so aligned it with Single now.

- Focusing on Multi selection both with tab and click, using screenreader utilities and making sure we have the selected values announced.  
- Multi selection changed behaviour in terms of focusing, it doesn't open the dropdown anymore on Tab focus. 
- Added config to Accessibility tab to modify the screen reader announcements.

### DIFF
--- a/packages/pluggableWidgets/combobox-web/src/Combobox.xml
+++ b/packages/pluggableWidgets/combobox-web/src/Combobox.xml
@@ -182,15 +182,15 @@
                 </property>
             </propertyGroup>
             <propertyGroup caption="Accesibility Status Message ">
-                <property key="i18nSelectedValue" type="string" defaultValue="Selected value: " required="true">
+                <property key="a11ySelectedValue" type="string" defaultValue="Selected value: " required="true">
                     <caption>Selected value</caption>
                     <description>Output example: "Selected value: Avocado, Apple, Banana."</description>
                 </property>
-                <property key="i18nOptionsAvailable" type="string" defaultValue="Number of options available:" required="true">
+                <property key="a11yOptionsAvailable" type="string" defaultValue="Number of options available:" required="true">
                     <caption>Options available</caption>
                     <description>Output example: "Number of options available: 1"</description>
                 </property>
-                <property key="i18nInstructions" type="string" defaultValue="Use up and down arrow keys to navigate. Press Enter or Space Bar keys to select." required="true">
+                <property key="a11yInstructions" type="string" defaultValue="Use up and down arrow keys to navigate. Press Enter or Space Bar keys to select." required="true">
                     <caption>Instructions</caption>
                     <description>Instructions to be read after announcing the status.</description>
                 </property>
--- a/packages/pluggableWidgets/combobox-web/src/Combobox.xml
+++ b/packages/pluggableWidgets/combobox-web/src/Combobox.xml
@@ -188,10 +188,7 @@
                 </property>
                 <property key="i18nSelectedItemPlural" type="string" defaultValue="Selected items are" required="true">
                     <caption>Selected items (Plural)</caption>
-                    <description>
-                        Used when Multi-selection is enabled and there are multiple
-                        selected items. Output example: Selected items are Avocado, Apple, Banana.
-                    </description>
+                    <description>Used when Multi-selection is enabled and there are multiple selected items. Output example: Selected items are Avocado, Apple, Banana.</description>
                 </property>
                 <property key="i18nNoResults" type="string" defaultValue="No results are available." required="true">
                     <caption>No results</caption>
--- a/packages/pluggableWidgets/combobox-web/src/Combobox.xml
+++ b/packages/pluggableWidgets/combobox-web/src/Combobox.xml
@@ -178,7 +178,36 @@
                 </property>
                 <property key="removeValueAriaLabel" type="string" defaultValue="Remove value" required="true">
                     <caption>Remove value</caption>
-                    <description>Used for the individual selected value labels in multi-selection</description>
+                    <description>Used for the individual selected value labels in multi-selection.</description>
+                </property>
+            </propertyGroup>
+            <propertyGroup caption="Accesibility Status Message ">
+                <property key="i18nSelectedItemSingular" type="string" defaultValue="Selected item is" required="true">
+                    <caption>Selected item (Singular)</caption>
+                    <description>Output example: "Selected item is Avocado."</description>
+                </property>
+                <property key="i18nSelectedItemPlural" type="string" defaultValue="Selected items are" required="true">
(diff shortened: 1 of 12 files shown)


packages/pluggableWidgets/combobox-web/src/hooks/useDownshiftMultiSelectProps.ts
```diff
@@ -127,16 +127,43 @@ function useComboboxProps(
...
+                }
+
+                if (options.resultCount !== options.previousResultCount) {
+                    message += `${options.resultCount} result${
+                        options.resultCount === 1 ? " is" : "s are"
+                    } available, use up and down arrow keys to navigate. Press Enter or Space Bar keys to select.`;
+                }
+
+                return message;
+            },
```
iobuhov, line 152: I'm bit concerned, how we going to translate this messsage?

rahmanunver: Not sure either, I think we didn't handle that at all, will check if downshift is doing anything there

packages/pluggableWidgets/combobox-web/src/hooks/useDownshiftMultiSelectProps.ts
```diff
@@ -127,16 +127,43 @@ function useComboboxProps(
...
+                        ? `Selected item${selectedItems.length === 1 ? " is" : "s are"} ${selectedItems
+                              .map(itemId => selector.caption.get(itemId))
+                              .join(",")}.`
+                        : "";
+                if (!options.isOpen) {
+                    message += "";
+                }
+
+                if (!options.resultCount) {
+                    message += "No results are available.";
```
gjulivan, line 142: isn't this should use empty item message?

rahmanunver: Do you mean emptyOptionsText? This is announced when there's no search result onChange of the combobox input, wasn't empty item message

gjulivan (approved):

gjulivan (dismissed):

packages/pluggableWidgets/combobox-web/src/Combobox.xml
```diff
@@ -165,7 +165,21 @@
...
+            <propertyGroup caption="Accesibility Status Message ">
+                <property key="i18nSelectedValue" type="string" defaultValue="Selected value: " required="true">
+                    <caption>Selected value</caption>
+                    <description>Output example: "Selected value: Avocado, Apple, Banana."</description>
+                </property>
+                <property key="i18nOptionsAvailable" type="string" defaultValue="Number of options available:" required="true">
+                    <caption>Options available</caption>
+                    <description>Output example: "Number of options available: 1"</description>
+                </property>
+                <property key="i18nInstructions" type="string" defaultValue="Use up and down arrow keys to navigate. Press Enter or Space Bar keys to select." required="true">
```
iobuhov, line 180: Can we change type to `textTempalte`?
We can have default values using `translations` tag (for en_US)

iobuhov, line 180: NIT: i18n - stands from "internationalization", I more correct is to use a11y. As this props about a11y and not i18n

packages/pluggableWidgets/combobox-web/src/hooks/useDownshiftMultiSelectProps.ts
```diff
@@ -127,16 +139,37 @@ function useComboboxProps(
...
+                        ? `${a11yStatusMessage.i18nSelectedValue} ${selectedItems
+                              .map(itemId => selector.caption.get(itemId))
+                              .join(",")}. `
+                        : "";
+
+                if (options.resultCount > 0) {
+                    message += `${a11yStatusMessage.i18nOptionsAvailable} ${options.resultCount}. ${a11yStatusMessage.i18nInstructions}`;
+                } else {
+                    message += a11yStatusMessage.i18nNoOption;
+                }
```
iobuhov, line 154: Yeah, looks much better!

packages/pluggableWidgets/combobox-web/src/hooks/useDownshiftMultiSelectProps.ts
```diff
@@ -127,16 +139,37 @@ function useComboboxProps(
...
+
+                console.log(message);
```
iobuhov, line 156: Remove log.

iobuhov (changes requested):

packages/pluggableWidgets/combobox-web/src/Combobox.xml
rahmanunver: We can yeah, but what would be the benefit?

iobuhov: Main benefits are:
1. Developer can use template expression
2. Text template require template for each language configured in SP. So, this will make sure this labels get translated translation for other languages.

packages/pluggableWidgets/combobox-web/src/Combobox.xml
```diff
@@ -165,7 +165,30 @@
                 </property>
                 <property key="removeValueAriaLabel" type="string" defaultValue="Remove value" required="true">
                     <caption>Remove value</caption>
-                    <description>Used for the individual selected value labels in multi-selection</description>
+                    <description>Used for the individual selected value labels in multi-selection.</description>
+                </property>
+            </propertyGroup>
+            <propertyGroup caption="Accesibility Status Message ">
```
samuelreichert, line 171:
```suggestion
            <propertyGroup caption="Accessibility Status Message ">
```

samuelreichert (commented):
Hey, is it expected to have these missing dependencies on `packages/pluggableWidgets/combobox-web/src/hooks/useDownshiftSingleSelectProps.ts`? (a11yStatusMessage.a11yInstructions, ...)

samuelreichert (dismissed):

samuelreichert (approved):

iobuhov (dismissed):
LGTM 👍

gjulivan (approved):

PR #776, opened 2023-10-20
feat(datagrid-web): double click action

Andries-Smit:
Data Grid 1 has a double-click option. This PR adds the double-click option to the DataGrid 2. This will allow developers to replace old data grids with matching functionality.

The double click will be even more effective when the DG2 can emit events for the data view Data source Type "Listen to widget."

An option to consider is to add an extra action for "onClickDouble.". Though this might conflict with the single action, both actions might delay the click action.


iobuhov (approved):
LGTM. This could go to Data Widgets 2.12 I think.

iobuhov:
Need rebasing btw

iobuhov:
Work is continued in #873

PR #783, opened 2023-10-24
[WC-1704]: prepare v2.9.0 of web actions 

r0b1n:


packages/modules/web-actions/package.json
```diff
@@ -15,7 +15,7 @@
   },
   "moduleFolderNameInModeler": "webactions",
   "marketplace": {
-    "minimumMXVersion": "9.18.4.62522",
+    "minimumMXVersion": "9.24.10.18073",
```
iobuhov, line 18: Just one question, could it be 9.24.0?

r0b1n: Should be possible, I will downgrade the project.

iobuhov (dismissed):

leonardomendix (approved):

iobuhov (approved):

PR #792, opened 2023-10-30
[WC-2174] Keyboard navigation for data grid

iobuhov:
### Pull request type

---

### Description
Keyboard navigation implemented according to WAI ARIA guide.

The keyboard navigation bidings:
Arrows - move focus
PageUp/PageDown - move focus one page (according to page size in settings)
Home/End - move focus to row start/end (first/last column)
CTRL+Home - go to first cell (top left corner)
CTRL+End - go to the last cell (bottom right corner)

- Keyboard navigation
- Virtual scroll with key nav
- Selection with key nav
- Focus states (cell should have outline)
- Keep focus state between page reloads, sort
- Focus reset when column is become hidden (focus mvoe to initial position, first cell)
- Checkbox for selection


packages/pluggableWidgets/datagrid-web/src/components/Widget.tsx
```diff
@@ -250,39 +260,35 @@ export function Widget<C extends GridColumn>(props: WidgetProps<C>): ReactElemen
...
-                                        >
-                                            <div className="empty-placeholder">{children}</div>
-                                        </div>
+                                        <Row
+                                            CellComponent={CellComponent}
+                                            className={props.rowClass?.(item)}
+                                            columns={columnsToShow}
+                                            index={rowIndex}
+                                            item={item}
+                                            key={`row_${item.id}`}
```
gjulivan, line 272: are we sure that the item.id is already unique between rows? shouldn't we add index here as well?

iobuhov: It's okay to use `item.id` as id is GUID and rows never repeat (if you now such case, let me know as it's probably a bug)

packages/pluggableWidgets/datagrid-web/src/features/keyboard-navigation/context.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export function useKeyNavProvider(props: LayoutProps): Provider {
+    const focusController = useFocusTargetController(props);
+
+    return useMemo<Provider>(
+        () =>
+            function Provider({ children }) {
+                return <context.Provider value={{ focusController }}>{children}</context.Provider>;
+            },
+        [focusController]
```
gjulivan, line 23: this code doesn't make sense..
the focusController never change, it is a constant.
so the useMemo is not needed here.

Thus, can't we simply return the component directly?
something like:
```
export function KeyNavProvider(props: LayoutProps extends PropsWithChildren): ReactElement {
    const {children, ...focusProps} = props;
    const focusController = useFocusTargetController(focusProps);

    return <context.Provider value={{ focusController }}>{props.children}</context.Provider>;
}
```

rahmanunver (approved):
LGTM

packages/pluggableWidgets/datagrid-web/src/features/keyboard-navigation/context.tsx
iobuhov: `useMemo` is used to avoid unnecessary re-renders for context consumers. By using memo we `value` prop stable between DG re-renders. So, `useMemo` will stay.

iobuhov:
Closing as this is code will be merged together with #817

PR #795, opened 2023-10-31
feat(datagrid-web): no pagination

Andries-Smit:
Data grid 1 has an option "Show paging bar": "No" The Data Grid 2, will always show pagination buttons. This PR will improve the property "Pagination" by adding the option "None"

As a developer/user we do not always need pagination, some use cases will only have a limited number of items, and the paging buttons only confuses the end users, and taking valuable space on the screen.

This is a non breaking change, by adding an option to the property and requires not code change.
The page Size property is moved below the pagination, and the page Size is hidden when the  "Pagination" option "None".

image


gjulivan:
moved internally https://github.com/mendix/web-widgets/pull/865

PR #800, opened 2023-11-03
[WC-2227] DataGrid filtering cleanup

iobuhov:
### Pull request type

---

### Description
Extract filtering code to the separate "feature" module.


gjulivan:
generally, I'm not in favor of the idea of having custom hooks that returns jsx components, 
probably we can think of something more towards clean MVC pattern.
will take a look deeper into this.

PR #809, opened 2023-11-10
[WC-2148] Combobox menu outside view

rahmanunver:
### Pull request type

---

### Description
Added hook for detecting when menu is out of body, and flipping it to top of the combobox.


packages/pluggableWidgets/combobox-web/src/hooks/useMenuPlacement.ts
```diff
@@ -0,0 +1,48 @@
...
+    const ref = useRef<T>(null);
+    const [style, setStyle] = useState<React.CSSProperties>({ visibility: "hidden", position: "fixed" });
+    const [setStyleDebounced, abort] = useMemo(() => debounce(setStyle, 32), [setStyle]);
+    const menuHeight = ref.current?.offsetHeight ?? 0;
+    const targetBox = usePositionObserver(ref.current?.parentElement ?? null, isOpen);
+
+    useEffect(() => {
+        if (!isOpen) {
+            setStyle(prev => {
+                if (prev.visibility === "hidden") {
```
rahmanunver, line 15: Is this part needed after the debounce approach? I can't see which case it handles now. @iobuhov

gjulivan (approved):

leonardomendix (approved):

leonardomendix (approved):

gjulivan (approved):

packages/pluggableWidgets/combobox-web/src/hooks/useMenuPlacement.ts
iobuhov: This part indeed has very little change with debounce approach, but in fact, I would prefer to keep it as it's prevent undesired state once we close menu.

iobuhov (approved):

PR #810, opened 2023-11-10
feat(rich-text-web): configurable styles drop down based on style …

tobias-at-work:
…classes from app theme

### Pull request type

---

### Description

The rich text widget has a styles menu that allows to style a selection or starting from the current cursor position. However, only very few choices are available, and the hardcoded styles might not fit the apps needs

image

Besides the fonts in the menu look different than the fonts in the editor:

image

This feature leverages CKEditor4's capability to use custom styles, in case of a Mendix application the theme defined in App/Styling: https://ckeditor.com/docs/ckeditor4/latest/guide/dev_howtos_styles.html

By default, nothing is changed, so existing apps should behave as before (updating widgets is required though). When the developer decides to use custom styles, he sets the "Styleset name" in the widget properties. 

The developer can then assign a subset of style classes from the App's theme to the rich text widget. 
image

The styles menu can be explicitly designed so that proper styles are selectable depending on the context (text, table, selection, etc.). The "Valid Selectors" property defines available style classes from the theme. In the "Edit Styles" dialog, the available styles can be assigned to HTML tags, so that the "Styles" menu shows available options based on the current context.

image

Styles in the menu look the same as in the editor, so there is a better WYSIWYG experience.

Menu:

image

Editor:

image

Result in HTML Element:

image

Inspecting the resulting content stored in the entity shows that style information is stored in classes instead of style attributes and HTML, which is more in line with Mendix best practices:

image

I have created a Mendix app including data snapshot to showcase the improved widget. If so desired, I can invite any reviewer as a team member to test the changes.

### DIFF
--- a/packages/pluggableWidgets/rich-text-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/rich-text-web/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Added
+
+-   We added the posibility to customize the styles menu and use a selection of styles from the app's theme.
+
 ## [2.2.3] - 2023-10-27
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


tobias-at-work:
Will resubmit a pull request for version 3.x

PR #813, opened 2023-11-14
feat(richt-text-web): Improved content templates

tobias-at-work:
### Pull request type

---

### Description

This PR replaces an older open PR, as I had to reorganize my forked repository in order to deal with multiple open PRs. All former feedback has been incorporated to the best of my abilities. Furthermore, the number of changes could be reduced to achieve the same goal. 

Before the change, the rich-text-widget includes the CKEditor4 content templates plugin with its default settings, if either the "full" toolbar is chosen, or the "Templates" button is configured in an advanced custom toolbar. However, the default settings offer very limited value, as they would be useful in only very few use cases:

image

So the logical choice would be to either remove the content template plugin from the rich text widget entirely or make the content templates configurable in the Mendix application. Obviously, this pull request aims for the second choice, as we have multiple use cases for the feature in our project.

When the changes are applied, the default behavior of the widget does not change, only the widgets need to be updated once.
Even if an existing app uses the content template plugin for some reason, the default configuration is applied and nothing would change.

Changing the property "Template name" to anything other than "default", allows the developer to provide custom content templates from the app. Therefore, the developer needs to configure a data source to provide the templates and select title, descriptions and html attributes. Custom images are not yet supported due to reviewer concerns.

image

Having configured the content templates, the content templates menu can be filled based on data from the domain model. The user can then add or replaced HTML blocks in the document part handled by the rich text widget:

image

A demo application is available to showcase this feature. Any reviewer can be invited to the project if so desired.

### DIFF
--- a/packages/pluggableWidgets/rich-text-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/rich-text-web/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Added
+
+-   Enabled customization of CKEditor 4 content template plugin.
+
 ## [2.2.3] - 2023-10-27
 
 ### Fixed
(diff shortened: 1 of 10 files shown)


gjulivan:
Hi @tobias-at-work , we are currently working on a major version update for the rich text editor. we will take a look at this PR if it is still relevant by then.
thank you for your contribution!

tobias-at-work:
Will resubmit a pull request for version 3.x

PR #814, opened 2023-11-14
[WC-2200]: Fix virtual scrolling for Gallery Widget

samuelreichert:
### Pull request type

---

### Description
Add a hook to observe when component is visible on screen. Use the hook to determine when useInfiniteScroll should update and recalculate the height.

This hook fixes a weird behavior on Gallery widget where there is no vertical scroll on an infinite scroll gallery

### DIFF
--- a/packages/pluggableWidgets/gallery-web/src/components/__tests__/__mocks__/intersectionObserverMock.ts
+++ b/packages/pluggableWidgets/gallery-web/src/components/__tests__/__mocks__/intersectionObserverMock.ts
@@ -1,0 +1,4 @@
+const intersectionObserverMock = () => ({
+    observe: () => null
+});
+window.IntersectionObserver = jest.fn().mockImplementation(intersectionObserverMock);
--- a/packages/pluggableWidgets/gallery-web/src/components/__tests__/__mocks__/intersectionObserverMock.ts
+++ b/packages/pluggableWidgets/gallery-web/src/components/__tests__/__mocks__/intersectionObserverMock.ts
@@ -1,4 +1,8 @@
-const intersectionObserverMock = () => ({
+type ObserverMock = { observe: () => void };
+type IntersectionObserverMockType = () => ObserverMock;
+
+const intersectionObserverMock: IntersectionObserverMockType = () => ({
     observe: () => null
 });
+
 window.IntersectionObserver = jest.fn().mockImplementation(intersectionObserverMock);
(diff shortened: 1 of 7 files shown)


packages/shared/widget-plugin-grid/src/components/InfiniteBody.tsx
```diff
@@ -58,6 +61,20 @@ export function useInfiniteControl(
     return [trackScrolling, bodySize, containerRef];
 }
 
+export function useOnScreen(ref: RefObject<HTMLElement>): boolean {
```
gjulivan, line 64: First, this attempt of finding is object onScreen, seems like has been tried more than once, 
check this one too: `datagrid-web/src/utils/useIsElementInViewport.tsx`
So I think we need to move this to widget-plugin-hooks instead.

Second, this code doesn't make sense, useEffect runs without dependency that trigger the observer, but the create Observer itself runs with "ref" dependency ? this confusion also being shown on the github action, please fix.
I also found a good code example here: https://www.reacthooklab.com/use-on-screen if you want to take a look

packages/shared/widget-plugin-hooks/src/useOnScreen.ts
```diff
@@ -0,0 +1,25 @@
+import { RefObject, useEffect, useRef, useState } from "react";
+
+export function useOnScreen(ref: RefObject<HTMLElement>): boolean {
+    const [isIntersecting, setIsIntersecting] = useState(false);
+
+    const observer = useRef(
+        new IntersectionObserver(([entry]) => {
+            setIsIntersecting(entry.isIntersecting);
+        })
+    );
```
iobuhov, line 10: Wrap constructor call into function. And, later, you can use `?=` operator to assign it.
This will avoid unnecessary constructor calls on each render.

packages/shared/widget-plugin-hooks/src/useOnScreen.ts
```diff
@@ -0,0 +1,25 @@
...
+    useEffect(() => {
+        observer.current.disconnect();
+
+        if (ref.current) {
+            observer.current.observe(ref.current);
+        }
+
+        return () => {
+            observer.current.disconnect();
+        };
```
iobuhov, line 21: Why we have to call disconnect two times?

packages/pluggableWidgets/gallery-web/src/components/__tests__/__mocks__/intersectionObserverMock.ts
iobuhov, line 1: Can you pleas add return type?

iobuhov (dismissed):
LGTM

leonardomendix (dismissed):

leonardomendix (approved):

iobuhov (approved):

PR #819, opened 2023-11-20
WC-2233 Change Aria label to text template

gjulivan:
### Pull request type

---

### Description
update aria label type to text template


packages/pluggableWidgets/combobox-web/src/Combobox.tsx
```diff
@@ -18,8 +18,8 @@ export default function Combobox(props: ComboboxContainerProps): ReactElement {
         noOptionsText: props.noOptionsText?.value,
         a11yConfig: {
             ariaLabels: {
-                clearSelection: props.clearButtonAriaLabel,
-                removeSelection: props.removeValueAriaLabel
+                clearSelection: props.clearButtonAriaLabel?.value as string,
```
r0b1n, line 21: I think it makes sense to either make those optional in typings or use `?? ""` fallbackm whatever makes more sense. 
I think while `clearSelection` and `removeSelection` might be undefined, ones in `a11yStatusMessage` have to have fallback as otherwise we end up with text like `undefined red,green,blue` in messages in case of `a11ySelectedValue` in not available yet because they got concatenated with other strings.

gjulivan: fix: the value is required, default value is provided. Thus, it will never be undefined or null

r0b1n: This is not about default value. The value still might be undefined when expression is not able to compute, for example the entity used in the textTemplate is unavailable, then the status is unavailable with no value, so it is undefined while `as string` hides the fact.

packages/pluggableWidgets/combobox-web/src/Combobox.xml
```diff
@@ -174,11 +174,17 @@
             <propertyGroup caption="Aria labels">
                 <property key="clearButtonAriaLabel" type="textTemplate" required="true">
                     <caption>Clear selection</caption>
-                    <description />
+                    <description>Used to clear all selected value.</description>
```
r0b1n, line 177: We should iterate on naming/description texts with PM, those are a bit unclear

packages/pluggableWidgets/combobox-web/src/Combobox.tsx
gjulivan: fix: handled using `?? ""`

rahmanunver (approved):

iobuhov (commented):
LGTM

iobuhov (approved):

leonardomendix (approved):

PR #855, opened 2023-12-06
Rename grid column prop

iobuhov:
### Pull request type

---


packages/pluggableWidgets/datagrid-web/src/typings/GridColumn.ts
```diff
@@ -12,7 +12,7 @@ export interface GridColumn {
     canResize: boolean;
     canSort: boolean;
     columnClass(item: ObjectItem): string | undefined;
-    columnId: string;
+    htmlId: string;
```
r0b1n, line 15: I don't think we should have a separate field for the sake of generating html structures

PR #861, opened 2023-12-07
Massive DG state refactoring

iobuhov:
### Pull request type

---

### Description

With this refactoring we migrate to a new library called [effector](https://effector.dev). With this change it's much easer to mange state and side effects. Of course this solution has it's own weak parts, but we get much more value out of it. Beside that I also improved how we restore filters and added option (disabled for now) to save settings to local storage. For now, we just need to merge and test.


packages/pluggableWidgets/datagrid-web/src/components/Header.tsx
```diff
@@ -144,25 +140,27 @@ function useDraggable(
...
     const handleOnDrop = useCallback(
         (e: DragEvent<HTMLDivElement>): void => {
             handleDragEnd();
-            const columnA = colNumFromString(getColNum(e.target as HTMLDivElement));
-            const columnB = colNumFromString(e.dataTransfer.getData(DATA_FORMAT_NUM));
+            const columnA = (e.target as HTMLDivElement).dataset.columnId as ColumnId;
+            const columnB = e.dataTransfer.getData(DATA_FORMAT_ID) as ColumnId;
 
             setColumnOrder(prevOrder => {
                 const indexA = prevOrder.indexOf(columnA);
```
r0b1n, line 166: This feels like this manipulation should happen in the reducer and columnA and columnB have to be passed there.

packages/pluggableWidgets/datagrid-web/src/features/initialization/base.ts
```diff
@@ -0,0 +1,10 @@
+import { GridState } from "../../typings/GridState";
+
+export type InitState = GridState;
+
+export interface InitViewState {
+    filter?: undefined;
+    sortOrder?: unknown[];
+}
+
+export type ComputedInitState = [initState: InitState, initView: InitViewState] | [initState: InitState];
```
r0b1n, line 10: Having `init` states makes it a bit complicated to reason about. Maybe we can think if we can step away from this concept by having the state that possibly contains a field with the current status. Also eliminates a need of `undefined` state as well when loading.

iobuhov: Done as part of migration to effector

packages/pluggableWidgets/datagrid-web/src/components/Header.tsx
iobuhov: Done

packages/pluggableWidgets/datagrid-web/src/Datagrid.editorPreview.tsx
```diff
@@ -55,12 +56,40 @@ export function preview(props: DatagridPreviewProps): ReactElement {
...
+        currentPage: 0,
+        filter: undefined,
+        hidden: new Set<ColumnId>(),
+        order: [],
+        size: {},
+        sort: [],
+        visible: [],
+        splitFilter: [undefined, undefined] as [Filter, Filter]
+    };
+    const noop = (): void => {
```
rahmanunver, line 76: noop?

packages/pluggableWidgets/datagrid-web/src/components/Datagrid.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+export default function Datagrid(props: Props): ReactElement {
+    const id = useRef(`DataGrid${generateUUID()}`);
+    const { FilterContext } = useFilterContext();
+    const { model, actions } = props;
+
+    const viewModel = useViewModel(model);
+    const [columnFilters, headerFilters] = viewModel.splitFilter;
+
+    const exportColumns = useStoreMap(props.model.visible, visible =>
```
rahmanunver, line 31: props, model is already destructured

packages/pluggableWidgets/datagrid-web/src/features/model/bootstrap.ts
```diff
@@ -0,0 +1,87 @@
...
+import { Model, InitParams, Status } from "./base";
+import { Column } from "../../helpers/Column";
+import { StorageDone } from "../storage/base";
+import { paramsFromColumns, paramsFromSettings, sortToInst } from "./utils";
+import { ModelEffects } from "./effects";
+
+type InitArgs = { columns: Column[]; ds: ListValue; storage: StorageDone };
+type InitPayload = [InitArgs, InitParams];
+
+export function bootstrap(
```
rahmanunver, line 13: Correct me if I'm wrong, we create a separate Store to keep the status of the datasource availability? Figuring out what the "bootstrap" implies.

packages/pluggableWidgets/datagrid-web/src/Datagrid.editorPreview.tsx
iobuhov: No operation. This is just to please typescript.

packages/pluggableWidgets/datagrid-web/src/components/Datagrid.tsx
iobuhov: Can you explain a bit more?

packages/pluggableWidgets/datagrid-web/src/features/model/bootstrap.ts
iobuhov: > Correct me if I'm wrong, we create a separate Store to keep the status of the datasource availability?

Well, yes and no. In fact, it's not just about datasources. Is about all props that we are waiting for (columns, datasource, storage, etc).

Bootstrap - is just system startup. Initialization.

packages/pluggableWidgets/datagrid-web/src/components/Datagrid.tsx
rahmanunver: you're using `props.model.visible` , but `model` is destructured already at ln.26

packages/pluggableWidgets/datagrid-web/src/helpers/ColumnBase.ts
```diff
@@ -13,7 +13,7 @@ interface BaseColumnProps {
 }
 
 export class BaseColumn {
-    constructor(private properties: BaseColumnProps) {}
+    constructor(protected properties: BaseColumnProps) {}
```
r0b1n, line 17: I have an idea how we can fix the duplication. This should work.
We can do the following: 
```ts
export class BaseColumn<Props extends BaseColumnProps> {
    constructor(protected props: Props) {}
    ...
}
```

And then in dependent classes remove `props` and it's assignment and use:
```ts
export class Column extends BaseColumn<ColumnsType> implements GridColumn {
    // no assignment to this.props anymore
}
```
and
```ts
export class ColumnPreview extends BaseColumn<ColumnsPreviewType> implements GridColumn {
    // no assignment to this.props anymore
}
```

packages/pluggableWidgets/datagrid-web/src/typings/GridColumn.ts
```diff
@@ -12,11 +15,12 @@ export interface GridColumn {
     canResize: boolean;
     canSort: boolean;
     columnClass(item: ObjectItem): string | undefined;
-    columnId: string;
+    columnId: ColumnId;
     columnNumber: number;
     header: string;
     initiallyHidden: boolean;
     renderCellContent: (item: ObjectItem) => ReactElement;
+    sortInstruction(dir: "asc" | "desc"): SortInstruction | undefined;
```
r0b1n, line 23: Unless I miss something I see that this is used in one place in utils. We can probably inline the logic there and remove `sortInstruction` from the interface.

packages/pluggableWidgets/datagrid-web/src/features/model/bootstrap.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    return $status;
+}
+
+function createInitParams(params: { columns: Column[]; ds: ListValue; storage: StorageDone }): InitParams {
+    const { storage } = params;
+    const settings = storage.status === "ready" ? storage.value.load() : undefined;
+
+    return settings ? paramsFromSettings(settings, params.ds) : paramsFromColumns(params.columns, params.ds);
+}
```
r0b1n, line 87: I think `storage` should be of type `DynamicStorage`, otherwise the check of the `status` field doesn't make sense.

packages/pluggableWidgets/datagrid-web/src/features/model/bootstrap.ts
```diff
@@ -0,0 +1,87 @@
...
+type InitPayload = [InitArgs, InitParams];
+
+export function bootstrap(
+    grid: Model,
+    propsUpdated: Event<DatagridContainerProps>,
+    dataReady: EventCallable<InitParams>,
+    effects: ModelEffects
+): Store<Status> {
+    const payload = createEvent<InitPayload | "pending">();
+    const payloadReady = payload.filterMap(payload => (payload === "pending" ? undefined : payload));
```
r0b1n, line 20: I think we should not use "pending" here as it makes it appear it has something to do with `Status` from line 27 while it is not. Let's try to use `undefined` or `null`, and filter out by that value.

packages/pluggableWidgets/datagrid-web/src/features/model/bootstrap.ts
```diff
@@ -0,0 +1,87 @@
...
+                return "pending";
+            }
+
+            const args = {
+                columns,
+                ds: datasource,
+                storage
+            };
+            return [args, createInitParams(args)] as InitPayload;
+        },
```
r0b1n, line 55: In the current situation typescript is fine with returning any value from line 46 which is not type safe, so changing it to `return "asdfasfasd";` will not trigger any errors.
As for the previous comment, we should not return string literal at all. Let's explicitly define return type (`InitPayload | undefined`) of `fn` then we don't have to cast at line 54 and we get type safety back.

packages/pluggableWidgets/datagrid-web/src/features/model/bootstrap.ts
```diff
@@ -0,0 +1,87 @@
...
+import { paramsFromColumns, paramsFromSettings, sortToInst } from "./utils";
+import { ModelEffects } from "./effects";
+
+type InitArgs = { columns: Column[]; ds: ListValue; storage: StorageDone };
+type InitPayload = [InitArgs, InitParams];
+
+export function bootstrap(
+    grid: Model,
+    propsUpdated: Event<DatagridContainerProps>,
+    dataReady: EventCallable<InitParams>,
```
r0b1n, line 16: I guess something like `initParamsReady` is a clearer name for this argument.

packages/pluggableWidgets/datagrid-web/src/components/Header.tsx
```diff
@@ -26,28 +26,26 @@ export interface HeaderProps {
     filterable: boolean;
     filterWidget?: ReactNode;
     draggable: boolean;
-    dragOver: string;
+    dragOver: ColumnId | undefined;
```
gjulivan, line 29: why not just make it optional.
`dragOver?: ColumnId`

packages/pluggableWidgets/datagrid-web/src/features/model/grid-model.ts
```diff
@@ -0,0 +1,240 @@
...
+import { ColumnsType, DatagridContainerProps } from "../../../typings/DatagridProps";
+import { Column } from "../../helpers/Column";
+import { ColumnId } from "../../typings/GridColumn";
+import * as Grid from "../../typings/GridModel";
+import { storageUnit } from "../storage/storage-model";
+import { Model, InitParams } from "./base";
+import { getHash, sortByOrder } from "./utils";
+import { createFilter } from "../filters/main";
+
+const USE_MULTI_SORT = false;
```
gjulivan, line 11: we probably need some common place for this feature flag constant variables.
(i see we also have `ENABLE_LOCAL_STORAGE` flag)

packages/pluggableWidgets/datagrid-web/src/features/model/bootstrap.ts
```diff
@@ -0,0 +1,87 @@
...
+        target: setViewStateFx
+    });
+
+    // End bootstrap on next reload
+    sample({
+        clock: propsUpdated,
+        source: $status,
+        filter: status => status === "waitingDatasource",
+        target: bootstrapEnd
+    });
```
r0b1n, line 77: If I understand correctly, this assumes that the next properties update is always a data source update and always updates to the 'available' status of the data source. This might not be the case as properties might update for plenty of reasons, so the status going to be off.

packages/pluggableWidgets/datagrid-web/src/features/model/grid-model.ts
```diff
@@ -0,0 +1,240 @@
...
+        fn: reduceHidden
+    });
+
+    sample({
+        clock: [initValue, newValue],
+        source: $available,
+        // Pass new value only if at least 1 column is visible
+        filter: (available, nextHidden) => {
+            const visibleCount = available.reduce((n, column) => (nextHidden.has(column.columnId) ? n : n + 1), 0);
+            return visibleCount > 0;
```
r0b1n, line 227: Looks like we can return early as we don't need exact count, something like `return available.some(c => !nextHidden.has(c.columnId));`

packages/pluggableWidgets/datagrid-web/src/features/model/grid-model.ts
```diff
@@ -0,0 +1,240 @@
...
+        },
+        fn: (_, next) => next,
+        target: $hidden
+    });
+
+    sample({
+        source: $visible,
+        filter: visible => visible.length === 0,
+        target: resetHidden
+    });
```
r0b1n, line 237: Maybe drop a comment line what the logic behind this (iiuc, it resets hidden state if all visible column are hidden by external factors).

packages/pluggableWidgets/datagrid-web/src/features/model/grid-model.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    // When setPage is called one of this event will be emitted,
+    // depending on current pagination type.
+    const limitChanged = createEvent<number>();
+    const offsetChanged = createEvent<number>();
+    const cleanup = createEvent<unknown>();
+
+    const $columns = createStore<ColumnsType[]>([])
+        .on(propsUpdate, (_, props) => props.columns)
+        .map<Column[]>(columns => columns.map((columnsType, index) => new Column(columnsType, index)));
```
r0b1n, line 41: I see that `ColumnType[]` is never used directly, but mapped to `Column[]`. Is there a benefit for doing so instead of mapping in `.on()` directly?

packages/pluggableWidgets/datagrid-web/src/helpers/Column.tsx
```diff
@@ -1,21 +1,19 @@
...
 
-    constructor(props: ColumnsType, columnNumber: number, gridId: string) {
+    constructor(props: ColumnsType, columnNumber: number) {
         super(props);
 
         this.props = props;
-        this.gridId = gridId;
         this.columnNumber = columnNumber;
 
         this.visible = props.visible?.value ?? false;
```
r0b1n, line 19: Let's move this into a getter.

packages/pluggableWidgets/datagrid-web/src/features/model/grid-model.ts
```diff
@@ -0,0 +1,240 @@
...
+        .on(sortBy, USE_MULTI_SORT ? reduceSortMulti : reduceSortSingle)
+        .on(hide, reduceSortRemove)
+        .on(paramsReady, (_, params) => params.sort);
+
+    const $available = combine($columns, $order, (columns, order) => {
+        if (order.length === columns.length) {
+            columns = sortByOrder(columns, order);
+        }
+        return columns.filter(column => column.visible);
+    });
```
r0b1n, line 64: In which case lengths are not equal?

packages/pluggableWidgets/datagrid-web/src/features/model/grid-model.ts
```diff
@@ -0,0 +1,240 @@
...
+    events: Grid.Events;
+} {
+    const hide = createEvent<ColumnId>();
+    const sortBy = createEvent<ColumnId>();
+    const swap = createEvent<[a: ColumnId, b: ColumnId]>();
+    const resize = createEvent<[id: ColumnId, size: number]>();
+    const setPage = createEvent<number>();
+    const mapPage = createEvent<(n: number) => number>();
+    const nextPage = createEvent<unknown>();
+    const prevPage = createEvent<unknown>();
```
r0b1n, line 30: We probably can use only one event which is now called `mapPage`. Other actions don't need a separate event I think.
I would also get rid on `mapPage` action name, and let `setPage` be able to use both approaches: `(currentPage: number) => number) | number`.

packages/pluggableWidgets/datagrid-web/src/features/filters/header-filters-model.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export function createHeaderFilters(
+    _paramsReady: Event<InitParams>
+): [composedFilter: Store<Filter>, setFilter: SetHeaderFilter] {
+    const setFilter = createEvent<FilterState>();
+
+    const $filters = createStore<HeaderFilters>(new Map()).on(setFilter, (state, filterState) => {
+        if (!filterState.filterType) {
+            return state;
+        }
```
r0b1n, line 17: In which case `filterType` is absent and what does it mean in business logic terms?

packages/pluggableWidgets/datagrid-web/src/features/model/bootstrap.ts
iobuhov: Yeah, agree, we should avoid "pending" and be clear about types, however, I don't think changing return type to "asdfadf" will be allowed. So, types working, it's just me "gambling" removing "pending" should solve the issues.

packages/pluggableWidgets/datagrid-web/src/features/model/bootstrap.ts
iobuhov: Good catch 👍

packages/pluggableWidgets/datagrid-web/src/features/model/grid-model.ts
iobuhov: Yes, I can do it 👍

packages/pluggableWidgets/datagrid-web/src/features/model/bootstrap.ts
iobuhov: Yes, my assumption about datasource update is very weak, it's indeed could be any update for any other reason. But, this is the best model I came so far. In general I'm not quite happy how this "bootstrap" logic came out - it has obvious disadvantages and code is hard to read. If you know better approach, we can talk about how to implement it, but for now, we have to "assume" that next "update" is coming after we called `setViewStateFx`

packages/pluggableWidgets/datagrid-web/src/features/model/grid-model.ts
iobuhov: Good one, I don't remember 😩 
I think this is just left over from my attempt to "recover" from "saved" settings, but when config is changed.
I would better remove this part as I really not sure why it here.

packages/pluggableWidgets/datagrid-web/src/features/model/grid-model.ts
iobuhov: Oh, you right! 🤩

packages/pluggableWidgets/datagrid-web/src/features/model/grid-model.ts
iobuhov: Yes, exactly, while I was testing I found weird case. In short, You can "hide" some columns and reload the page, but once you do it, some columns may already become "unavailable" and in fact we end up in situation where we have no visible columns which is very confusing for the end user. And to avoid this state, we reset "hidden" store if we have 0 visible columns.

packages/pluggableWidgets/datagrid-web/src/features/filters/header-filters-model.ts
```diff
@@ -0,0 +1,38 @@
...
+import { InitParams } from "../model/base";
+
+type HeaderFilters = Map<FilterType, FilterState>;
+
+export function createHeaderFilters(
+    _paramsReady: Event<InitParams>
+): [composedFilter: Store<Filter>, setFilter: SetHeaderFilter] {
+    const setFilter = createEvent<FilterState>();
+
+    const $filters = createStore<HeaderFilters>(new Map()).on(setFilter, (state, filterState) => {
```
r0b1n, line 14: I suggest improving naming for the sake of readability. Something like `state` -> `currentHeaderFilters`, `filterState` -> `newFilter`, `next` -> `newHeaderFilters`.

packages/pluggableWidgets/datagrid-web/src/features/storage/AttrStorage.ts
```diff
@@ -0,0 +1,25 @@
...
+export class AttrStorage implements SettingsStorage {
+    constructor(private attr: EditableValue<string>, private hash: string) {}
+
+    load(): GridSettings | undefined {
+        if (typeof this.attr.value === "string" && this.attr.value.length > 0) {
+            const data: GridSettings = JSON.parse(this.attr.value);
+            if (data.settingsHash === this.hash) {
+                return data;
+            }
+            this.reset();
```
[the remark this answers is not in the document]
iobuhov: Yeah, good question.
In short, I think yes.
The case is that, if hash don't match, then probably config is outdated and we have to reset before we can save new config.

packages/pluggableWidgets/datagrid-web/src/features/storage/utils.ts
```diff
@@ -0,0 +1,36 @@
+import { LocalSettingsStorage } from "./LocalSettingsStorage";
+
+const NS = "com.mendix.widgets.web.datagrid.storageKeys";
```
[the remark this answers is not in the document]
iobuhov: NS - namespace.
But, you probably right, I'll rename it.

packages/pluggableWidgets/datagrid-web/src/helpers/Column.tsx
iobuhov: Yes

packages/pluggableWidgets/datagrid-web/src/typings/ColumnSettings.ts
```diff
@@ -0,0 +1,8 @@
+import { ColumnId } from "./GridColumn";
+
+export interface ColumnSettings {
+    columnId: ColumnId;
+    size: number | undefined;
+    hidden: boolean;
+    filterSettings: unknown;
```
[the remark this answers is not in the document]
iobuhov: For now we going to keep it as `unknown` to avoid any use of this prop. This is intentional.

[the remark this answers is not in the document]
iobuhov: We going to change this type once we add ability to store filters and filter values along with config.

packages/pluggableWidgets/datagrid-web/src/features/filters/header-filters-model.ts
iobuhov: I have no idea. I just left it as it is. But I think this is just "type guard" to avoid `undefined`. The problem is that `FilterState` used for both, columns and header and `filterType` is optional...
Any way, I'm not sure we really need this.

packages/pluggableWidgets/datagrid-web/src/features/model/grid-model.ts
iobuhov: I agree that we can reduce all this events to just `setPage`. I think I'll remove `mapPage`, but I'll leave `nextPage` and `prevPage` for convenience.
Having this events make it bit simple:
instead of `() => setPage(n => n + 1)` you can just pass `nextPage`

But, I can do it in a way that `nextPage` and `prevPage` will be just "sugar" on top of `setPage`. So, `setPage` will be main way of changing pages.

PR #872, opened 2023-12-20
[WC-2046]: Add keyboard navigation support to gallery-web

samuelreichert:
### Pull request type

---

### Description

Add keyboard navigation support to gallery to improve user experience and a11y support.


packages/pluggableWidgets/datagrid-web/src/features/keyboard-navigation/useFocusTargetController.ts
iobuhov, line 1: I think you can move this hook to plugin as well.

PR #873, opened 2023-12-21
[WC-910] feat(datagrid-web): double click action

iobuhov:
### Pull request type

---

### Description

Data Grid 1 has a double-click option. This PR adds the double-click option to the DataGrid 2. This will allow developers to replace old data grids with matching functionality.

The double click will be even more effective when the DG2 can emit events for the data view Data source Type "Listen to widget."

An option to consider is to add an extra action for "onClickDouble.". Though this might conflict with the single action, both actions might delay the click action.


packages/pluggableWidgets/datagrid-web/src/helpers/useRowInteractionProps.ts
samuelreichert, line 1: Should we change the name of the file? `use-row-interaction-props`?

packages/pluggableWidgets/datagrid-web/src/features/keyboard-navigation/useFocusTargetProps.ts
samuelreichert, line 1: Should we change the name of the file? `use-focus-target-props`?

packages/pluggableWidgets/datagrid-web/src/features/row-interaction/base.ts
```diff
@@ -0,0 +1,19 @@
...
+    selectionType: SelectionType;
+    clickTrigger: ClickTrigger;
+}
+
+export interface CheckboxContext {
+    item: ObjectItem;
+    pageSize: number;
+    selectionType: SelectionType;
+    selectionMethod: SelectionMethod;
+}
```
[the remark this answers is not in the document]
iobuhov: Done

packages/pluggableWidgets/datagrid-web/src/helpers/useRowInteractionProps.ts
iobuhov: This hook is removed

packages/pluggableWidgets/datagrid-web/src/features/keyboard-navigation/useFocusTargetProps.ts
iobuhov: Yes, but I think it's better to do in you pr, wile you moving staff around.

packages/pluggableWidgets/datagrid-web/src/helpers/use-select-action-helper.ts
```diff
@@ -0,0 +1,25 @@
+import { SelectionHelper } from "@mendix/widget-plugin-grid/selection";
```
gjulivan, line 1: why we need to separate this if this is only a hook wrapper for SelectActionHelper ?
we can combine that on the same file for now with "SelectActionHelper"
unless they serve different helper purpose?

similar with ClickActionHelper and use-click-action-helper.

packages/pluggableWidgets/gallery-web/src/features/item-interaction/use-item-events-controller.ts
```diff
@@ -0,0 +1,15 @@
+import { useMemo } from "react";
```
gjulivan, line 1: combine file with ItemEventsController ?

gjulivan:
i see here that you started using kebab-case for hooks file naming, is this new file naming convention?
this is seems inconsistent.
don't we have a rule for this?
camelCase for non-react component.
PascalCase for react component.

iobuhov:
> i see here that you started using kebab-case for hooks file naming, is this new file naming convention? this is seems inconsistent. don't we have a rule for this? camelCase for non-react component. PascalCase for react component.

There no rules yet.
My strategy:
PascalCase for classes and components
desh-case for other files

gjulivan:
> > i see here that you started using kebab-case for hooks file naming, is this new file naming convention? this is seems inconsistent. don't we have a rule for this? camelCase for non-react component. PascalCase for react component.
> 
> There no rules yet. My strategy: PascalCase for classes and components desh-case for other files

kebab-case (dash) is already being used for folder naming,
and all of the other hooks file is already using camelCase (similar to react hooks function naming convention)

adding another naming convention here is too much.
why the sudden change?

packages/pluggableWidgets/datagrid-web/src/helpers/use-widget-props.ts
r0b1n, line 1: This file uses kebab-case

packages/pluggableWidgets/datagrid-web/src/features/keyboard-navigation/useFocusTargetProps.ts
r0b1n, line 1: This file uses camelCase. I guess we we lack some consistency.

packages/pluggableWidgets/datagrid-web/src/helpers/use-select-action-helper.ts
iobuhov: Sure, I'll move it

packages/pluggableWidgets/datagrid-web/src/features/keyboard-navigation/useFocusTargetProps.ts
iobuhov: As we agreed, this will stay camelCasse.

r0b1n (approved):

leonardomendix (approved):

PR #874, opened 2023-12-21
[WC-2247]: PopupMenu: change the element class is applied

rahmanunver:
### Pull request type

---

### Description
After switching to React Portal, the element we apply the class has changed. We are reverting the change, and applying it on the PopupMenu widget itself, rather than the child portal that pops up.

See which element the class is applied. Iy should be on the wrapping widget, not the menu that pops up.


gjulivan (approved):

leonardomendix (approved):

samuelreichert (approved):
LGTM

PR #880, opened 2023-12-29
[WC-1659] Add new workflow for RC testing

leonardomendix:
### Pull request type

---

### Description

This is the first step to enable the E2E automation to run against an RC build.


iobuhov (approved):
LGTM. But I would remove ls commands after few "test fly" runs

rahmanunver (approved):

PR #881, opened 2024-01-03
[WC-2281]: Combobox losing focus on item selection (single selection)

rahmanunver:
### Pull request type

---

### Description
We had a blur() method being used inside onSelectedItemChange of the downshift configuration. Reverting it solves the issue.

Currently the blur() is called on combobox when an item is selected. 
With this PR the focus on combobox shouldn't be lost when an item is selected (single selection).

### DIFF
--- a/packages/pluggableWidgets/combobox-web/src/hooks/useDownshiftSingleSelectProps.ts
+++ b/packages/pluggableWidgets/combobox-web/src/hooks/useDownshiftSingleSelectProps.ts
@@ -118,7 +118,9 @@ export function useDownshiftSingleSelectProps(
         selectedItem: selector.currentValue
     });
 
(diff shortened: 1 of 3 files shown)


packages/pluggableWidgets/combobox-web/src/hooks/useDownshiftSingleSelectProps.ts
iobuhov, line 1: Should we fix eslint warning?

samuelreichert (dismissed):
LGTM

leonardomendix (approved):

iobuhov (approved):
LGTM 👍

samuelreichert (approved):

PR #884, opened 2024-01-08
fix(a11y): add role combobox

bsgriggs:
### Pull request type

---

### Description
added role 'combobox' per a11y requirements https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#combobox


iobuhov:
We need to update snapshots.

gjulivan:
role combobox already exist in the input.
https://github.com/mendix/web-widgets/blob/b0ef811e6c239d895b52b5d079d0c2644fd19ca5/packages/pluggableWidgets/combobox-web/src/__tests__/__snapshots__/SingleSelection.spec.tsx.snap#L28C11-L28C26

PR #895, opened 2024-01-17
[WC-2300]: DG2: fix not exporting dynamicText

rahmanunver:
### Pull request type

---

### Description
Added a check to see if dynamic texts are returning `"undefined"` with `status:"loading"`, so that we can wait and update the value in next datasource update.

Try exporting a dg2 with dynamicText column items, at its current state it will export empty (`""`) string. 
With this change it should be exporting it too.

### DIFF
--- a/packages/pluggableWidgets/datagrid-web/src/features/export.ts
+++ b/packages/pluggableWidgets/datagrid-web/src/features/export.ts
@@ -219,24 +219,49 @@ function exportColumns(columns: ColumnsType[]): Message {
     return { type: "columns", payload: exportColumns };
 }
 
-function exportData(data: ObjectItem[], columns: ColumnsType[]): Message {
-    const items = data.map(item => {
-        return columns.map(column => {
-            let value = "";
+type ExportDataResult =
+    | {
+          status: "pending";
+      }
+    | {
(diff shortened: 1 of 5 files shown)


packages/pluggableWidgets/datagrid-web/src/features/export.ts
```diff
@@ -219,24 +219,49 @@ function exportColumns(columns: ColumnsType[]): Message {
...
+                    hasLoadingItem = true;
+                } else if (dynamicText?.status === "unavailable") {
+                    value = "n/a";
+                } else {
+                    value = dynamicText?.value ?? "";
+                }
             } else {
                 value = "n/a (custom content)";
             }
+            row.push(value);
```
gjulivan, line 253: why we can't keep the data.map and column.map?

iobuhov: @rahmanunver , I think this proves that we have to reveit code back to map of map 😄

iobuhov (approved):
LGTM

samuelreichert (approved):

leonardomendix (approved):

PR #905, opened 2024-01-24
[WC-2288] DataGrid: fix column resizing 

iobuhov:
### Pull request type

---

### Description
In this PR I'm trying to fix colum resizing. Right now we resize only single column and let rest to be calculated by CSS. This approach give "stretch" feeling on resize (by resizing on column, all other columns resize as well). As a solution I'm trying to set fixed width for each column on resize start.

- Column resizing

Before:
before
---
After:
after

### DIFF
--- a/packages/pluggableWidgets/datagrid-web/src/Datagrid.editorPreview.tsx
+++ b/packages/pluggableWidgets/datagrid-web/src/Datagrid.editorPreview.tsx
@@ -111,7 +111,9 @@ export function preview(props: DatagridPreviewProps): ReactElement {
                 resize: noop,
                 setFilter: noop,
                 sortBy: noop,
-                swap: noop
+                swap: noop,
+                setSize: noop,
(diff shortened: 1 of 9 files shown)


packages/pluggableWidgets/datagrid-web/src/Datagrid.editorConfig.ts
```diff
@@ -104,7 +110,7 @@ export function getProperties(
                     column.width === "autoFill"
                         ? "Auto-fill"
                         : column.width === "autoFit"
-                        ? "Auto-fit content"
+                        ? `Auto-fit content`
```
samuelreichert, line 113: ?

samuelreichert (dismissed):
LGTM

leonardomendix (approved):

samuelreichert (approved):

leonardomendix (approved):

PR #906, opened 2024-01-25
[WC-2335] (web-action): fix csp for web action

gjulivan:
move style into separate file.

impact:
style for TakePicture will always bundled into theme.compile.css
it will not be lazyloaded only when take picture is being called.

but this solve the issue with CSP style-src: unsafe-inline


samuelreichert (dismissed):
LGTM

iobuhov (approved):

iobuhov (approved):

leonardomendix (approved):

PR #915, opened 2024-02-01
[WC-941] feat: reset filters

iobuhov:
### Pull request type

---

### Description
This PR adds new plugin that focus on external events for data-widgets. Also, it introduce two new actions: Reset_All_Filters and Reset_Filter. This actions can be called from Nanoflow.

Reset_All_Filters
Reset_Filter

- Gallery filters
- Data grid 2 filters


packages/modules/data-widgets/scripts/steps/bundle-export-to-excel.ts
```diff
@@ -28,7 +28,7 @@ const xlsxExportTools: Entry = {
 };
 
 export async function bundleExportToExcelAction(params: ModuleStepParams): Promise<void> {
```
gjulivan, line 30: now. this function needs to be renamed, right?

iobuhov: yes

gjulivan (dismissed):

leonardomendix (approved):

gjulivan (approved):

PR #916, opened 2024-02-01
[WC-2155]: Combobox Database Lists

rahmanunver:
### Pull request type

---

### Description
Adding Database List feature in the Combobox with new helper classes. 
Widget config has new layout for selection between Context (Current structure for association/enum/bool) and a new option for Database.

Addition of the new configs should be clear enough to build a combobox with Database option selected, we want to see the behaviour similar with previous (association) configurations, and the new feature should also handle the cases and issues which are as a result of setting non-existing values.


packages/pluggableWidgets/combobox-web/src/helpers/Database/DatabaseValuesProvider.ts
```diff
@@ -0,0 +1,41 @@
...
+            return this.emptyValue?.value;
+        }
+        const value = this.attribute.get(item);
+        if (value.status === "unavailable") {
+            return this.emptyValue?.value;
+        }
+        return value.value;
+    }
+
+    getEmptyValue() {
```
gjulivan, line 38: return type: string | Big | undefined

packages/pluggableWidgets/combobox-web/src/helpers/Database/DatabaseCaptionsProvider.tsx
```diff
@@ -0,0 +1,103 @@
...
+    customContentType: OptionsSourceAssociationCustomContentTypeEnum;
+    caption?: string;
+}
+
+interface CaptionContentProps extends PropsWithChildren {
+    htmlFor?: string;
+    onClick?: (e: MouseEvent) => void;
+}
+
+export function CaptionContent(props: CaptionContentProps): ReactElement {
```
gjulivan, line 19: i think we should export this to another file,
there's also import for CaptionContent in SelectAllButton.tsx and icons.tsx,
which means, that this component has been reused several times, 
thus, a common probably moved it under /components/ make sense.

gjulivan (approved):

gjulivan (dismissed):

leonardomendix (approved):

leonardomendix (approved):

gjulivan (approved):

PR #921, opened 2024-02-05
[WC-2175]: Gallery Multi Selection

samuelreichert:
### Pull request type

---

### Description

Adding a new feature to Gallery: You can use the keyboard to select multiple items inside the gallery, just as simple as shift + arrow keys. You can also unselect an item with shift+space.
* Shift + arrow up
* Shift + arrow down
* Shift + arrow left
* Shift + arrow right
* Shift + Home
* Shift + End
* Shift + Page Up
* Shift + Page Down

### DIFF
--- a/packages/pluggableWidgets/gallery-web/src/components/ListItem.tsx
+++ b/packages/pluggableWidgets/gallery-web/src/components/ListItem.tsx
@@ -25,7 +25,6 @@ const component = memo(function ListItem(props: ListItemProps): ReactElement {
     const { columnIndex, rowIndex } = getPosition(itemIndex);
     const keyNavProps = useFocusTargetProps({ columnIndex: columnIndex ?? -1, rowIndex });
     const handlers = useMemo(() => eventsController.getProps(item), [eventsController, item]);
-    console.info("Listview has handlers", { handlers });
 
     return (
         <div
--- a/packages/pluggableWidgets/gallery-web/src/components/ListItem.tsx
+++ b/packages/pluggableWidgets/gallery-web/src/components/ListItem.tsx
@@ -1,6 +1,6 @@
 import classNames from "classnames";
 import { ObjectItem } from "mendix";
-import { createElement, memo, ReactElement, useMemo, RefObject } from "react";
+import { createElement, ReactElement, useMemo, RefObject } from "react";
 import { useFocusTargetProps } from "@mendix/widget-plugin-grid/keyboard-navigation/useFocusTargetProps";
 import { getAriaProps } from "../features/item-interaction/get-item-aria-props";
 import { Positions } from "../features/useGridPositions";
@@ -17,8 +17,7 @@ type ListItemProps = Omit<JSX.IntrinsicElements["div"], "ref" | "role"> & {
     eventsController: ItemEventsController;
 };
 
-// eslint-disable-next-line prefer-arrow-callback
-const component = memo(function ListItem(props: ListItemProps): ReactElement {
+export function ListItem(props: ListItemProps): ReactElement {
     const { eventsController, getPosition, helper, item, itemIndex, selectHelper, ...rest } = props;
     const clickable = helper.hasOnClick(item) || selectHelper.selectionType !== "None";
     const ariaProps = getAriaProps(item, selectHelper);
@@ -50,6 +49,4 @@ const component = memo(function ListItem(props: ListItemProps): ReactElement {
             {helper.render(item)}
         </div>
     );
-});
-
-export const ListItem = component as (props: ListItemProps) => ReactElement;
+}
--- a/packages/pluggableWidgets/gallery-web/src/components/ListItem.tsx
+++ b/packages/pluggableWidgets/gallery-web/src/components/ListItem.tsx
@@ -1,6 +1,6 @@
 import classNames from "classnames";
 import { ObjectItem } from "mendix";
-import { createElement, memo, ReactElement, useMemo, RefObject } from "react";
+import { createElement, ReactElement, useMemo, RefObject } from "react";
 import { useFocusTargetProps } from "@mendix/widget-plugin-grid/keyboard-navigation/useFocusTargetProps";
 import { PositionInGrid } from "@mendix/widget-plugin-grid/selection";
 import { getAriaProps } from "../features/item-interaction/get-item-aria-props";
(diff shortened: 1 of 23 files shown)


packages/pluggableWidgets/gallery-web/src/components/ListItem.tsx
```diff
@@ -17,8 +17,7 @@ type ListItemProps = Omit<JSX.IntrinsicElements["div"], "ref" | "role"> & {
     eventsController: ItemEventsController;
 };
 
-// eslint-disable-next-line prefer-arrow-callback
-const component = memo(function ListItem(props: ListItemProps): ReactElement {
+export function ListItem(props: ListItemProps): ReactElement {
```
iobuhov, line 20: Why `memo` is removed?

samuelreichert: @iobuhov The list item selection (classes) were not kept up to date with the actual keyboard selection

packages/shared/widget-plugin-grid/src/selection/helpers.ts
```diff
@@ -277,3 +296,22 @@ function objectListEqual(a: ObjectItem[], b: ObjectItem[]): boolean {
     const setB = new Set(b.map(obj => obj.id));
     return a.every(obj => setB.has(obj.id));
 }
+
+export type Positions = {
+    columnIndex: number;
+    rowIndex: number;
+};
+
+function getRowIndex(index: number, numberOfColumns: number): number {
```
iobuhov, line 305: 4. (NIT) Rename this funciton to `getRowBasedOnItemIndex` and add jsdoc comment explainging what is doing.

packages/shared/widget-plugin-grid/src/selection/helpers.ts
```diff
@@ -277,3 +296,22 @@ function objectListEqual(a: ObjectItem[], b: ObjectItem[]): boolean {
...
+    return Math.floor(index / numberOfColumns);
+}
+
+export function getPosition(numberOfColumns: number, totalItems: number, index: number): Positions {
```
iobuhov, line 309: 5. (NIT) Rename this funciton to `getColumnAndRowBasedOnIndex` and add jsdoc comment explainging what is doing.

packages/shared/widget-plugin-grid/src/selection/helpers.ts
```diff
@@ -277,3 +296,22 @@ function objectListEqual(a: ObjectItem[], b: ObjectItem[]): boolean {
     const setB = new Set(b.map(obj => obj.id));
     return a.every(obj => setB.has(obj.id));
 }
+
+export type Positions = {
```
iobuhov, line 300: 6. (NIT) Rename this type to `PositionInGrid` (or something with grid in name).

packages/shared/widget-plugin-grid/src/selection/helpers.ts
iobuhov, line 1: 7. As we're about to introduce a new method, it's a good idea to do the same thing for `SelectActionHelper` - go to the implementation and add a new method `onSelectAdjacentGrid`.

iobuhov (changes requested):
All in all, PR looks good, but I think we need to split the code into methods to avoid misuse and overload - methods should be simple.

packages/pluggableWidgets/gallery-web/src/components/ListItem.tsx
iobuhov: Ah, I see, but then can we split this component into "container" and "view" and keep the view part wrapped with in memo?

packages/shared/widget-plugin-grid/src/selection/types.ts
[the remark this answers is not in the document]
samuelreichert: In this case I will need both `size` and `numberOfColumns`. For arrow navigation, I still need size, but for the other commands ( "page up", "home", ...) then I can use numberOfColumns.

leonardomendix (dismissed):

packages/shared/widget-plugin-grid/src/selection/helpers.ts
```diff
@@ -277,3 +306,51 @@ function objectListEqual(a: ObjectItem[], b: ObjectItem[]): boolean {
...
+    rowIndex: number;
+};
+
+/**
+ * Given an index and number of columns,
+ * this function returns which row the index is located in a grid.
+ * @function getRowBasedOnItemIndex
+ * @param {number} index - The index used to find the row.
+ * @param {number} numberOfColumns - The number of columns of the grid.
+ * @return {number}
```
iobuhov, line 321: Wow, nice 👍

packages/shared/widget-plugin-grid/src/selection/helpers.ts
```diff
@@ -113,15 +113,14 @@ export class MultiSelectionHelper {
...
+            this.selectionValue.setSelection(itemsToAdd);
             return;
         }
 
         const itemsToRemove = this._getRange(start, currentEnd);
         const itemsToAdd = this._getRange(start, nextEnd);
 
-        let selection = [...this.selectionValue.selection];
+        let selection: ObjectItem[] = [];
         selection = this._diff(selection, itemsToRemove);
```
iobuhov, line 124: There one little thing that I'm afraid of - what about checkboxes in datagrid?
I suspect this change will brake current behavior. But, let's see. I'll approve.
@leonardomendix, can you please test Data Grid as well and report any unusual stfuff?

leonardomendix: Already checked and it doesn't seem to be affected. Thanks for the heads up.

gjulivan (dismissed):

leonardomendix (dismissed):

leonardomendix (approved):

gjulivan (dismissed):

leonardomendix (approved):

packages/shared/widget-plugin-grid/src/selection/__tests__/helpers.spec.ts
```diff
@@ -37,7 +37,7 @@ describe("MultiSelectionHelper", () => {
             expect(gridState()).toEqual([x, _, x, _, x, _, _, _]);
 
             helper.selectUpTo(items[7]);
-            expect(gridState()).toEqual([x, _, x, _, x, x, x, x]);
+            expect(gridState()).toEqual([_, _, _, _, x, x, x, x]);
```
iobuhov, line 40: Can we move this test out of this block?
We have new behavior that don't match previous. Can we add just one more block to explain new behavior?

iobuhov (changes requested):

iobuhov (approved):
LGTM

PR #924, opened 2024-02-08
[WC-2347][WC-2348] Disable version check in the CKEditor config, add validation message

leonardomendix:
### Pull request type

---

### Description

Since yesterday we started to receive one alert from the CKEditor 4, we're disabling the version check in the CKEditor config to avoid this unnecessary alert.

Automation should pass.


gjulivan (dismissed):

samuelreichert (dismissed):

packages/pluggableWidgets/rich-text-web/src/components/RichText.tsx
```diff
@@ -30,6 +31,7 @@ export function RichText(props: RichTextContainerProps): JSX.Element | null {
         >
             <div ref={setElement} />
             {element ? <Editor element={element} widgetProps={props} /> : null}
+            {stringAttribute.validation ? <ValidationAlert>{stringAttribute.validation}</ValidationAlert> : null}
```
r0b1n, line 34: Likely can be rendered wihout the check

r0b1n (approved):

PR #951, opened 2024-03-18
[WC-2408]: Update datagrid-date-filter-web react-datepicker

samuelreichert:
### Pull request type

---

### Description

To fix Jira ticket WC-2328, we are updating react-datepicker to latest version (6.4.0) which uses last version of date-fns and floating-ui instead of popper-js. Also, the datepicker uses `useAdditionalWeekYearTokens` set to `true` which fixes some issues that caused this Jira ticket.

### DIFF
--- a/packages/pluggableWidgets/datagrid-date-filter-web/src/components/DatePicker.tsx
+++ b/packages/pluggableWidgets/datagrid-date-filter-web/src/components/DatePicker.tsx
@@ -15,7 +15,8 @@ import DatePickerComponent from "react-datepicker";
 import classNames from "classnames";
 import { isDate, isValid } from "date-fns";
 import replaceAllInserter from "string.prototype.replaceall";
-import { doubleMonthOrDayWhenSingle } from "../utils/utils";
+import { offset } from "@floating-ui/dom";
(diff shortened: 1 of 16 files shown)


gjulivan (dismissed):

packages/pluggableWidgets/datagrid-date-filter-web/src/components/DatePicker.tsx
```diff
@@ -108,14 +111,7 @@ export const DatePicker = forwardRef(
...
                     }}
-                    popperModifiers={[
-                        {
-                            name: "offset",
-                            options: {
-                                offset: [0, 0]
-                            }
-                        }
-                    ]}
+                    popperModifiers={[offset(0)]}
```
gjulivan, line 114: is this possible to do without directly importing floating UI?
if the offset is 0, isn't it default value? can we simply remove this and all the dependencies from floating UI?
(even if we need to migrate popper to floating UI later on)

gjulivan (approved):

leonardomendix (approved):

gjulivan (approved):

PR #952, opened 2024-03-19
fix: unexpected behaviour on ios safari

HedwigAR:
### Pull request type

---

### Description
A weird issue occurred on Safari iOS with the take picture action, this issue does not occur in the examples of the library we're using (https://webrtc.github.io/samples/)
This PR changes the configuration of the video to make it more explicit what we're setting there. So far this seems to work and gets rid of that black bar that covers half the screen.

On iOS Safari browser, this should render the video without a black bar covering half the screen.


gjulivan (changes requested):
add patch version increase and changelog.

gjulivan (approved):

iobuhov (approved):

PR #954, opened 2024-03-19
[WC-2381] fix: input text does not work in the combobox footer area

gjulivan:
### Pull request type

---

### Description

To prevent menu closing directly on clicking, we previously added event propagate stopper in menu footer click event.
But apparently, this is also stopping element getting focus.
thus, this PR adding focus to each element target clicked on footer click event.

note for testing: please also validate with multi-selection and select all button.

### DIFF
--- a/packages/pluggableWidgets/combobox-web/src/components/ComboboxMenuWrapper.tsx
+++ b/packages/pluggableWidgets/combobox-web/src/components/ComboboxMenuWrapper.tsx
@@ -16,8 +16,6 @@ interface ComboboxMenuWrapperProps extends PropsWithChildren, Partial<UseCombobo
 }
 
 function PreventMenuCloseEventHandler(e: React.MouseEvent): void {
-    (e.target as HTMLElement)?.focus();
-    e.preventDefault();
     e.stopPropagation();
 }
 
@@ -55,6 +53,7 @@ export function ComboboxMenuWrapper(props: ComboboxMenuWrapperProps): ReactEleme
                 <div
                     className="widget-combobox-menu-header widget-combobox-item"
                     onMouseDown={PreventMenuCloseEventHandler}
+                    tabIndex={0}
                 >
                     {menuHeaderContent}
                 </div>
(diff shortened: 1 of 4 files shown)


packages/pluggableWidgets/combobox-web/src/components/ComboboxMenuWrapper.tsx
```diff
@@ -16,6 +16,7 @@ interface ComboboxMenuWrapperProps extends PropsWithChildren, Partial<UseCombobo
 }
 
 function PreventMenuCloseEventHandler(e: React.MouseEvent): void {
+    (e.target as HTMLElement)?.focus();
     e.preventDefault();
```
r0b1n, line 20: Forcefully focusing feels a bit overkill. I am not sure how the menu closing is happening and where the listener for it, but what happens if we remove `preventDefault`, but keep `stopPropogation`? In theory this still let's the browser do what it has to do, but prevents the event to going to another elements, so their listeners are not fired.

gjulivan: you are correct, we can just use stopPropagation.

but this will make clicking on non-focusable object still goes through and close the combobox dropdown.
sample case: clicking on checkbox's label instead of the input.

but this is can be handled by making the container as focusable. hence, the fix.

gjulivan:
NOTE: we can just use stopPropagation to allow child component to have focus.

but this will make clicking on non-focusable object still goes through and close the combobox dropdown.
sample case: clicking on checkbox's label instead of the input.

but this is can be handled by making the container as focusable. hence, the fix.

packages/pluggableWidgets/combobox-web/src/components/MultiSelection/SelectAllButton.tsx
```diff
@@ -23,7 +23,7 @@ export function SelectAllButton({ id, ariaLabel, value, onChange }: SelectAllBut
                 <ThreeStateCheckBox value={value} id={id} onChange={onChange} />
             </span>
             {ariaLabel ? (
-                <CaptionContent onClick={onChange} htmlFor={id}>
+                <CaptionContent onClick={() => {}} htmlFor={id}>
```
r0b1n, line 26: Let's add a comment why an empty callback is passed.

r0b1n (approved):

leonardomendix (approved):

PR #960, opened 2024-03-25
[WC-1856]: Limit items render in design mode

samuelreichert:
### Pull request type


iobuhov (dismissed):

leonardomendix (approved):

packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid.scss
```diff
@@ -374,6 +374,13 @@ $grid-selected-row-background: $brand-light;
             background-color: $grid-selected-row-background;
         }
     }
+
+    .tr.tr-preview:nth-last-child(2) > .td {
+        opacity: 0.5;
+    }
+    .tr.tr-preview:last-child > .td {
```
gjulivan, line 381: if this is only for preview mode, can we separate the file and import only on editorPreview? so that it doesn't bundle to runtime

packages/pluggableWidgets/datagrid-web/src/Datagrid.editorPreview.tsx
```diff
@@ -53,7 +53,8 @@ const initColumns: ColumnsPreviewType[] = [
 
 export function preview(props: DatagridPreviewProps): ReactElement {
     const EmptyPlaceholder = props.emptyPlaceholder.renderer;
-    const data: ObjectItem[] = Array.from({ length: props.pageSize ?? 5 }).map((_, index) => ({
+    const numberOfItems = 3;
```
gjulivan, line 56: better put constant on the top. out of the function scope

packages/pluggableWidgets/gallery-web/src/Gallery.editorPreview.tsx
```diff
@@ -47,7 +48,7 @@ function Preview(props: GalleryPreviewProps): ReactElement {
     return (
         <GalleryComponent
             className={props.class}
-            desktopItems={props.desktopItems!}
+            desktopItems={props.content.widgetCount > 0 ? numberOfItems : props.desktopItems!}
```
gjulivan, line 51: what if widgetCount is 1 or 2? shall we still display 3?

samuelreichert: Yes, widgetCount bigger than zero means there is a widget to render inside the Gallery, therefore we will display 3 items. With widgetCount being zero means that the DropZone will show, then we use the desktopitems

gjulivan (dismissed):

leonardomendix (dismissed):

gjulivan (approved):

iobuhov (approved):

PR #970, opened 2024-04-03
[WC-2409] baseUrl fix for rich text editor

gjulivan:
### Pull request type

---

### Description

adding baseUrl to let tinymce figure out correct path for skins and plugins.


packages/pluggableWidgets/rich-text-web/src/utils/contants.ts
```diff
@@ -1,3 +1,13 @@
+function getBaseUrl(): string {
+    const bases = document.getElementsByTagName("base");
```
iobuhov, line 2: NIT: you can use destructuring with nullish coalescing operator (`??`)
```suggestion
    const [base] = document.getElementsByTagName("base");
    return base?.href ?? "";
```

iobuhov (dismissed):

iobuhov (approved):

PR #977, opened 2024-04-08
[WC-2421]: DG2: Make visible property required

rahmanunver:
### Pull request type

---

### Description
Made the `visible` property for columns required since leaving it blank(`undefined`) causes a constant state of "loading", therefore keeping the DG in a constant "pending" state.

Test case from the story should be fixed, DG should be visible completely.


gjulivan:
add changelog

r0b1n (approved):

iobuhov (approved):

PR #980, opened 2024-04-08
[WC-2405]: Add barcode format controls to prevent incorrect scanning

HedwigAR:
### Pull request type

---

### Description
We found a case where a Code-39 code was recognized as a RSS-14 code instead. Since we support both of them and they do look very similar, we've added an option to be able to specify what formats should be recognized.
Now the user can select if they want to be able to recognize all formats (this is the option that's enabled by default). If the user does not want this, they'll be able to specify only the codes they're expecting.
This PR also includes a fix for scanning with the mask enabled, where it wasn't scanning the full resolution.

Testing should check whether it's still possible to recognize all barcode types when `useAllFormats` is set to Yes. Test should also check whether selecting specific formats is effective, and if this does exclude other codes from being recognized (aka, when only code-39 is selected it should no longer trigger on QR codes)


packages/pluggableWidgets/barcode-scanner-web/src/BarcodeScanner.editorConfig.ts
```diff
@@ -13,6 +13,9 @@ export function getProperties(
     if (platform === "web") {
         transformGroupsIntoTabs(defaultProperties);
     }
+    if (_.useAllFormats) {
```
iobuhov, line 16: The `_` means "unused". But now I see that we use this param, can we assign a full name here?

iobuhov (changes requested):

iobuhov:
Great PR and description BTW 👍

packages/pluggableWidgets/barcode-scanner-web/src/BarcodeScanner.editorConfig.ts
HedwigAR: Sure!

packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
```diff
@@ -83,6 +77,29 @@ export const useReader: UseReaderHook = args => {
...
+                BarcodeFormat.RSS_14,
+                BarcodeFormat.QR_CODE,
+                BarcodeFormat.DATA_MATRIX,
+                BarcodeFormat.AZTEC,
+                BarcodeFormat.PDF_417
+            ];
+        } else {
+            args.barcodeFormats.forEach((val, index) => {
+                formats[index] = BarcodeFormat[val.barcodeFormat];
+            });
```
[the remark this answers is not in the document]
HedwigAR: Ofcourse!

packages/pluggableWidgets/barcode-scanner-web/src/BarcodeScanner.editorConfig.ts
```diff
@@ -1,18 +1,21 @@
...
 export function getProperties(
-    _: BarcodeScannerContainerProps,
+    barcodeScannerContainerProps: BarcodeScannerContainerProps,
     defaultProperties: Properties,
     platform: "web" | "desktop"
 ): Properties {
     if (platform === "web") {
         transformGroupsIntoTabs(defaultProperties);
     }
+    if (barcodeScannerContainerProps.useAllFormats) {
```
iobuhov, line 16: NIT: in other `editorConfig` files we use `values` as name

packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
```diff
@@ -83,6 +77,27 @@ export const useReader: UseReaderHook = args => {
...
+                BarcodeFormat.CODE_128,
+                BarcodeFormat.ITF,
+                BarcodeFormat.RSS_14,
+                BarcodeFormat.QR_CODE,
+                BarcodeFormat.DATA_MATRIX,
+                BarcodeFormat.AZTEC,
+                BarcodeFormat.PDF_417
+            ];
+        } else {
+            args.barcodeFormats!.map((val, index) => (formats[index] = BarcodeFormat[val.barcodeFormat]));
```
iobuhov, line 98: nit:
```suggestion
            formats = args.barcodeFormats!.map(val => BarcodeFormat[val.barcodeFormat]);
```

HedwigAR: amazing

packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
iobuhov, line 1: Just one last thing: can you please fix eslint warnings in this file?

iobuhov (changes requested):

iobuhov (approved):

packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
```diff
@@ -8,63 +8,70 @@ import {
...
+        captureCanvas.width = videoCropWidth;
+        captureCanvas.height = videoCropWidth;
+
+        if (widthHeigher) {
+            videoCropWidth = videoCropWidth / aspectRatioWidth;
+        } else {
+            videoCropHeight = videoCropWidth;
+        }
+
+        const loop = (resolve: (value: Result) => void, reject: (reason?: Error) => void): void => {
```
gjulivan, line 61: function inside a function, either take it out or useCallback?

packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
```diff
@@ -8,63 +8,70 @@ import {
...
+
         const captureCanvas = reader.createCaptureCanvas(videoRef.current!);
-        captureCanvas.width = cropWidth;
-        captureCanvas.height = cropWidth;
-        const loop = (resolve: (value: Result) => void, reject: (reason?: Error) => void) => {
+        captureCanvas.width = videoCropWidth;
+        captureCanvas.height = videoCropWidth;
+
+        if (widthHeigher) {
+            videoCropWidth = videoCropWidth / aspectRatioWidth;
```
gjulivan, line 56: if the value here is re-set, isn't the captureCanvas.width and height need to be adjusted again or no?

HedwigAR: That's a good point, I thought the drawImage would do that automatically but I'm not actually sure. I'll change it

packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
HedwigAR: I took it out

packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
```diff
@@ -8,89 +8,124 @@ import {
...
+                setTimeout(() => loop(resolve, reject, captureCanvas, videoCropWidth, videoCropHeight), 50);
+            } else {
+                reject(error);
+            }
+        }
+    };
+
     const scanWithCropOnce = (reader: BrowserMultiFormatReader): Promise<Result> => {
-        const cropWidth = videoRef.current!.videoWidth * scale;
+        const element = document.getElementById("canvas-middle-middle")!;
```
iobuhov, line 79: Maybe we should use refs? It's more react way.

HedwigAR: Thanks! I fixed it and used a ref instead, let me know if I did that right

packages/pluggableWidgets/barcode-scanner-web/src/hooks/useReader.ts
```diff
@@ -8,89 +8,124 @@ import {
...
+        videoCropWidth: number,
+        videoCropHeight: number
+    ): void => {
+        try {
+            const canvasContext = captureCanvas.getContext("2d", { willReadFrequently: true });
+            if (canvasContext !== null) {
+                canvasContext.drawImage(
+                    videoRef.current!,
+                    (videoRef.current!.videoWidth - videoCropWidth) / 2,
+                    (videoRef.current!.videoHeight - videoCropHeight) / 2,
```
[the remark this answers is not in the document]
HedwigAR: You're right, I got lazy. Should be fixed now

iobuhov (approved):
LGTM

leonardomendix (approved):

iobuhov (approved):

PR #997, opened 2024-04-10
[WC-2311] do not retrieve full dropdown list when readonly

gjulivan:
### Pull request type

---

### DIFF
--- a/packages/pluggableWidgets/combobox-web/src/helpers/Association/BaseAssociationSelector.ts
+++ b/packages/pluggableWidgets/combobox-web/src/helpers/Association/BaseAssociationSelector.ts
@@ -37,6 +37,15 @@ export class BaseAssociationSelector<T extends string | string[], R extends Refe
             customContent,
             customContentType
         ] = extractAssociationProps(props);
+
+        if (attr.status === "available") {
+            if (!attr.readOnly) {
+                ds.setLimit(undefined);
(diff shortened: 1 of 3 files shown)


packages/pluggableWidgets/combobox-web/src/helpers/Association/BaseAssociationSelector.ts
```diff
@@ -37,6 +37,15 @@ export class BaseAssociationSelector<T extends string | string[], R extends Refe
             customContent,
             customContentType
         ] = extractAssociationProps(props);
+
+        if (attr.status === "available") {
+            if (!attr.readOnly) {
```
iobuhov, line 42: NIT: Merge in one `if`

gjulivan: if data is available and still readonly, 
will it go to another round of "else" clause there, and reseting limit, and trigger re-render ?

iobuhov: Yes, you right, merging is not valid here

iobuhov (approved):

leonardomendix (approved):

iobuhov (approved):

PR #998, opened 2024-04-10
[WC-2391] Add custom content events prop

iobuhov:
### Pull request type

---

### Description
Add prop to control event propagation on custom content.

### DIFF
--- a/packages/pluggableWidgets/datagrid-web/src/Datagrid.editorConfig.ts
+++ b/packages/pluggableWidgets/datagrid-web/src/Datagrid.editorConfig.ts
@@ -205,7 +205,8 @@ export const getPreview = (
                   width: "autoFit",
                   wrapText: false,
                   minWidth: "auto",
-                  minWidthLimit: 100
+                  minWidthLimit: 100,
+                  allowEventPropagation: true
               }
           ];
     const columns = rowLayout({
--- a/packages/pluggableWidgets/datagrid-web/src/Datagrid.editorConfig.ts
+++ b/packages/pluggableWidgets/datagrid-web/src/Datagrid.editorConfig.ts
@@ -32,7 +32,7 @@ export function getProperties(
             hidePropertyIn(defaultProperties, values, "columns", index, "dynamicText");
         }
         if (column.showContentAs !== "customContent") {
-            hidePropertyIn(defaultProperties, values, "columns", index, "content");
+            hideNestedPropertiesIn(defaultProperties, values, "columns", index, ["content", "allowEventPropagation"]);
         }
         if (column.showContentAs === "customContent") {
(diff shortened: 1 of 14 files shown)


packages/pluggableWidgets/datagrid-web/src/Datagrid.xml
```diff
@@ -136,6 +136,10 @@
                                     <enumerationValue key="no">No</enumerationValue>
                                 </enumerationValues>
                             </property>
+                            <property key="allowEventPropagation" type="boolean" defaultValue="true">
+                                <caption>Allow default events</caption>
+                                <description>If set to yes, then all default evens, such as "on click" or selection, will be triggered when the user interacts with custom content.</description>
+                            </property>
```
r0b1n, line 142: I think we never use word _events_, should call this actions probably. Something like "Allow row actions", or maybe better invert the meaning like "Block row actions".

iobuhov: I also was thinking firs of actions, but then I changed my mind as there also selection and IMO it's better to use more generic "events".

r0b1n: Yeah, true. Let's align with @takumagott on the names/descriptions.

packages/pluggableWidgets/datagrid-web/src/Datagrid.editorConfig.ts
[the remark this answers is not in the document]
iobuhov: Yeah, good one, I forgot about it.

packages/pluggableWidgets/datagrid-web/src/helpers/state/column/ColumnStore.tsx
```diff
@@ -262,10 +266,21 @@ const onKeyDown = (event: React.KeyboardEvent): void => {
     event.stopPropagation();
 };
 
-function CustomContent(props: { children: ReactNode }): ReactElement {
+function CustomContent({
```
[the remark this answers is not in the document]
iobuhov: Done

packages/pluggableWidgets/datagrid-web/src/Datagrid.xml
iobuhov: done

gjulivan: check again on the suggested caption.
it should be `<caption>Allow row events</caption>`

gjulivan (dismissed):

leonardomendix (dismissed):

leonardomendix (approved):

gjulivan (approved):

PR #1007, opened 2024-04-12
[WC-2442] fix sorting for custom content

r0b1n:
### Pull request type

### Description

Because of the logic of tracking underlying properties we lost track of the underlying attribute, this resulted in broken sorting for non-attribute columns even if attributes for those columns were selected.

### DIFF
--- a/packages/pluggableWidgets/datagrid-web/src/helpers/state/column/ColumnStore.tsx
+++ b/packages/pluggableWidgets/datagrid-web/src/helpers/state/column/ColumnStore.tsx
@@ -234,7 +234,9 @@ export class ColumnStore implements GridColumn {
 
     applySettings(conf: ColumnPersonalizationSettings): void {
         this.size = conf.size;
-        this.isHidden = conf.hidden;
+        if (this.canHide) {
+            this.isHidden = conf.hidden;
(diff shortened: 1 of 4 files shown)


gjulivan (approved):

packages/pluggableWidgets/datagrid-web/src/helpers/state/column/ColumnStore.tsx
```diff
@@ -83,20 +83,9 @@ export class ColumnStore implements GridColumn {
...
-            }
-            case "customContent": {
-                this._content = props.content;
-                break;
-            }
-            case "dynamicText": {
-                this._dynamicText = props.dynamicText;
-                break;
-            }
-        }
```
iobuhov, line 99: Lol, probably reactive values passed with props not subscribed with switch 🦶 🔫

r0b1n: They did, but we just had a bug in the logic. For example when `customContent` is selected we **still** need `attribute` to be able to sort 🙈

leonardomendix (approved):

PR #1023, opened 2024-04-17
Update MX version to 10.8.1 and Cypress version

leonardomendix:
### Pull request type

---

### Description

Update MX version to 10.8.1 and Cypress version.

Automation should pass.


iobuhov (dismissed):

gjulivan:
isn't 10.9 have styling problem?

leonardomendix:
> isn't 10.9 have styling problem? we can't do combo box e2e with 10.9 due to this.
> 
> https://mendix.slack.com/archives/C06RK0K0LMQ/p1712148981542289

Oh, good one! Thanks!

But for these tests, we're not using the modern client enabled. Might be an issue in the other workflow we have, but this one is not included in our PRs pipeline, so it's not a blocker. What do you think, should I change it back to 10.8? Or keep this change and wait for the fix in the next version?

gjulivan:
> > isn't 10.9 have styling problem? we can't do combo box e2e with 10.9 due to this.
> > https://mendix.slack.com/archives/C06RK0K0LMQ/p1712148981542289
> 
> Oh, good one! Thanks!
> 
> But for these tests, we're not using the modern client enabled. Might be an issue in the other workflow we have, but this one is not included in our PRs pipeline, so it's not a blocker. What do you think, should I change it back to 10.8? Or keep this change and wait for the fix in the next version?

I think we can increase 1 version first. which is going to 10.8.1

gjulivan (approved):

PR #1041, opened 2024-04-24
[WC-2441]: Fix datagrid export when exporting numbers

samuelreichert:
### Pull request type

---

### Description

When values are exported, numbers were being sent as string, this PR aims to fix that by sending the number instead of a string.

### DIFF
--- a/packages/pluggableWidgets/datagrid-web/src/features/export.ts
+++ b/packages/pluggableWidgets/datagrid-web/src/features/export.ts
@@ -231,37 +231,55 @@ type ExportDataResult =
 
 function exportData(data: ObjectItem[], columns: ColumnsType[]): ExportDataResult {
     let hasLoadingItem = false;
-    const items = data.map(item => {
-        return columns.map(column => {
+    type ExportDataColumn = Array<boolean | number | string>;
+    const items: ExportDataColumn[] = [];
+
+    for (const item of data) {
+        if (hasLoadingItem) {
+            break;
+        }
+        const cols: ExportDataColumn = [];
+
+        for (const column of columns) {
             if (column.showContentAs === "attribute") {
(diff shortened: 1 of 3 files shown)


packages/pluggableWidgets/datagrid-web/src/features/export.ts
```diff
@@ -232,11 +233,19 @@ function exportData(data: ObjectItem[], columns: ColumnsType[]): ExportDataResul
...
+
             if (column.showContentAs === "attribute") {
-                value = column.attribute?.get(item)?.displayValue ?? "";
+                const attributeItem = column.attribute?.get(item);
+
+                if (attributeItem?.formatter.type === "number") {
+                    value = Big(attributeItem.value as Big).toNumber();
+                } else {
+                    value = attributeItem?.displayValue ?? "";
+                }
```
r0b1n, line 245: We should check if we can return more data types than just number, bool, date/time might as well need a separate treatment to be able to have better properties in the exported excel.

packages/pluggableWidgets/datagrid-web/src/features/export.ts
```diff
@@ -232,11 +233,19 @@ function exportData(data: ObjectItem[], columns: ColumnsType[]): ExportDataResul
...
             } else if (column.showContentAs === "dynamicText") {
                 const dynamicText = column.dynamicText?.get(item);
+
                 if (dynamicText?.status === "loading") {
                     hasLoadingItem = true;
```
r0b1n, line 250: Maybe we should `throw` here instead and catch it later to prevent some unnecessary iteration if we clearly don't want the results anyway.

rahmanunver (approved):

r0b1n (approved):

PR #1044, opened 2024-04-25
[WC-2460] Remove useless warnings

r0b1n:
As we discovered those warnings trigger in a completely valid use case, so it is better to remove them.


packages/pluggableWidgets/datagrid-web/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@mendix/datagrid-web",
   "widgetName": "Datagrid",
-  "version": "2.17.0",
+  "version": "2.17.1",
```
iobuhov, line 4: Ah, I just merged my PR. now this will be part of 2.18

iobuhov (approved):

PR #1053, opened 2024-05-01
feat : added dynamic page size

sharadsums:
### **Pull request type**
 
 
Added feature to dynamically add page size.
 
### **Description**
This is to facilitate users when they want to change the no of records displayed per page from UI.
 
### **What should be covered while testing?**


gjulivan:
Hi, Thank you for your PR,
we will check with the product side first if this is the correct way to go.

iobuhov:
If this feature will be approved by our PO we need some time to move all the paging logic to hook. The way it's written right now is out of optimal and will result in huge tech debt. So, some refactoring will be required.

PR #1066, opened 2024-05-08
[WC-2366] Migrate e2e to playwright

leonardomendix:
### Pull request type

---

### Description

This PR aims to change our e2e tooling from Cypress to Playwright, focusing on stability, velocity, and collaboration. All the references from Cypress were removed and the pipeline execution started to use Playwright, the same way as locally is using Playwright UI for easier debugging.

Major changes to note:
- Moved to Playwright in CI execution and locally;
- Removed usage of a docker when using the Playwright test;
- Refactored Firefox and Modern Client workflows to use Playwright as well;
- When running the `e2edev` script the screenshots are ignored because of the dependency of the OS;

To Do:

- Find way to easily update the snapshots locally;
- Re-enable a11y tests.

Automation should work as before.


automation/run-e2e/package.json
```diff
@@ -1,9 +1,9 @@
 {
   "name": "@mendix/run-e2e",
-  "version": "0.2.0",
+  "version": "0.3.0",
```
iobuhov, line 3: 👍

iobuhov (commented):
Man, you really wanted PW

iobuhov (approved):

gjulivan (approved):

PR #1075, opened 2024-05-14
[WC-1619]: Combobox lazy load option

samuelreichert:
### Pull request type

---

### Description

Add new property `lazyLoading` for association and database source, with default value set to false.
When set to true, only a set of data is loaded in the beginning, and the rest will be loaded when scrolling to the bottom.

There are no visual changes, all visual changes will be implemented in another PR.

### DIFF
--- a/packages/pluggableWidgets/combobox-web/src/helpers/Association/BaseAssociationSelector.ts
+++ b/packages/pluggableWidgets/combobox-web/src/helpers/Association/BaseAssociationSelector.ts
@@ -1,4 +1,4 @@
-import { ObjectItem, ReferenceValue, ReferenceSetValue, ActionValue } from "mendix";
+import { ObjectItem, ReferenceValue, ReferenceSetValue, ActionValue, ValueStatus } from "mendix";
 import { ComboboxContainerProps, OptionsSourceAssociationCustomContentTypeEnum } from "../../../typings/ComboboxProps";
 import { Status } from "../types";
 import { AssociationOptionsProvider } from "./AssociationOptionsProvider";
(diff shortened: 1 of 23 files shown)


packages/pluggableWidgets/combobox-web/src/helpers/Association/BaseAssociationSelector.ts
```diff
@@ -35,11 +37,19 @@ export class BaseAssociationSelector<T extends string | string[], R extends Refe
             filterType,
             onChangeEvent,
             customContent,
-            customContentType
+            customContentType,
+            lazyLoading
         ] = extractAssociationProps(props);
 
         if (attr.status === "available") {
```
gjulivan, line 44: if combobox is readonly, should always set limit = 0. (do not retrieve value at all)
this logic will breaks it.

packages/pluggableWidgets/combobox-web/src/helpers/Database/DatabaseSingleSelector.ts
```diff
@@ -42,11 +44,19 @@ export class DatabaseSingleSelector<T extends string | Big, R extends EditableVa
...
             customContentType,
             valueAttribute,
-            emptyValue
+            emptyValue,
+            lazyLoading
         ] = extractDatabaseProps(props);
 
         if (attr.status === "available") {
-            if (!attr.readOnly) {
+            if (lazyLoading) {
```
gjulivan, line 52: same here

gjulivan (approved):

leonardomendix (approved):

gjulivan (approved):

PR #1090, opened 2024-05-21
[WC-2480] add missing events from RTE v2

gjulivan:
### Pull request type

---

### Description

return back onChange and onKeyPress event that is available on v2.

for consistency, add onChangeType : onLeave or onDataChange

the difference between onchange with onleave changetype is that there is a validation for value change, which makes event triggers only if value change.
as for current on-blur event, it will always trigger regardless of the data.

### DIFF
--- a/packages/pluggableWidgets/rich-text-web/src/RichText.tsx
+++ b/packages/pluggableWidgets/rich-text-web/src/RichText.tsx
@@ -51,8 +51,8 @@ export default function RichText(props: RichTextContainerProps): JSX.Element {
                 style={wrapperStyle}
                 {...wrapperAttributes}
             >
-                {stringAttribute.status === "loading" ? (
-                    <div></div>
+                {stringAttribute.status === "loading" || stringAttribute.status !== "available" ? (
+                    <div className="mx-progress"></div>
                 ) : (
                     <BundledEditor
                         {...props}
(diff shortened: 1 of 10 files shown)


packages/pluggableWidgets/rich-text-web/src/RichText.tsx
```diff
@@ -51,8 +51,8 @@ export default function RichText(props: RichTextContainerProps): JSX.Element {
                 style={wrapperStyle}
                 {...wrapperAttributes}
             >
-                {stringAttribute.status === "loading" ? (
-                    <div></div>
+                {stringAttribute.status === "loading" || stringAttribute.status !== "available" ? (
```
samuelreichert, line 54:
```suggestion
                {stringAttribute.status !== "available" ? (
```

samuelreichert: It will already cover the loading status

gjulivan: this is to prevent undefined. seems like some async process sometimes make incorrect rendering here.
blindly trying to fix WC-2475

samuelreichert (approved):

leonardomendix (approved):

PR #1101, opened 2024-06-04
[WC-2054]: Add combobox datasource filtering

samuelreichert:
### Pull request type

---

### Description

Implement filtering (search) using datasource `setFilter` instead of the matchSorter, when using lazyLoading.


packages/pluggableWidgets/combobox-web/src/helpers/Association/AssociationOptionsProvider.ts
gjulivan, line 1: I notice that now this `AssociationOptionsProvider` is almost very similar to `DatabaseOptionsProvider`.
Thus, instead of extend from `BaseOptionsProvider`,

can we make both extend from something new like `BaseDatasourceOptionsProvider` which then extend from `BaseOptionsProvider` ?

gjulivan: image

packages/pluggableWidgets/combobox-web/src/helpers/BaseDatasourceOptionsProvider.ts
```diff
@@ -0,0 +1,86 @@
...
+import { BaseOptionsProvider } from "./BaseOptionsProvider";
+import { datasourceFilter } from "./datasourceFilter";
+import { CaptionsProvider, Status, SortOrder } from "./types";
+import { DEFAULT_LIMIT_SIZE } from "./utils";
+
+export interface BaseProps {
+    ds: ListValue;
+    filterType: FilterTypeEnum;
+    lazyLoading: boolean;
+    filterId?: ListAttributeValue["id"];
```
iobuhov, line 12: nit: I would advice to not change the semantics here. Instead I would align name with type `attributeId`

iobuhov (approved):
LGTM 🔥

gjulivan:
increase minor version to prepare for release after this.

packages/pluggableWidgets/combobox-web/src/components/Loader.tsx
```diff
@@ -25,6 +26,10 @@ export function Loader(props: LoaderProps): ReactElement | null {
                 <SkeletonLoader withCheckbox={withCheckbox} key={i} />
             ))}
         </Fragment>
+    ) : isEmpty ? (
+        <div style={{ margin: "52px 0" }}>
```
gjulivan, line 30: let's stick to use classname here, to allow users to override easily

gjulivan (approved):

leonardomendix (approved):

gjulivan (approved):

iobuhov (approved):

PR #1109, opened 2024-06-10
[WC-2483] :use css variables for data-widgets modules

rahmanunver:
### Pull request type

---

### Description
Using css variables instead of scss vars for values already set from atlas.


packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid-filters.scss
```diff
@@ -2,6 +2,7 @@ $hover-color: #f8f8f8 !default;
 $background-color: #fff !default;
```
gjulivan, line 2: remove !default flag,
this no longer takes precedence over atlas-core values.

gjulivan (dismissed):

gjulivan (approved):

leonardomendix (approved):

gjulivan (approved):

PR #1117, opened 2024-06-17
[WC-2518]: Fix slider-web tooltip positioning

rahmanunver:
### Pull request type

---

### Description
Same fix applied in #1118 for `range-slider-web`
Added `getTooltipContainer` method from rc-slider to keep the tooltip in its containing element instead of rendering up in the body.

The tooltip should be following the slider.

### DIFF
--- a/packages/pluggableWidgets/slider-web/src/utils/createHandleGenerator.tsx
+++ b/packages/pluggableWidgets/slider-web/src/utils/createHandleGenerator.tsx
@@ -34,6 +34,7 @@ export function createHandleGenerator(props: CreateHandleGeneratorParams): Handl
 
         return (
             <Tooltip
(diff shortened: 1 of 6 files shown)


packages/pluggableWidgets/slider-web/src/utils/createHandleGenerator.tsx
gjulivan, line 1: this will not work if there are multiple slider in the page, right?

gjulivan: try using ref.. might work with `ref.current.querySelector`
this way you scope the selector based on the container.

r0b1n (approved):

leonardomendix (approved):

PR #1118, opened 2024-06-17
[WC-2518]: Fix range-slider-web tooltip positioning

rahmanunver:
### Pull request type

---

### Description
Linked with #1117 

Same as #1117


packages/pluggableWidgets/range-slider-web/src/utils/createHandleGenerator.tsx
gjulivan, line 1: same comment, this will not work if there are multiple slider on the page.
try to create the DOM scope for the querySelector.

leonardomendix (approved):

PR #1125, opened 2024-06-21
feat(Datagrid): made pagination visible in virtual scrolling & load more

sharadsums:
Pull request type
Added feature to have pagination also visible on virtual scrolling & loadmore.

Description
This is to facilitate users when they want to see paging count in all pagination option
it will be in format **<< **1** to **page  item count** of **total count**>>** for virtual scrolling and load more

What should be covered while testing?


sharadsums:
Hi @iobuhov 
Yes sure will do

PR #1128, opened 2024-06-24
[WC-2535]: Fix barcodescanner error on module load

HedwigAR:
### Pull request type

---

### Description

Currently an issue exists where the module won't load because the reference to the canvas it needs is undefined. This PR changes this reference from a state to a ref, which ensures the reference can be found in the async process.

If the camera view appears with crop enabled, it means the canvas reference is not undefined, so the issue is fixed. While this shouldn't impact other features in the widget, it would be good to double-check they all work so we won't have to unpublish the widget again.

### DIFF
--- a/packages/pluggableWidgets/barcode-scanner-web/src/components/BarcodeScanner.tsx
+++ b/packages/pluggableWidgets/barcode-scanner-web/src/components/BarcodeScanner.tsx
@@ -1,4 +1,4 @@
-import { createElement, ReactElement, ReactNode, useCallback, SyntheticEvent, useState } from "react";
+import { createElement, ReactElement, ReactNode, useCallback, SyntheticEvent, useRef } from "react";
 import classNames from "classnames";
 import { Alert } from "@mendix/widget-plugin-component-kit/Alert";
 import { Dimensions, getDimensions } from "@mendix/widget-plugin-platform/utils/get-dimensions";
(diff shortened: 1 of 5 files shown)


gjulivan (dismissed):

packages/pluggableWidgets/barcode-scanner-web/src/components/BarcodeScanner.tsx
```diff
@@ -112,7 +112,7 @@ export function BarcodeScanner({
             showMask={showMask}
             canvasMiddleMiddleRef={ref => {
                 if (ref !== null) {
-                    setCanvasMiddleRef(ref);
+                    canvasMiddleRef.current = ref;
                 }
             }}
```
iobuhov, line 117: I would suggest instead, change prop type on line 15, to React.RefObject and pass it as it is without using callback. Callback as a ref it's legacy react approach and in this case it not add any value. Just use ref as it is.

iobuhov (changes requested):

packages/pluggableWidgets/barcode-scanner-web/src/components/BarcodeScanner.tsx
HedwigAR: Alright, thanks for the suggestion, I thought doing that made it too cluttered, and just having the object there would be nicer. I'll change it.

HedwigAR: I changed it, let me know if this is right

iobuhov (dismissed):
LGTM

iobuhov (approved):
LGTM

leonardomendix (approved):

PR #1138, opened 2024-06-28
[WC-2511] accordion unsync content display with state

gjulivan:
### Pull request type

Bug Fix
---

### Description

child content tries to expand, and calculate the content height while parent is not finished expanding. 
thus, child content is still not fully visible, which cause their bounding rectangle calculation incorrect.

fixing this by adding resize observer to the content.

Anything related to accordion content's resizing. make sure the resize observer doesn't try to expand content while content should be collapsing

### DIFF
--- a/packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
+++ b/packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
@@ -48,7 +48,7 @@ export function AccordionGroup(props: AccordionGroupProps): ReactElement | null 
     useEffect(() => {
         const resizeObserver = new ResizeObserver(entries => {
             for (const entry of entries) {
-                if (entry.contentBoxSize && !props.collapsed && contentWrapperRef.current && contentRef.current) {
+                if (entry.contentBoxSize && !renderCollapsed && contentWrapperRef.current && contentRef.current) {
                     contentWrapperRef.current.style.height = `${contentRef.current.getBoundingClientRect().height}px`;
                 }
             }
@@ -59,7 +59,7 @@ export function AccordionGroup(props: AccordionGroupProps): ReactElement | null 
         return () => {
             resizeObserver.disconnect();
         };
-    }, [props.collapsed]);
+    }, [renderCollapsed]);
 
     const completeTransitioning = useCallback((): void => {
         if (contentWrapperRef.current && rootRef.current && animatingContent.current) {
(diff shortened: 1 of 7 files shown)


packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
```diff
@@ -45,6 +45,28 @@ export function AccordionGroup(props: AccordionGroupProps): ReactElement | null
...
+            for (const entry of entries) {
+                if (entry.contentBoxSize) {
+                    if (!props.collapsed) {
+                        if (contentWrapperRef.current && contentRef.current) {
+                            contentWrapperRef.current.style.height = `${
+                                contentRef.current.getBoundingClientRect().height
+                            }px`;
+                        }
+                    }
+                }
```
iobuhov, line 59: Can we reduce amount of ifs with && operator?

iobuhov: Also, are you sure that setting height on the wrapper is a good idea? I'm not very convinced that this is enough to fix this issue.

gjulivan: setting height is needed for the transition.
transition from-and-to need exact height calculation.
thus, this is needed. 
probably need to add checking for : if animation is true here.

packages/pluggableWidgets/accordion-web/src/components/AccordionGroup.tsx
```diff
@@ -45,6 +45,22 @@ export function AccordionGroup(props: AccordionGroupProps): ReactElement | null
...
+            }
+        });
+        if (contentRef.current) {
+            resizeObserver.observe(contentRef.current);
+        }
+        return () => {
+            resizeObserver.disconnect();
+        };
+    }, [props.collapsed]);
```
r0b1n, line 62: Do I understand correctly that this prop is static and coming from SP? I wonder why we depend on it and not on the current status of the state, can you clarify?

gjulivan: there's no actual difference in this case.

but I think you are correct. localized state is a better use for this.
updated

r0b1n (approved):

r0b1n (dismissed):

leonardomendix (approved):

r0b1n (approved):

PR #1147, opened 2024-07-08
[WC-2478]: Filters next

iobuhov:
### Pull request type

---


packages/shared/widget-plugin-filter-selector/src/FilterSelector.tsx
```diff
@@ -22,17 +20,15 @@ export function FilterSelector<T extends string>(props: FilterSelectorProps<T>):
...
         },
         [onChange]
     );
 
     useEffect(() => {
-        setValue(defaultFilter);
-        onChange(defaultFilter, false);
-    }, [defaultFilter, onChange]);
+        onChange(value, false);
+    }, [value, onChange]);
```
r0b1n, line 31: Since we switched to "controlled" component I wonder if we even need this `useEffect` (and the whole `isFromUserInteraction` flag).

packages/shared/widget-plugin-filtering/src/stores/DateInputFilterStore.ts
```diff
@@ -77,13 +77,8 @@ export class DateInputFilterStore
...
             this._attributes = attributes;
             const formatter = attributes.at(0)?.formatter;
-            // Just pleasing TypeScript.
-            if (formatter?.type !== "datetime") {
-                console.error("InputFilterStore: encounter invalid attribute type while updating props.");
-                return;
-            }
-            this.arg1.updateProps(formatter);
-            this.arg2.updateProps(formatter);
+            this.arg1.updateProps(formatter as DateTimeFormatter);
```
samuelreichert, line 80: @iobuhov 🚀

leonardomendix (approved):

PR #1162, opened 2024-07-23
Tooltip: switch to Floating UI

rahmanunver:
### Pull request type

---

### Description
Switching Tooltip from popper to floating-ui due to positioning and overflow detection issues, the work is in progress.


packages/pluggableWidgets/tooltip-web/src/utils/useFloatingUI.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export interface FloatingProps {
+    position: Placement;
+    showTooltip: boolean;
+    setShowTooltip: React.Dispatch<React.SetStateAction<boolean>>;
+    arrowElement: HTMLDivElement | null;
+    openOn: OpenOnEnum;
+}
+
+export function useFloatingUI(props: FloatingProps) {
```
gjulivan, line 27: missing return type

packages/pluggableWidgets/tooltip-web/src/components/Tooltip.tsx
```diff
@@ -22,96 +20,45 @@ export interface TooltipProps {
...
             </div>
         ) : null;
     };
     return (
-        <div
-            className={classNames(props.class, "widget-tooltip", `widget-tooltip-${position}`)}
-            ref={componentReference}
-        >
+        <div className={classNames(props.class, "widget-tooltip", `widget-tooltip-${position}`)}>
             {renderTrigger()}
```
gjulivan, line 62: you can move renderTrigger and renderTooltip directly inside jsx,
unless it is a separate component, then make it as separate react component.
keep the render view code clean.

packages/pluggableWidgets/tooltip-web/src/ui/Tooltip.scss
```diff
@@ -40,22 +14,48 @@
         word-break: break-word;
     }
 
-    .widget-tooltip-arrow,
-    .widget-tooltip-arrow::before {
+    [class^="widget-tooltip-arrow"],
```
gjulivan, line 17: isn't it better to use "contains" rather than starts with?
what if the class sequence change in the future?

gjulivan:
add changelog and increase version

gjulivan (dismissed):

leonardomendix (approved):

r0b1n (approved):

PR #1190, opened 2024-08-21
[WC-2542]: Fix issue with barcodescanner when changing between portrait and landscape mode

HedwigAR:
### Pull request type

---

### Description
This PR fixes an issue in the barcode scanner. If you have the crop enabled, and change between portrait and landscape while scanning, it stops being able to scan. This PR also refactors the cropping and manual scanning code, and moves it to a separate file.

During testing, the cropping should be started in portrait, then switched to landscape and vice versa. After this, scanning should still work as expected
Normal scanning should also not be effected by the changes.

### DIFF
--- a/packages/pluggableWidgets/barcode-scanner-web/src/components/utils.tsx
+++ b/packages/pluggableWidgets/barcode-scanner-web/src/components/utils.tsx
@@ -4,7 +4,8 @@ import {
     HTMLCanvasElementLuminanceSource,
     HybridBinarizer,
     BarcodeFormat,
-    DecodeHintType
+    DecodeHintType,
+    Result
 } from "@zxing/library";
 import { BarcodeFormatsType } from "typings/BarcodeScannerProps";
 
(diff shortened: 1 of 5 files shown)


packages/pluggableWidgets/barcode-scanner-web/src/components/utils.tsx
```diff
@@ -0,0 +1,93 @@
...
+            0,
+            0,
+            videoWidthHeight.width,
+            videoWidthHeight.height
+        );
+    }
+    return captureCanvas;
+};
+
+export const decodeCanvas = (reader: BrowserMultiFormatReader, captureCanvas: HTMLCanvasElement) => {
```
gjulivan, line 59: let's not skipped the typescript here and add a return type

packages/pluggableWidgets/barcode-scanner-web/src/components/utils.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const luminanceSource = new HTMLCanvasElementLuminanceSource(captureCanvas);
+
+    const binaryBitmap = new BinaryBitmap(new HybridBinarizer(luminanceSource));
+    const result = reader.decodeBitmap(binaryBitmap);
+    return result;
+};
+
+export const createHints = (useAllFormats: boolean, barcodeFormats?: BarcodeFormatsType[]) => {
```
gjulivan, line 67: let's not skipped the typescript here and add a return type

gjulivan (dismissed):

leonardomendix (dismissed):

gjulivan (dismissed):

leonardomendix (approved):

gjulivan (approved):

PR #1215, opened 2024-09-16
[WC-2553] DG2: Add reference set support.

rahmanunver:
### Pull request type

---

### Description
Added new `Reference` and `Reference Set` association types in the DG2 linked attribute.

Should be able to show multiple associated values separated by comma.

Updated PWT for the new client type, so run `pnpm install` to make sure it's updated.

Need a domain model with an association of Reference Set to be able to select them in DG2.

### DIFF
--- a/packages/pluggableWidgets/datagrid-web/package.json
+++ b/packages/pluggableWidgets/datagrid-web/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@mendix/datagrid-web",
   "widgetName": "Datagrid",
-  "version": "2.21.2",
+  "version": "2.21.3",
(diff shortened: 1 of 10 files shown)


packages/pluggableWidgets/datagrid-web/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@mendix/datagrid-web",
   "widgetName": "Datagrid",
-  "version": "2.21.2",
+  "version": "2.21.3",
```
gjulivan, line 4: added feature, minor version should increase.. thus v2.22.0

PR #1230, opened 2024-09-20
[WC-2609] FileUploader: Phase 1 (MVP)

r0b1n:


packages/pluggableWidgets/fileuploader-web/src/stores/FileStore.ts
```diff
@@ -0,0 +1,185 @@
...
+                });
+            }
+        } catch (e: unknown) {
+            runInAction(() => {
+                this.fileStatus = "uploadingError";
+            });
+        }
+    }
+
+    renderTitle(): ReactNode {
```
iobuhov, line 109: I would change this to getter and use "string" as return type to avoid "render" in name.
This is jsut properties ("title", "size") and I think it's better to not mix model with view.
Let's use components for rendering.

packages/pluggableWidgets/fileuploader-web/src/utils/allowedFormatChecker.ts
```diff
@@ -0,0 +1,42 @@
+import { AllowedFileFormatsPreviewType } from "../../typings/FileUploaderProps";
```
iobuhov, line 1: The file name is a bit confusing. I see that this is "format" validation. But "allowedFormatCheckers" present only in file name.

packages/pluggableWidgets/fileuploader-web/src/stores/FileStore.ts
```diff
@@ -0,0 +1,185 @@
...
+            callback: () => {
+                runInAction(() => {
+                    this.fileStatus = "removedFile";
+                    this.mxObject = undefined;
+                });
+            },
+            error(e: unknown) {
+                console.error("Could not remove object:", e);
+            }
+        });
```
iobuhov, line 154: I see that in other places we use utils to "hide" calls to mx.data,
but here we calling this API directly.
Should we create "remove" funciton in utils as well?

packages/pluggableWidgets/fileuploader-web/src/stores/FileStore.ts
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    async fetchMxObject(): Promise<void> {
+        if (this.objectItem) {
+            const obj = await fetchMxObject(this.objectItem!);
+            runInAction(() => {
+                this.mxObject = obj;
+            });
```
iobuhov, line 162: Why we wrapping assignement in runInAction?
It looks redundant.

packages/pluggableWidgets/fileuploader-web/src/stores/FileStore.ts
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    static newFileWithError(file: File, errorMessage: string, rootStore: FileUploaderStore): FileStore {
+        const store = new FileStore("validationError", rootStore, file, undefined);
+        runInAction(() => {
+            store.errorDescription = errorMessage;
+        });
+
+        return store;
+    }
```
iobuhov, line 184: More of a question - we create store on evry ds update? Why?

packages/pluggableWidgets/fileuploader-web/src/utils/allowedFormatChecker.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (/^[^/]+\/[^/]+$/.test(c)) {
+        // "type/subtype" string
+        const [type, subtype] = c.split("/");
+        return [type, subtype];
+    }
+
+    throw new Error(`Value '${c}' is not recognized. Accepted format: 'image/jpeg'`);
+}
+
+function parseExtensionsList(config: string): string[] {
```
iobuhov, line 27: Can we add few tests just to show what strings are valid input?

packages/pluggableWidgets/fileuploader-web/src/stores/FileUploaderStore.ts
```diff
@@ -0,0 +1,153 @@
...
+            processDrop: action,
+            files: observable,
+            existingItemsLoaded: observable,
+            errorMessage: observable
+        });
+
+        this.updateProps(props);
+    }
+
+    updateProps(props: FileUploaderContainerProps): void {
```
iobuhov, line 44: Can we have a comment explaining what happens here?

packages/pluggableWidgets/fileuploader-web/src/stores/FileUploaderStore.ts
```diff
@@ -0,0 +1,153 @@
...
+
+            this.currentWaiting.push(resolve);
+        });
+    }
+
+    async processDrop(acceptedFiles: File[], fileRejections: FileRejection[]): Promise<void> {
+        if (!this._createFileAction || !this._createFileAction.canExecute) {
+            console.warn(
+                `'Action to create new files' is not available or can't be executed. Please check if '${this._widgetName}' widget is configured correctly.`
+            );
```
iobuhov, line 102: Should it be en error as well? I think Aciton is required and we should "fail fast" if it's not present.
Alos, I think this should be part of consistency check.

packages/pluggableWidgets/fileuploader-web/src/stores/FileUploaderStore.ts
```diff
@@ -0,0 +1,153 @@
...
+            runInAction(() => {
+                this.errorMessage = "Can't upload files at this time. Please contact you system administrator.";
+            });
+            return;
+        }
+
+        if (fileRejections.length && fileRejections[0].errors[0].code === "too-many-files") {
+            runInAction(() => {
```
iobuhov, line 110: Maybe it make sens to create "setMessage" action to avoid
wrapping each assignement into "runInAction"?

packages/pluggableWidgets/fileuploader-web/src/stores/FileUploaderStore.ts
iobuhov: Alos, maybe it make sense to break this method into smaller as seems that
it's doing too much.

packages/pluggableWidgets/fileuploader-web/src/stores/FileStore.ts
r0b1n: ✅

packages/pluggableWidgets/fileuploader-web/src/utils/allowedFormatChecker.ts
r0b1n: ✅

packages/pluggableWidgets/fileuploader-web/src/stores/FileUploaderStore.ts
r0b1n: ✅

packages/pluggableWidgets/fileuploader-web/src/stores/FileStore.ts
r0b1n: ✅

packages/pluggableWidgets/fileuploader-web/src/stores/FileStore.ts
r0b1n: It is not on every update we create it, only when we see new external non-empty items/files. In `updateProps` we listen on the new objects that might appear in the data source, if the object is not empty we assume it is an externally added file, so we create a store around it. But we don't re-create stores for items we already saw in the data grid.

packages/pluggableWidgets/fileuploader-web/src/components/FileEntry.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { UploadInfo } from "./UploadInfo";
+import { createElement, ReactElement } from "react";
+import { FileStore } from "../stores/FileStore";
+import { observer } from "mobx-react-lite";
+
+interface FileEntryProps {
+    store: FileStore;
+}
+
+export const FileEntry = observer(({ store }: FileEntryProps): ReactElement => {
```
iobuhov, line 12: I know it's a bit controversial, but It's better if we have pure component that just renders props.
Yeah, I know, in dg, we passing store directly, but thats for a reason.
Here I would create chain like this:
FileEntryContainer->FilteEntry
FileContainer - knows about store and map store props to component props
FileEntry -> dumb component, just renders props.

iobuhov: This way, only container should be wrapped with observer, and FileEntry will remain pure react component.

packages/pluggableWidgets/fileuploader-web/src/stores/FileUploaderStore.ts
r0b1n: I simplified it, should be a bit better now.

packages/pluggableWidgets/fileuploader-web/src/stores/FileStore.ts
r0b1n: The fetching of the object is async, so it have to be wrapped, even though the method itself is marked as action. mobx can't track async operations as part of one action.

packages/pluggableWidgets/fileuploader-web/src/components/FileEntry.tsx
r0b1n: ✅

packages/pluggableWidgets/fileuploader-web/src/utils/allowedFormatChecker.ts
r0b1n: ✅

packages/pluggableWidgets/fileuploader-web/src/stores/FileUploaderStore.ts
r0b1n: I added this as part of consistency errors, good point! Let's keep this one for the case where it is not available for some reason, as this situation can happen dynamically.

gjulivan (approved):

leonardomendix (approved):

PR #1242, opened 2024-09-26
[WC-1840]: Store GUID in JSON

iobuhov:
### Pull request type

---

### Description
Extend Data grid API to store GUID values for drop down filters.


packages/pluggableWidgets/datagrid-web/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Added
+
+-   We have extended the Filters API to support drop-down values. If "Use lazy" is false, the drop-down filter can restore it's value from the datagrid config.
```
r0b1n, line 11: I think we should not mention Filters API, those are not for the users. We should say that we removed (a part) of the limitation [mentioned in the docs](https://docs.mendix.com/appstore/modules/data-grid-2/#drop-down-filter). It is now possible to restore values when navigating back and from the personalization.

r0b1n (approved):

packages/pluggableWidgets/datagrid-web/CHANGELOG.md
iobuhov: Done

gjulivan (approved):

PR #1250, opened 2024-10-03
chore(data-widget): fix build number

gjulivan:
### Pull request type

---

### DIFF
--- a/packages/modules/data-widgets/package.json
+++ b/packages/modules/data-widgets/package.json
@@ -27,13 +27,13 @@
   },
   "moduleFolderNameInModeler": "datawidgets",
   "marketplace": {
-    "minimumMXVersion": "10.12.0",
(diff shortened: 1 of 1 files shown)


iobuhov (dismissed):

packages/modules/data-widgets/package.json
```diff
@@ -27,13 +27,13 @@
...
   "marketplace": {
-    "minimumMXVersion": "10.12.0",
+    "minimumMXVersion": "10.12.6.46815",
     "appNumber": 116540,
     "appName": "Data Widgets"
   },
   "testProject": {
     "githubUrl": "https://github.com/mendix/DataWidgets-module",
-    "branchName": "main"
+    "branchName": "10-12-6-test"
```
iobuhov, line 36: can we remove test from branch name? And I think we can use `10.12` as branch name

PR #1255, opened 2024-10-08
[WC-2659] Data widgets Mx10 release fix

leonardomendix:
### Pull request type

---

### Description

This PR is adding a new condition to check if the module mx minimum version is 10, if so then we need to use a different mxbuild build, since mxutil was discontinued and all the tooling now is part of mx executable.


gjulivan (approved):

packages/modules/data-widgets/package.json
```diff
@@ -27,7 +27,7 @@
   },
   "moduleFolderNameInModeler": "datawidgets",
   "marketplace": {
-    "minimumMXVersion": "10.12.0",
+    "minimumMXVersion": "10.12.0.38909",
```
gjulivan, line 30: min SP to 10.12.6.46815

because there was a problem with datagrid 2 reference set in 10.5

leonardomendix: ✅  Done!

gjulivan (approved):

automation/utils/src/mpk.ts
```diff
@@ -10,11 +10,17 @@ async function ensureMxBuildDockerImageExists(mendixVersion: Version): Promise<v
...
+        const dockerfilePathMx10 = join(__dirname, "../docker/mxbuildMx10.Dockerfile");
+        const versionMajor = mendixVersion.major;
+        const command =
+            versionMajor < 10
+                ? `docker build -f ${dockerfilePath} ` +
+                  `--build-arg MENDIX_VERSION=${version} ` +
+                  `-t mxbuild:${version} ${process.cwd()}`
+                : `docker build -f ${dockerfilePathMx10} ` +
+                  `--build-arg MENDIX_VERSION=${version} ` +
+                  `-t mxbuild:${version} ${process.cwd()}`;
```
iobuhov, line 22: Not a big deal, but I think we can use same command, and only change `dockerfilePath` depending on major version

leonardomendix: Thanks! Done! ✅

gjulivan (approved):

iobuhov (approved):

PR #1260, opened 2024-10-10
[WC-2661] fix charts not resized correctly in a popup page

gjulivan:
### Pull request type

---

### Description

plotly only recognize window resize event.
by having the charts inside a modal dialog, the resize event is not triggered.
this is proven by trying: 
resize modal dialog, then try resize window.. 
the charts will be rendered as expected, because window resize is triggered afterwards.

fixed by trying to add resize observer to the parent's container of the charts widget.
and trigger window resize event for plotly to re-render the correct size.


packages/shared/charts/src/components/ChartWidget.tsx
```diff
@@ -60,6 +62,16 @@ export const ChartWidget = ({
...
+    const [callResizeObserverDebounce] = useMemo(
+        () =>
+            debounce((_target: HTMLDivElement) => {
+                window.dispatchEvent(new Event("resize"));
+            }, 100),
+        []
+    );
+
+    const ref = useResizeObserver(callResizeObserverDebounce);
+
```
iobuhov, line 74: Looks good, but I think it probably make sense to wrap this `useMemo` and `useResizeObserver` with `useDispatchResizeEffect` hook to make code more readable and couple logical pieces together.

iobuhov: Also, I would probably add `useEffect` with `abort` function created by `debounce` to avoid memory leaks.

iobuhov (changes requested):

iobuhov (approved):
LGTM

leonardomendix (approved):

PR #1261, opened 2024-10-10
[WC-2647]: Fix AutoNumber formatter

rahmanunver:
### Pull request type

---

### Description
Fixing filters not working for AutoNumber, WIP for falsy values and tests for overriden formatter.

### DIFF
--- a/packages/shared/widget-plugin-filtering/src/stores/Argument.ts
+++ b/packages/shared/widget-plugin-filtering/src/stores/Argument.ts
@@ -44,18 +44,8 @@ export class ArgumentBase<T extends Big | Date | string> {
 
     get displayValue(): string {
         if (this.isValid) {
-            console.log(
-                "displayValue",
-                this._formatter?.format(this.value),
-                this.value?.toString(),
-                "outcome",
-                this._formatter?.format(this.value) || this.value?.toString() || ""
-            );
-
             return this._formatter?.format(this.value) || this.value?.toString() || "";
         } else {
-            console.log("displayValue invalid", this._lastSetTextValue);
-
             return this._lastSetTextValue;
         }
     }
@@ -63,7 +53,6 @@ export class ArgumentBase<T extends Big | Date | string> {
     set displayValue(textValue: string) {
         this._lastSetTextValue = textValue;
         const parseResult = this._formatter?.parse(this._lastSetTextValue);
-        console.log("parseResult", parseResult);
 
         this.isValid = parseResult.valid || (parseResult as any).type === "AutoNumber";
         if ((parseResult as any).valid || (this.isValid && !this.equals(this._value, (parseResult as any).value))) {
--- a/packages/shared/widget-plugin-filtering/src/stores/Argument.ts
+++ b/packages/shared/widget-plugin-filtering/src/stores/Argument.ts
@@ -44,8 +44,18 @@ export class ArgumentBase<T extends Big | Date | string> {
 
     get displayValue(): string {
         if (this.isValid) {
+            console.log(
+                "displayValue",
+                this._formatter?.format(this.value),
+                this.value?.toString(),
+                "outcome",
+                this._formatter?.format(this.value) || this.value?.toString() || ""
+            );
+
             return this._formatter?.format(this.value) || this.value?.toString() || "";
         } else {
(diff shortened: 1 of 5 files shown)


packages/shared/widget-plugin-filtering/src/stores/NumberInputFilterStore.ts
```diff
@@ -1,32 +1,36 @@
...
+        let { formatter } = attributes[0];
+        formatter = formatterFix(formatter);
         super(new NumberArgument(formatter), new NumberArgument(formatter), "equal", attributes);
         makeObservable(this, {
             updateProps: action,
             fromJSON: action,
             fromViewState: action
         });
-
+        this.formatter = formatter;
```
iobuhov, line 33: iI think you should not set formatter on FilterStore. It should be used by NumberArgument internally.

packages/shared/widget-plugin-filtering/src/stores/NumberInputFilterStore.ts
```diff
@@ -36,9 +40,8 @@ export class NumberInputFilterStore
         if (!comparer.shallow(this._attributes, attributes)) {
             this._attributes = attributes;
         }
-        const formatter = attributes.at(0)?.formatter;
-        this.arg1.updateProps(formatter as NumberFormatter);
-        this.arg2.updateProps(formatter as NumberFormatter);
+        this.arg1.updateProps(this.formatter as NumberFormatter);
+        this.arg2.updateProps(this.formatter as NumberFormatter);
```
iobuhov, line 44: I feel like we can ignore updating formatter. So you can just remove the whole thing.

iobuhov: And actually it's a good question that you can ask wtf, weather it's changes or it's safe to keep this object.

iobuhov (changes requested):

packages/shared/widget-plugin-filtering/src/stores/NumberInputFilterStore.ts
rahmanunver: Got in touch, first impression no changes apart from the top of the props assigning initially, I'd say a valid suggestion from your side.

packages/shared/widget-plugin-filtering/src/stores/NumberInputFilterStore.ts
rahmanunver: made the change

packages/shared/widget-plugin-filtering/src/stores/NumberInputFilterStore.ts
rahmanunver: removed the formatter updates

packages/shared/widget-plugin-filtering/src/stores/Argument.ts
```diff
@@ -53,9 +53,10 @@ export class ArgumentBase<T extends Big | Date | string> {
     set displayValue(textValue: string) {
         this._lastSetTextValue = textValue;
         const parseResult = this._formatter?.parse(this._lastSetTextValue);
-        this.isValid = parseResult.valid;
-        if (parseResult.valid && !this.equals(this._value, parseResult.value)) {
-            this._value = parseResult.value;
+
+        this.isValid = parseResult.valid || (parseResult as any).type === "AutoNumber";
+        if ((parseResult as any).valid || (this.isValid && !this.equals(this._value, (parseResult as any).value))) {
+            this._value = (parseResult as any).value;
```
iobuhov, line 59: Why you changing this part? Formatter should work as before I think.

rahmanunver: If the formatter gets the value as `""` then a `new Big(value)` isn't created, which falls into catch block, eventually leaving the empty filter value unset. Looking for another way to handle this.

iobuhov: Then we have to fix this in formatter implementation. Just check what is returned from other formatter (Integer e.g.) and do same in the fix

iobuhov (approved):

leonardomendix (approved):

PR #1269, opened 2024-10-17
[WC-2667]: Gallery default filter value fix

rahmanunver:
### Pull request type

---

### Description
Added a setFilter method to have properly filtered with the initial default filter value.

### DIFF
--- a/packages/pluggableWidgets/gallery-web/src/helpers/useRootGalleryStore.ts
+++ b/packages/pluggableWidgets/gallery-web/src/helpers/useRootGalleryStore.ts
@@ -19,9 +19,6 @@ export function useRootGalleryStore(props: GalleryContainerProps): RootGallerySt
     useEffect(() => {
         const disposers: IReactionDisposer[] = [];
 
-        // Apply initial filter condition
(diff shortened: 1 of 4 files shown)


packages/pluggableWidgets/gallery-web/src/helpers/useRootGalleryStore.ts
```diff
@@ -19,6 +19,9 @@ export function useRootGalleryStore(props: GalleryContainerProps): RootGallerySt
     useEffect(() => {
         const disposers: IReactionDisposer[] = [];
 
+        // Apply initial filter condition
+        datasourceRef.current.setFilter(rootStore.conditions);
+
```
r0b1n, line 24: We can make it even simpler, mobx has a [special flag](https://mobx.js.org/reactions.html#fireimmediately-_reaction_) to reactions to let them run immediately after initialization. So we can pass `{ fireImmediately: true }` to the reaction down this file.

r0b1n (approved):

iobuhov (approved):
LGTM

PR #1270, opened 2024-10-18
feat(Datagrid-web): added pagination for load more & virtual scrolling

sharadsums:
**Pull request type**
Added feature to have pagination also visible on virtual scrolling & loadmore.

**Description**
This is to facilitate users when they want to see paging count in all pagination option
it will be in format << 1 to page item count of total count>> for virtual scrolling and load more

**What should be covered while testing?**


sharadsums:
Hi @iobuhov 
I have done my changes in this new branch taking from main

gjulivan:
implemented correctly in https://github.com/mendix/web-widgets/pull/1392

PR #1290, opened 2024-10-28
chore(datagrid): remove unused imports

gjulivan:
### Pull request type

---

### Description

**"string.prototype.replaceall"** is a pollyfil that we don't need anymore
**"deep-equal"** also is not being used.

this 2 package seems to have a lot of dependency tree scanned in snyk. 
Thus, if not being used, let's remove it.


iobuhov (approved):
LGTM

samuelreichert (approved):

PR #1302, opened 2024-10-31
[WC-2680] onChange not trigger on empty or not empty filter

gjulivan:
### Pull request type

---

### DIFF
--- a/packages/pluggableWidgets/datagrid-number-filter-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/datagrid-number-filter-web/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+-   We fixed an issue where onChange wouldn't triggered on empty or not empty filter.
+
 ## [2.8.3] - 2024-10-31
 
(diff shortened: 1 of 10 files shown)


samuelreichert (dismissed):

iobuhov (changes requested):
I think we should not use Math.random to trigger change...

iobuhov:
I think it should be fixed on useBasicSync level. Just change line 26 from
```
return reaction(() => [store.arg1.value, store.arg2.value], createPusher(pbox));
```
To this:
```
return reaction(() => [store.arg1.value, store.arg2.value, store.filterFunction], createPusher(pbox));
```

gjulivan:
> I think it should be fixed on useBasicSync level. Just change line 26 from
> 
> ```
> return reaction(() => [store.arg1.value, store.arg2.value], createPusher(pbox));
> ```
> 
> To this:
> 
> ```
> return reaction(() => [store.arg1.value, store.arg2.value, store.filterFunction], createPusher(pbox));
> ```

but this will trigger onchange for every filter change, of which we don't need.

iobuhov (dismissed):
LGTM

samuelreichert (dismissed):

leonardomendix (approved):

iobuhov (approved):

PR #1313, opened 2024-11-01
[WC-2710] allow set reference filter value

gjulivan:
### Pull request type

---

### Description

value have to set to guid, which supported by nanoflow
something like this:
{1F147F10-E466-428A-AABF-29F6ECBFE185}

or for multi select,
use NPE templist to get guid and reduce it to comma separated string.

{6B9AE64C-F2D5-41A8-A4FA-3C680C22CD6F}

### DIFF
--- a/packages/shared/widget-plugin-filtering/src/controllers/RefFilterController.ts
+++ b/packages/shared/widget-plugin-filtering/src/controllers/RefFilterController.ts
@@ -90,6 +90,15 @@ export class RefFilterController {
         }
     };
 
+    handleSetValue = (useDefaultValue: boolean, params: { stringValue: string }): void => {
+        if (useDefaultValue) {
+            this.store.reset();
+            return;
+        }
+
+        this.store.replace(params.stringValue.split(","));
+    };
+
     handleResetValue = (useDefaultValue: boolean): void => {
         if (useDefaultValue) {
             this.store.reset();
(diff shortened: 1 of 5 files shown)


packages/shared/widget-plugin-filtering/src/controllers/RefFilterController.ts
```diff
@@ -90,6 +90,14 @@ export class RefFilterController {
         }
     };
 
+    handleSetValue = (useDefaultValue: boolean, params: { operators: any; stringValue: string }): void => {
+        if (useDefaultValue) {
+            this.store.reset();
+            return;
+        }
+        this.store.replace([params.stringValue]);
```
iobuhov, line 98: Are we going to support multiselect? If yes, then split stringValue with `split(",")` to support multiselect?

iobuhov (approved):

packages/shared/widget-plugin-filtering/src/controllers/RefFilterController.ts
gjulivan: updated

packages/shared/widget-plugin-filtering/src/controllers/RefFilterController.ts
```diff
@@ -90,6 +90,15 @@ export class RefFilterController {
         }
     };
 
+    handleSetValue = (useDefaultValue: boolean, params: { operators: any; stringValue: string }): void => {
```
leonardomendix, line 93: Are the operators necessary in the parameters here? As we don't have that in the dropdown filter.

leonardomendix (approved):

iobuhov (approved):

PR #1326, opened 2024-11-06
[WC-2576][WC-2717] Fix/a11y for filter type selector

iobuhov:
### Pull request type

---

### Description
Enhanced `floating-ui` for better positioning and `downshift` for improved dropdown management.

#### Screen Reader Integration

1. Navigate through the application using a screen reader (e.g., NVDA, VoiceOver).
2. Verify that all interactive elements (buttons, links, form fields) are announced correctly by the screen reader.
3. Ensure that any dynamic content updates (e.g., modal dialogs, notifications) are also announced by the screen reader.

#### Filter Type Select Button Accessibility

1. Navigate to the filter type select button using the keyboard.
2. Press the `Enter` key and verify that the select menu opens.
3. Press the `Space` key and verify that the select menu opens.
4. Use the `Arrow` keys to navigate through the options in the select menu.
5. Verify that the select menu opens and navigates correctly with each key press.


automation/utils/bin/rui-create-translation.ts
gjulivan, line 1: is this removed to somewhere else? why is it empty?

packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid-filters.scss
```diff
@@ -60,55 +60,23 @@ $dg-item-min-height: 32px;
...
-        }
     }
 }
 
 .filter-selectors {
+    position: absolute;
+    width: max-content;
+    left: 0;
     padding: 0;
+    margin: 0;
```
gjulivan, line 71: spacing configuration is being removed from the filter?

packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid-scroll.scss
```diff
@@ -1,14 +1,16 @@
...
-                position: -webkit-sticky; /* Safari */
-                position: sticky;
-                z-index: 50;
-            }
-        }
-    }
+.container-stuck + .table .th {
+    position: -webkit-sticky; /* Safari */
+    position: sticky;
+    z-index: 10;
```
gjulivan, line 4: i'm not sure about this.
combobox z-index is 25,

in a position absolute,
sticky header should go above that.
in case datagrid DOM rendered after combobox.

i made this before: https://mendix.atlassian.net/wiki/spaces/WC/pages/2410774795/Z-Index+war

packages/shared/widget-plugin-filtering/src/controls/filter-selector/FilterSelector.tsx
```diff
@@ -0,0 +1,108 @@
...
+    open: boolean;
+    buttonProps: JSX.IntrinsicElements["button"];
+    listboxProps: JSX.IntrinsicElements["ul"];
+    getItemProps: UseSelectPropGetters<Option>["getItemProps"];
+    selectedItem: Option | null;
+    highlightedIndex: number;
+    floatingStyles: React.CSSProperties;
+}
+
+function useController(props: FilterSelectorProps): ViewProps {
```
gjulivan, line 66: i don't think this is a "controller"
we have the Ref and Static Filter controller for it. which makes this ambiguous naming.
this is more to MVVM pattern, which made this as "View Model" that ties solely to the FilterSelector view.
maybe separate this into a new file inside the same folder.
so the adapter to downshift and floating-ui localized to that file.

gjulivan:
super nice improvement. let's make it works!

packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid-scroll.scss
iobuhov: Yeah, good point, I would missed that 👍

packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid-filters.scss
iobuhov: It's to align with previous styles.

leonardomendix (approved):

gjulivan (approved):

PR #1328, opened 2024-11-06
fix: clearer type mismatch error message

gjulivan:
### Pull request type

---

### DIFF
--- a/packages/shared/widget-plugin-filtering/src/errors.ts
+++ b/packages/shared/widget-plugin-filtering/src/errors.ts
@@ -23,12 +23,12 @@ export const EMISSINGSTORE: APIError = Object.freeze({
     message: "Unable to get filter store. Check parent widget configuration."
 });
 
-export const ESTORETYPE: APIError = Object.freeze({
-    code: Code.ESTORETYPE,
-    message:
-        "The type of the filter and parent widget configuration is incompatible. The filter must " +
-        "be used with correct attribute/group type."
-});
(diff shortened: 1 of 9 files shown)


packages/shared/widget-plugin-filtering/src/errors.ts
```diff
@@ -23,12 +23,12 @@ export const EMISSINGSTORE: APIError = Object.freeze({
...
-    message:
-        "The type of the filter and parent widget configuration is incompatible. The filter must " +
-        "be used with correct attribute/group type."
-});
+export const ESTORETYPEMISTMATCH: (...args: string[]) => APIError = (...args: string[]): APIError => {
+    return Object.freeze({
+        code: Code.ESTORETYPE,
+        message: `The ${args[0]} is expecting ${args[1]} but received ${args[2]}. Ensure it being placed on a compatible parent configuration`
+    });
+};
```
iobuhov, line 31: Let's create normal function? All capital in `ESTORETYPEMISTMATCH()` is bit weird

iobuhov: Also, instead of array of arguments, I would use params object to pass arguments as properties

gjulivan: resolved and updated

iobuhov (approved):

leonardomendix (approved):

leonardomendix (approved):

iobuhov (approved):

PR #1354, opened 2024-11-19
[WC-2737] Fix export to excel issues

r0b1n:
### Pull request type

### Description

While switching pages it could happen that the new page is mounted while the old one is not yet unmounted, this led to issue with export controllers not being placed for the new data grids. This change makes it always override the existing controller, but removing a controller on unmount stage only affects own controller.


iobuhov (approved):
LGTM

leonardomendix (approved):

PR #1360, opened 2024-11-26
fix: unavailable caption shows on lazy load

gjulivan:
### Pull request type

---

### DIFF
--- a/packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelection.tsx
+++ b/packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelection.tsx
@@ -1,5 +1,5 @@
 import classNames from "classnames";
-import { Fragment, ReactElement, createElement, useRef } from "react";
+import { Fragment, ReactElement, createElement, useMemo, useRef } from "react";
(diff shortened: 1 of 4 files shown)


packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelection.tsx
```diff
@@ -42,6 +42,11 @@ export function SingleSelection({
         readOnly: selector.readOnly
     });
 
+    const selectedItemCaption = useMemo(
+        () => selector.caption.render(selectedItem, "label"),
+        [selectedItem, selector.status]
```
rahmanunver, line 47: `selector.caption` as dependency?

rahmanunver (approved):

leonardomendix (dismissed):

packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelection.tsx
gjulivan: not needed. we expect it to change only on selectedItem change and status change from loading to available.

rahmanunver (approved):

PR #1390, opened 2025-01-09
[WC-2705] feat(pie-chart): listen to widget

gjulivan:
### Pull request type

---


packages/pluggableWidgets/pie-doughnut-chart-web/src/hooks/data.ts
```diff
@@ -71,7 +73,15 @@ export const usePieChartDataSeries = ({
         tooltipHoverText
     ]);
 
-    const onClick = onClickAction ? (item: ObjectItem) => executeAction(onClickAction?.get(item)) : undefined;
+    const onClick = useCallback(
+        (item: ObjectItem) => {
+            executeAction(onClickAction?.get(item));
+            if (optionsSourceDatabaseItemSelection && optionsSourceDatabaseItemSelection.type === "Single") {
+                (optionsSourceDatabaseItemSelection as SelectionSingleValue).setSelection(item);
```
r0b1n, line 80: We probably don't need to cast it, it should have proper typings already.

packages/pluggableWidgets/pie-doughnut-chart-web/src/hooks/data.ts
```diff
@@ -71,7 +73,15 @@ export const usePieChartDataSeries = ({
...
+            }
+        },
+        [onClickAction]
```
r0b1n, line 83: I guess the quality check is correct here, it is better to add a dependency.

packages/pluggableWidgets/pie-doughnut-chart-web/src/PieChart.xml
```diff
@@ -51,6 +51,14 @@
                     <description />
                     <returnType type="String" />
                 </property>
+                <property key="optionsSourceDatabaseItemSelection" type="selection" dataSource="seriesDataSource">
```
r0b1n, line 54: We don't have options source from database in this widget, so we can simplify the name

packages/pluggableWidgets/pie-doughnut-chart-web/src/hooks/data.ts
gjulivan: fixed.

packages/pluggableWidgets/pie-doughnut-chart-web/src/hooks/data.ts
gjulivan: fixed.

r0b1n (approved):

leonardomendix (approved):

iobuhov (approved):

PR #1393, opened 2025-01-14
[WC-2758] combobox: load initial data on database source

gjulivan:
### Pull request type

---

### Description

Issue: combobox database source, with lazy load, don't have object data on initial load.
this cause combobox to not be able to render caption on initial load.

fix: load single object based on stored value attribute data on initial render.
method: use database filter. set limit to 1.

risk: because this is using literal filter (not object guid retrieval), data could be duplicate,
but because the data used is the stored value attribute, this should already be unique (otherwise, model failed)

this method is not needed on multi select, because database multi select cannot store attribute value.


packages/pluggableWidgets/combobox-web/src/helpers/BaseDatasourceOptionsProvider.ts
```diff
@@ -60,6 +60,15 @@ export class BaseDatasourceOptionsProvider extends BaseOptionsProvider<ObjectIte
         }
     }
 
+    // used for initial load of selected value in case options are lazy loaded
+    loadSelectedValue(attributeValue: string): void {
+        if (this.lazyLoading && this.ds && this.attributeId) {
+            const filterCondition = datasourceFilter("containsExact", attributeValue, this.attributeId);
+            this.ds?.setFilter(filterCondition);
+            this.ds.setLimit(this.attributeId.length);
```
r0b1n, line 68: What does `this.attributeId.length` return and why do we set it is limit?

packages/pluggableWidgets/combobox-web/src/helpers/Database/DatabaseSingleSelectionSelector.ts
```diff
@@ -69,13 +69,17 @@ export class DatabaseSingleSelectionSelector<
...
                     this.lastSetValue = this._attr.value;
-                    if (!_valuesIsEqual(this.values.getEmptyValue(), targetAttribute.value)) {
+                    if (targetAttribute.value) {
                         const obj = this.options.getAll().find(option => {
                             return _valuesIsEqual(targetAttribute.value, this.values.get(option));
                         });
                         if (obj) {
                             this.currentId = obj;
                         } else {
+                            if (targetAttribute.value) {
```
r0b1n, line 79: This looks redundant as we already checked that it is true at line 72.

packages/pluggableWidgets/combobox-web/src/helpers/Database/DatabaseSingleSelectionSelector.ts
```diff
@@ -69,13 +69,17 @@ export class DatabaseSingleSelectionSelector<
             if (this.lastSetValue === null || !_valuesIsEqual(this.lastSetValue, targetAttribute.value)) {
                 if (ds.status === "available") {
                     this.lastSetValue = this._attr.value;
-                    if (!_valuesIsEqual(this.values.getEmptyValue(), targetAttribute.value)) {
+                    if (targetAttribute.value) {
```
r0b1n, line 72: Why is this check gone? I guess this will run the code even if the value is equals to "empty value", which seems incorrect.

packages/pluggableWidgets/combobox-web/src/helpers/BaseDatasourceOptionsProvider.ts
gjulivan: basically it returns 1. because only works for single selection. database multi selection don't have attributes set.

set limit to 1 will fetch the datasource, because on lazy load, initial limit is 0.

packages/pluggableWidgets/combobox-web/src/helpers/Database/DatabaseSingleSelectionSelector.ts
gjulivan: empty value is default value. initially we did a hack by setting default value to $currentObject/attribute.
this is no longer applicable.

packages/pluggableWidgets/combobox-web/src/helpers/Database/DatabaseSingleSelectionSelector.ts
gjulivan: correct. will fix.

gjulivan: fixed

packages/pluggableWidgets/combobox-web/src/helpers/BaseDatasourceOptionsProvider.ts
gjulivan: @r0b1n you are correct. this is string length, 
updated to hardcode: 1

gjulivan:
continue on https://github.com/mendix/web-widgets/pull/1408

PR #1408, opened 2025-01-23
[WC-2776][WC-2758]

gjulivan:
### Pull request type

---


packages/pluggableWidgets/combobox-web/src/helpers/Association/AssociationSimpleCaptionsProvider.tsx
```diff
@@ -44,12 +52,12 @@ export class AssociationSimpleCaptionsProvider implements CaptionsProvider {
...
         }
 
-        const captionValue = this.formatter.get(item);
-        if (!captionValue || captionValue.status === "unavailable") {
+        this.value = this.formatter.get(item);
+        if (!this.value || this.value.status === "unavailable") {
             return this.unavailableCaption;
         }
 
-        return captionValue.value ?? "";
```
r0b1n, line 52: Adding this field feels wrong, and it took me a while to find where this value is used. 
1. Changing a field in a getter is quite brutal, a getter should not have side effects.
2. If we must keep it, name it something like `lastSeenCaptionInGet` (but I hope there is a way to not add this field).

packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelection.tsx
```diff
@@ -44,7 +44,14 @@ export function SingleSelection({
...
     const selectedItemCaption = useMemo(
         () => selector.caption.render(selectedItem, "label"),
-        [selectedItem, selector.status, selector.caption.emptyCaption]
+        [
+            selectedItem,
+            selector.status,
+            selector.caption,
+            selector.caption.emptyCaption,
+            selector.currentId,
+            selector.caption.value?.value
```
r0b1n, line 53: Why should it depend on this `value` thingy? I naively assume that having `selectedItem` is enough as it is what is getting passed to `render()` and `get()` under the hood.

packages/pluggableWidgets/combobox-web/src/helpers/Association/AssociationSimpleCaptionsProvider.tsx
gjulivan: as discussed before,
either this, or we can expose the formatter. 

which way do you think better?
@r0b1n

packages/pluggableWidgets/combobox-web/src/components/SingleSelection/SingleSelection.tsx
gjulivan: as discussed before, if caption is expression, it will be retrieved at later stage, and no way to dispatch changes for now.

packages/pluggableWidgets/combobox-web/src/helpers/Association/AssociationSimpleCaptionsProvider.tsx
gjulivan: resolved by switching to expose formatter

r0b1n (dismissed):

leonardomendix (approved):

r0b1n (approved):

PR #1421, opened 2025-02-03
[WIP] wc 2773/focus lost on change

iobuhov:
### Pull request type

---


packages/pluggableWidgets/datagrid-web/src/components/GridBody.tsx
```diff
@@ -1,16 +1,84 @@
...
 
 export function GridBody(props: Props): ReactElement {
+    const { paginationType, setPage, hasMoreItems, style, children } = props;
+    const isInfinite = paginationType === "virtualScrolling";
+    const [trackScrolling, bodySize, containerRef] = useInfiniteControl({
+        hasMoreItems,
+        isInfinite,
+        setPage
+    });
+    const baseClass = "widget-datagrid-grid-body table-content";
```
r0b1n, line 32: I see only one usage of this const, I guess we can inline it.

packages/pluggableWidgets/datagrid-web/src/controllers/DatasourceController.ts
```diff
@@ -0,0 +1,146 @@
...
+            return false;
+        }
+        return this.isDSLoading;
+    }
+
+    get isRefreshing(): boolean {
+        return this.refreshing;
+    }
+
+    get isLoadingMore(): boolean {
```
r0b1n, line 76: We can possibly replace Loading word in the name of this field. I am thinking we can use Fetching instead: isFetching, isFetchingNextPage, isFetchingNextBatch or something.

iobuhov: I like `isFetching`, short and not very specific

rahmanunver:
Are mobx changes here because of rebase?

iobuhov:
> Are mobx changes here because of rebase?

No, the mobx-kit is introduced in this PR

packages/pluggableWidgets/datagrid-web/src/controllers/DatasourceController.ts
iobuhov: done

packages/pluggableWidgets/datagrid-web/src/components/GridBody.tsx
iobuhov: done

packages/pluggableWidgets/datagrid-web/src/controllers/DatasourceController.ts
```diff
@@ -0,0 +1,146 @@
+import { DerivedPropsGate } from "@mendix/widget-plugin-mobx-kit/props-gate";
+import { ReactiveController, ReactiveControllerHost } from "@mendix/widget-plugin-mobx-kit/reactive-controller";
+import { ListValue, ValueStatus } from "mendix";
+import { action, autorun, makeAutoObservable } from "mobx";
+import { QueryController } from "./query-controller";
+
+type Gate = DerivedPropsGate<{ datasource: ListValue }>;
+type Spec = { gate: Gate };
```
r0b1n, line 8: Sorry for boring naming comments, I am just exploring the code and firing what comes to mind. If I see correctly `Spec` as an alternative to `Props` for react components. I think we can name and treat them the same way as for react props for the sake of consistency. 
I naming examples like:
`DatasourceControllerSpec` or `DatasourceControllerSpecs` (as plural). Same for `DerivedLoaderControllerSpecs`, `PaginationControllerSpecs`, etc.

iobuhov: Well, I assumed "type Spec" to alway be a local type, but you right, we can improve naming.

packages/pluggableWidgets/datagrid-web/src/controllers/StateSyncController.ts
```diff
@@ -0,0 +1,84 @@
...
+    header: Header;
+};
+
+export class StateSyncController implements ReactiveController {
+    private columns: Columns;
+    private header: Header;
+    private query: QueryController;
+
+    constructor(host: ReactiveControllerHost, spec: Spec) {
+        host.addController(this);
```
r0b1n, line 31: Is there a particular reason to do this in the constructor here, instead of doing it in the parent like: `this.addController(new StateSyncController(...))` ? This feels like an extra dependency that we can avoid having.

iobuhov: I think this way it's bit easer for testing - you don't have to mock whole host to run setup. And it's just my copy of Lit API. I would keep it like this until we have really solid reason to call it from host itself. I kinda like having control on controller side, but we can discuss of course.

packages/pluggableWidgets/datagrid-web/src/helpers/state/column/ColumnStore.tsx
r0b1n, line 1: I liked early returns better :)

iobuhov: Agree, I reverted 😌

leonardomendix (approved):

samuelreichert (approved):

PR #1427, opened 2025-02-04
[WC-2762]: Accordion state/animation fix

rahmanunver:
### Pull request type

---

### Description
Added logic for both handling the collapsed state correctly in between different entity instance switches through drop-downs, and the way it triggers the animations.

Test project included covers the scenario pretty well. We need to make sure the issue is resolved for the scenario,  not expecting to have any new introduced cases.


samuelreichert (approved):

gjulivan:
rework on https://github.com/mendix/web-widgets/pull/1434

PR #1436, opened 2025-02-11
[WC-2678]: Add playground to Custom Chart 

samuelreichert:
### Pull request type

---


packages/pluggableWidgets/custom-chart-web/src/controllers/Host.ts
```diff
@@ -0,0 +1,10 @@
+import { BaseControllerHost } from "@mendix/widget-plugin-mobx-kit/BaseControllerHost";
+import { ResizeController } from "./ResizeController";
+
+export class Host extends BaseControllerHost {
```
iobuhov, line 4: "Host" is too generic, can we rename it to "CustomChartControllerHost"?

packages/pluggableWidgets/custom-chart-web/src/hooks/useCustomChart.ts
```diff
@@ -1,19 +1,30 @@
+import { EditorStoreState, PlaygroundData } from "@mendix/shared-charts/main";
```
iobuhov, line 1: I think we need extract some logic from this hook into controllers. Also resize logic is duplicated in ResizeController.

iobuhov: 1. We should minimize usage of debounce
2. Merging props (updateData) can be done in separate function/module
3. Same about managing PlotlyChart

packages/shared/widget-plugin-mobx-kit/src/GateProvider.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    get props(): T {
+        return this.atom.props;
+    }
+}
+
+class PropsAtom<T> {
+    constructor(public props: T) {
+        makeObservable(this, { props: observable.struct, setProps: action });
```
iobuhov, line 16: Make sure to copy code from my branch, I updated this class and added some comment in there.

iobuhov (changes requested):
LGTM, but need some adjustments.

rahmanunver (approved):

gjulivan (approved):

PR #1456, opened 2025-02-24
[WC-2723]: Add rollup config for i18n locales (Combobox)

rahmanunver:
### Pull request type

No code changes (changes to documentation, CI, metadata, etc.)

---

### Description
Added rollup config for including locales create by create-translation

### DIFF
--- a/packages/pluggableWidgets/combobox-web/rollup.config.js
+++ b/packages/pluggableWidgets/combobox-web/rollup.config.js
@@ -1,22 +1,20 @@
 const { join } = require("path");
-const { cp, mkdir } = require("shelljs");
+const { cp, mkdir, rm } = require("shelljs");
 
 const sourcePath = process.cwd();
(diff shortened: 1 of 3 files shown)


packages/pluggableWidgets/combobox-web/rollup.config.js
gjulivan, line 1: I think this is only to copy locales folder to mpk,
we also need to make sure that the locales is running before releasing it. otherwise, there will be no locales folder in the release pipeline.

rahmanunver: Updated the turbo.json to run create-tranlsation before building.

gjulivan:
tested and locales not inside the mpk when built

gjulivan (dismissed):

leonardomendix (approved):

gjulivan (approved):

PR #1481, opened 2025-03-11
fix: update column selector to use floating ui

gjulivan:
column selector overflow detection is not reliable using old method.

this PR is to change it to use floating-ui instead.


packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid.scss
```diff
@@ -222,6 +222,8 @@ $root: ".widget-datagrid";
                 padding: 0 16px;
                 background: var(--bg-color-secondary, $dg-background-color);
                 z-index: 102;
+                overflow-y: auto;
+                max-width: fit-content;
```
iobuhov, line 226: Are you sure about this one? Probably having fixed max-width will be better.

packages/pluggableWidgets/datagrid-web/src/components/ColumnSelector.tsx
```diff
@@ -15,31 +14,58 @@ export interface ColumnSelectorProps {
...
+    const { getReferenceProps, getFloatingProps } = useInteractions([click, dismiss]);
 
     const label = props.label ?? "Column selector";
 
-    const isInViewport = useIsElementInViewport(optionsRef);
-
     const firstHidableColumnIndex = useMemo(() => props.columns.findIndex(c => c.canHide), [props.columns]);
     const lastHidableColumnIndex = useMemo(() => props.columns.map(c => c.canHide).lastIndexOf(true), [props.columns]);
 
+    useLayoutEffect(() => {
```
iobuhov, line 46: I think instead of trying to use `setTimer` here in there, we should use FocusManager. Also worth to look at `useListNavigation`. I see floating ui have very nice example.
https://codesandbox.io/p/sandbox/fragrant-water-bsuirj?file=%2Fsrc%2FApp.tsx%3A126%2C7-126%2C22

iobuhov: I think it might be even faster to do with downshift

gjulivan: tried this. on tab out from last item, the focus doesn't go back to the triggering button.
onOpenChange does not even detect the event.
a lot of unit test for column selectors are failing.

gjulivan: using downshift might work, but that's over-engineering, isn't it?
we just need the dropdown to work in a good position.

iobuhov: I though that we can improve a11y as well. Not just fix position.

gjulivan: yes, but adding downshift will change the listbox behavior, right?
(added input filter, etc) 
it creates another sets of discussion of "how things should be"
which requires refinement and design changes.
Out of scope for this support ticket and we can discuss and still adding it later if necessary.

I have added "useRole" for simple a11y improvement.

iobuhov (dismissed):

leonardomendix (approved):

r0b1n (approved):

leonardomendix (approved):

PR #1482, opened 2025-03-11
[WC-2872] Fixed refresh interval to use seconds instead of ms

leonardomendix:
### Pull request type

---

### Description

Fixed refresh interval to use seconds instead of ms.

Set a refresh interval in the DataGrid 2 and verify in the network request if the calls are made at the intervals configured in the widget.

### DIFF
--- a/packages/pluggableWidgets/datagrid-web/src/controllers/RefreshController.ts
+++ b/packages/pluggableWidgets/datagrid-web/src/controllers/RefreshController.ts
@@ -18,7 +18,7 @@ export class RefreshController implements ReactiveController {
     constructor(host: ReactiveControllerHost, spec: RefreshControllerSpec) {
         host.addController(this);
         this.query = spec.query;
-        this.delay = Math.max(spec.delay * 1000, 0);
+        this.delay = Math.max(spec.delay, 0);
     }
 
     setup(): (() => void) | void {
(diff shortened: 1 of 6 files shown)


gjulivan (approved):

packages/pluggableWidgets/datagrid-web/src/controllers/RefreshController.ts
```diff
@@ -18,7 +18,7 @@ export class RefreshController implements ReactiveController {
     constructor(host: ReactiveControllerHost, spec: RefreshControllerSpec) {
         host.addController(this);
         this.query = spec.query;
-        this.delay = Math.max(spec.delay, 0);
+        this.delay = Math.max(spec.delay * 1000, 0);
```
iobuhov, line 21: Can we change it not in this part?

iobuhov: I think this should be done in "RootGridStore", line 62

leonardomendix: Thanks!

Done! ✅

gjulivan (approved):

iobuhov (approved):
LGTM

PR #1511, opened 2025-03-27
[WC-2868] add number filter linked datasource

gjulivan:
### Pull request type

---

### DIFF
--- a/packages/pluggableWidgets/datagrid-number-filter-web/src/DatagridNumberFilter.editorConfig.ts
+++ b/packages/pluggableWidgets/datagrid-number-filter-web/src/DatagridNumberFilter.editorConfig.ts
@@ -1,10 +1,4 @@
-import {
-    ContainerProps,
-    ImageProps,
-    structurePreviewPalette,
-    StructurePreviewProps,
-    text
-} from "@mendix/widget-plugin-platform/preview/structure-preview-api";
+import { hidePropertyIn, Properties } from "@mendix/pluggable-widgets-tools";
 import {
     emptyIcon,
     emptyIconDark,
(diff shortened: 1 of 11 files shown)


packages/pluggableWidgets/datagrid-number-filter-web/src/DatagridNumberFilter.editorConfig.ts
```diff
@@ -23,25 +17,26 @@ import {
...
-            hidePropertiesIn(defaultProperties, values, ["onChange", "valueAttribute"]);
-        }
-    } else {
-        hidePropertyIn(defaultProperties, values, "advanced");
+
+    if (values.attrChoice === "auto") {
+        hidePropertyIn(defaultProperties, values, "attributes");
     }
+
+    hidePropertyIn(defaultProperties, {} as { linkedDs: unknown }, "linkedDs");
```
r0b1n, line 39: TODO: check if this is really needed, it should be either hidden by default so we don't have to do it, or the preview typings have to exposed by PWT.

r0b1n: Or let it be shown, as a reference to the user, it is not configurable anyways.

packages/shared/widget-plugin-filtering/src/stores/input/store-utils.ts
```diff
@@ -52,3 +52,13 @@ export function isStringFilter(store: InputFilterInterface): store is String_Inp
 export function isDateFilter(store: InputFilterInterface): store is Date_InputFilterInterface {
     return store.arg1.type === "date";
 }
+
+export function getFormatter<T>(attr: { formatter?: SimpleFormatter<T> }): SimpleFormatter<T> {
+    return (
+        attr.formatter ??
+        ({
+            format: v => v ?? "",
+            parse: v => ({ valid: true, value: v ?? "" })
```
r0b1n, line 61: I would expect this not to be enough for Big, as it has to do at least some formatting to string and back 🤔 .

packages/pluggableWidgets/datagrid-number-filter-web/src/DatagridNumberFilter.editorConfig.ts
iobuhov: done

packages/shared/widget-plugin-filtering/src/stores/input/store-utils.ts
iobuhov: removed

PR #1514, opened 2025-03-28
[WC-2865]: Inconsistent Filter Behavior

samuelreichert:
### Pull request type

---

### Description

Inconsistent behavior on dropdown filter with text selected items. It should have ellipsis but it was behaving like labels instead.


packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid-dropdown-filter.scss
```diff
@@ -316,6 +316,10 @@ $root: ".widget-dropdown-filter";
         &:has(+ #{$root}-selected-item)::after {
             content: ",";
         }
+
+        & + #{$root}-input {
```
iobuhov, line 320: This change will likely break keyboard a11y. I might be wrong. I would advice to check weather you can reach input, open/close menu and remove items using only keyboard.

gjulivan (approved):

PR #1533, opened 2025-04-16
[WC-2882] Use text input for number filter

r0b1n:
### Pull request type

---

### Description

Input of type=number is having build it behavior for number parsing, this messes up with locale-specific number parsing, switch to type=text. 
Bonus: highlight the input in red when number parsing was not successful.


iobuhov (approved):
LGTM. But seems like prettier format and our changelog format is out of sync.

leonardomendix (approved):

PR #1549, opened 2025-04-24
[WC-2897] Disable Select All if there are no items

r0b1n:
### Pull request type

Improvement

---

### Description

Disable the Select All if there are nothing to select [yet].

### DIFF
--- a/packages/shared/widget-plugin-component-kit/src/ThreeStateCheckBox.tsx
+++ b/packages/shared/widget-plugin-component-kit/src/ThreeStateCheckBox.tsx
@@ -4,11 +4,12 @@ export type ThreeStateCheckBoxEnum = "all" | "some" | "none";
 
 export interface ThreeStateCheckBoxProps {
     id?: string;
+    disabled?: boolean;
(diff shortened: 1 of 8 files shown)


packages/shared/widget-plugin-component-kit/src/ThreeStateCheckBox.tsx
```diff
@@ -28,6 +29,7 @@ export function ThreeStateCheckBox({ id, value, onChange }: ThreeStateCheckBoxPr
             className="three-state-checkbox"
             ref={checkboxRef}
             checked={value !== "none"}
+            disabled={disabled}
```
gjulivan, line 32: there are no disabled=false 
disabled property should not exist at all if it is not disabled.

r0b1n: React is smart enough to handle this correctly, it omits the property if `false` is passed.

gjulivan (approved):

iobuhov (approved):

leonardomendix (approved):

PR #1553, opened 2025-04-30
[WC-2920]: Custom Chart config override

rahmanunver:
### Pull request type

---

### Description
We have size configuration for some elements of the custom chart, this pr adds conditional logic for size handling given for the layout of the chart, therefore if user does apply configuration to those "size handled" properties, they will be able override and use their own values.

### DIFF
--- a/packages/pluggableWidgets/custom-chart-web/src/controllers/ChartPropsController.ts
+++ b/packages/pluggableWidgets/custom-chart-web/src/controllers/ChartPropsController.ts
@@ -84,14 +84,18 @@ export class ChartPropsController implements ReactiveController {
             height: this.sizeProvider.height,
             autosize: true,
             font: {
-                family: "Open Sans, sans-serif",
-                size: Math.max(12 * (this.sizeProvider.width / 1000), 8)
+                family: this.layout.font?.family ? this.layout.font.family : "Open Sans, sans-serif",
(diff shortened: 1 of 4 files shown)


packages/pluggableWidgets/custom-chart-web/src/controllers/ChartPropsController.ts
```diff
@@ -100,23 +104,27 @@ export class ChartPropsController implements ReactiveController {
...
-                r: 60,
-                t: 60,
-                b: 60,
-                pad: 10
+                l: this.layout.margin?.l ? this.layout.margin.l : 60,
+                r: this.layout.margin?.r ? this.layout.margin.r : 60,
+                t: this.layout.margin?.t ? this.layout.margin.t : 60,
+                b: this.layout.margin?.b ? this.layout.margin.b : 60,
+                pad: this.layout.margin?.pad ? this.layout.margin.pad : 10
             }
```
iobuhov, line 128: Can we just move spread operator to the end? This will override default values with values from `this.layout.margin`? Same applies to other props

rahmanunver: Updated 👍

leonardomendix (approved):

PR #1555, opened 2025-05-02
[WC-2914]: add support for custom fonts in RichText widget

samuelreichert:
### Pull request type

New Feature

---

### DIFF
--- a/packages/pluggableWidgets/rich-text-web/src/RichText.xml
+++ b/packages/pluggableWidgets/rich-text-web/src/RichText.xml
@@ -163,6 +163,17 @@
                     <caption>Enable spell checking</caption>
                     <description />
                 </property>
+                <property key="customFonts" type="object" isList="true" required="false">
+                    <caption>Custom fonts</caption>
+                    <description />
+                    <properties>
+                        <property key="fontName" type="textTemplate" required="false">
+                            <caption>Font name</caption>
+                            <category>Item</category>
+                            <description />
+                        </property>
+                    </properties>
+                </property>
             </propertyGroup>
         </propertyGroup>
         <propertyGroup caption="Custom toolbar">
--- a/packages/pluggableWidgets/rich-text-web/src/RichText.xml
+++ b/packages/pluggableWidgets/rich-text-web/src/RichText.xml
@@ -170,12 +170,12 @@
                         <property key="fontName" type="string" required="false">
                             <caption>Font name</caption>
                             <category>Item</category>
-                            <description>Choose the main font you want to use (e.g., Arial)</description>
+                            <description>A title for this font combination (e.g., Arial).</description>
                         </property>
                         <property key="fontStyle" type="string" required="false">
                             <caption>Font style</caption>
                             <category>Item</category>
-                            <description>List of fonts to use in case the first isn't available (e.g., arial, helvetica, sans-serif)</description>
+                            <description>The full CSS font-family declaration that will be applied (e.g., arial, helvetica, sans-serif).</description>
                         </property>
                     </properties>
                 </property>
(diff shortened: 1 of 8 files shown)


packages/pluggableWidgets/rich-text-web/src/__tests__/RichText.spec.tsx
```diff
@@ -43,7 +43,8 @@ describe("Rich Text", () => {
             maxHeightUnit: "none",
             maxHeight: 0,
             minHeight: 75,
-            OverflowY: "auto"
+            OverflowY: "auto",
```
rahmanunver, line 46: camelcase -> `overflowY`
IF it doesn't cause breaking change.

samuelreichert: I checked the `RichTextContainerProps` and OverflowY is written like this, better not change it on the unit tests. But I agree we need to understand why overflowY is capitalised on props.

packages/pluggableWidgets/rich-text-web/src/utils/formats/fonts.ts
```diff
@@ -46,6 +54,11 @@ class FontStyleAttributor extends StyleAttributor {
     }
 }
 
-const FontStyle = new FontStyleAttributor("font", "font-family", config);
-
-Quill.register(FontStyle, true);
+export function formatFonts(fonts: CustomFontsType[] = []): typeof FONT_LIST {
+    console.info("formatFonts", { fonts });
```
r0b1n, line 58: console.log

packages/pluggableWidgets/rich-text-web/src/RichText.xml
```diff
@@ -163,6 +163,17 @@
...
                     <description />
                 </property>
+                <property key="customFonts" type="object" isList="true" required="false">
+                    <caption>Custom fonts</caption>
+                    <description />
+                    <properties>
+                        <property key="fontName" type="textTemplate" required="false">
+                            <caption>Font name</caption>
+                            <category>Item</category>
+                            <description />
```
r0b1n, line 173: We should give this a description, showing one or two examples of what could be placed there and which format this should be.

packages/pluggableWidgets/rich-text-web/src/RichText.xml
```diff
@@ -163,6 +163,22 @@
...
                     <description />
                 </property>
+                <property key="customFonts" type="object" isList="true" required="false">
+                    <caption>Custom fonts</caption>
+                    <description />
+                    <properties>
+                        <property key="fontName" type="string" required="false">
+                            <caption>Font name</caption>
+                            <category>Item</category>
+                            <description>Choose the main font you want to use (e.g., Arial)</description>
```
r0b1n, line 173: I think this should say that this is title in the list of fonts. This technically can be anything, not the exact font name as I see this is not applied in styles.

packages/pluggableWidgets/rich-text-web/src/RichText.xml
```diff
@@ -163,6 +163,22 @@
...
+                        </property>
+                        <property key="fontStyle" type="string" required="false">
+                            <caption>Font style</caption>
+                            <category>Item</category>
+                            <description>List of fonts to use in case the first isn't available (e.g., arial, helvetica, sans-serif)</description>
```
r0b1n, line 178: This description makes me think those are fallback font's (and the main font shouldn't be specified), while in reality iiuc, this is the style that gets applied to `font-family`? We should describe this behavior instead, it is fine that it might be quite a bit technical, this is an advanced feature anyway.

r0b1n (approved):

leonardomendix (approved):

PR #1559, opened 2025-05-07
[WC-2937][combobox] implement custom readonly if target attribute is null

gjulivan:
### Pull request type

---

### Description

only for database datasource.
database datasource makes the target attribute for storing value as optional,
this makes system property Editability can't be read, because readonly ties heavily with the attribute values.

this PR allows combobox to use custom readonly configuration if there are no target attribute.

### DIFF
--- a/packages/pluggableWidgets/combobox-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/combobox-web/CHANGELOG.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - We fixed an issue where database selection does not trigger when set to empty value from external logic.
 
-### Breaking
+### Breaking changes
 
 - We implement custom editability configuration to support readOnly mode if there are no target attribute for database source. This changes is only affecting editability configuration for Database data source with no stored value.
 


packages/pluggableWidgets/combobox-web/src/Combobox.editorConfig.ts
```diff
@@ -129,6 +133,13 @@ export function getProperties(values: ComboboxPreviewProps, defaultProperties: P
         }
         if (values.databaseAttributeString.length === 0) {
             hidePropertiesIn(defaultProperties, values, ["optionsSourceDatabaseValueAttribute"]);
+            // @ts-expect-error ignore error due to the Editability is part of system properties
+            hidePropertiesIn(defaultProperties, values, ["Editability"]);
```
r0b1n, line 137: I think `ts` directive is a bit harsh here, maybe we can change typings of `values` to something like `ComboboxPreviewProps & {"Editability"?: unknown}` in the function declaration part.

packages/pluggableWidgets/combobox-web/src/helpers/Database/utils.ts
```diff
@@ -86,3 +86,20 @@ export function extractDatabaseProps(props: ComboboxContainerProps): ExtractionR
...
+): boolean {
+    if (targetAttribute) {
+        return targetAttribute.readOnly;
+    }
+    if (customEditability === "never") {
+        return true;
+    }
+    if (customEditability === "conditionally") {
+        return customEditabilityExpression.value ?? false;
+    }
```
r0b1n, line 103: This expression might not be loaded [yet], and I think we should assume it is disabled if expression is still `loading` while editablily is set to `conditionally`. We also want to think of scenario when expression is `unavailable`, where disabling it also seems a good bet.

gjulivan: fixed by set to true

r0b1n (approved):

rahmanunver (approved):

packages/pluggableWidgets/combobox-web/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Breaking
```
leonardomendix, line 9:
```suggestion
### Breaking changes
```

leonardomendix (approved):

r0b1n (approved):

PR #1562, opened 2025-05-08
[WC-2922]: Add aggregate utils for charts

rahmanunver:
### Pull request type

---

### Description
Added a new `aggregations.ts` in the charts utilities to be used instead of the deprecated plotly.js aggregations.

All Charts (area, bar, bubble, column, line, time-series) should have working aggregations displayed in the charts/graphs.


packages/pluggableWidgets/bubble-chart-web/src/BubbleChart.tsx
```diff
@@ -75,18 +70,25 @@ export const BubbleChart = memo(
                 );
                 const markerColorExpression =
                     line.dataSet === "static" ? line.staticMarkerColor : line.dynamicMarkerColor;
+                const pts =
+                    line.aggregationType === "none"
+                        ? dataPoints
+                        : aggregateDataPoints(line.aggregationType, dataPoints);
```
gjulivan, line 76: checking type none is already done in the function aggregateDataPoints. this is no longer needed

packages/shared/charts/src/utils/aggregations.ts
```diff
@@ -0,0 +1,105 @@
+// packages/shared/charts/src/utils/aggregations.ts
+import { Datum } from "plotly.js-dist-min";
+import { PlotChartDataPoints } from "../hooks/usePlotChartDataSeries";
+
+export type AggregationType = "none" | "count" | "sum" | "avg" | "min" | "max" | "median" | "mode" | "first" | "last";
```
gjulivan, line 5: there is already a AggregationTypeEnum on usePlotChartDataSeries

gjulivan:
i think we can remove all `transform` property from all charts,
and simply put `aggregateDataPoints` inside `extractDataPoints` function.
this way all other charts code are intact.

gjulivan (approved):

iobuhov (approved):

gjulivan (approved):

PR #1627, opened 2025-06-16
[]: Migrate widgets to use enzyme-free command for unit tests

samuelreichert:
### Pull request type

---


packages/pluggableWidgets/datagrid-web/src/components/__tests__/ColumnResizer.spec.tsx
```diff
@@ -1,4 +1,5 @@
-import { render } from "enzyme";
+import "@testing-library/jest-dom";
```
iobuhov, line 1: Can we use asFragment as there is snapshot pollution

iobuhov (changes requested):

iobuhov (approved):

PR #1736, opened 2025-07-01
Test/remove enzyme from widgets

samuelreichert:
### Pull request type

---


rahmanunver (approved):
Would be cleaner to get rid of :enzyme-free completely, but I assume it's a gradual change from pwt side.

iobuhov (approved):

PR #1764, opened 2025-07-11
feat: initial setup of skiplink widget

HedwigAR:
### Pull request type

---


packages/pluggableWidgets/skiplink-web/src/ui/SkipLink.scss
```diff
@@ -0,0 +1,20 @@
+.skip-link {
```
iobuhov: We have some convention that we use - we use `.widget-` prefix in classes.
So, this will be `.widget-skip-link` use it please for naming.

packages/pluggableWidgets/skiplink-web/src/SkipLink.editorPreview.tsx
```diff
@@ -0,0 +1,54 @@
+import { ReactElement } from "react";
+import { SkipLinkPreviewProps } from "../typings/SkipLinkProps";
+
+export const preview = (props: SkipLinkPreviewProps): ReactElement => {
```
iobuhov: Do we want user to style this live in preview?

automation/run-e2e/docker/mxbuild.Dockerfile
```diff
@@ -24,10 +24,12 @@ echo "Downloading mxbuild ${MENDIX_VERSION} and docker building for ${BUILDPLATF
...
     && rm /tmp/mxbuild.tar.gz && \
 \
-    apt-get update -qqy && \
-    apt-get install -qqy libicu70 && \
+    rm -rf /var/lib/apt/lists/* && \
+    apt-get update --allow-insecure-repositories -qqy && \
+    apt-get install -qqy --allow-unauthenticated libicu70 && \
     apt-get -qqy remove --auto-remove wget && \
     apt-get clean && \
+    rm -rf /var/lib/apt/lists/* && \
```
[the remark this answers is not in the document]
HedwigAR: Because the widget only works in 11.1 and up. So the test project also has to be that, and it gave issues when updating everything. Leo made these changes to make sure it works. They're needed to make it work on Mac using Mx11 runtime & mxbuild

PR #1765, opened 2025-07-11
[WC-2838]: Implement RefreshIndicator component on Datagrid 2

samuelreichert:
### Pull request type

---

### Description

Implement a new property to show a refresh indicator. With the refresh indicator any datasource change triggers the progress bar on top of the Datagrid2.

### DIFF
--- a/packages/shared/widget-plugin-grid/src/__tests__/DatasourceController.spec.ts
+++ b/packages/shared/widget-plugin-grid/src/__tests__/DatasourceController.spec.ts
@@ -17,6 +17,7 @@ describe("DatasourceController loading states", () => {
         provider = new GateProvider({ datasource: list.loading() });
         host.setup();
         controller = new DatasourceController(host, { gate: provider.gate });
+        controller.setup();
     });
 
     describe("when datasource is loading", () => {
@@ -38,7 +39,6 @@ describe("DatasourceController loading states", () => {
         it("isRefreshing is true after refresh call", () => {
             provider.setProps({ datasource: list(0) });
             expect(provider.gate.props.datasource.status).toBe("available");
-            controller.setup();
             controller.refresh();
             expect(controller.isRefreshing).toBe(true);
             provider.setProps({ datasource: list.loading() });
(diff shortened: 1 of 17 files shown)


packages/shared/widget-plugin-grid/src/__tests__/DatasourceController.spec.ts
```diff
@@ -39,13 +39,13 @@ describe("DatasourceController loading states", () => {
             provider.setProps({ datasource: list.loading() });
             expect(provider.gate.props.datasource.status).toBe("loading");
             expect(controller.isRefreshing).toBe(true);
-            expect(controller.isLoading).toBe(false);
+            expect(controller.isFirstLoad).toBe(true);
```
iobuhov, line 42: I think something wrong here, isFirstLoad should be false after setProps on line 35. But it remains true, why?

packages/shared/widget-plugin-grid/src/__tests__/DatasourceController.spec.ts
```diff
@@ -21,8 +21,8 @@ describe("DatasourceController loading states", () => {
             provider.setProps({ datasource });
         });
 
-        it("isLoading returns true by default", () => {
-            expect(controller.isLoading).toBe(true);
+        it("isFirstLoad returns true by default", () => {
```
iobuhov, line 24: I think we should create more "real world" test. Let's use `BaseControllerHost` on line 13. And between lines 15 and 16 add `host.setup()` call. This will make sure we setup listeners.

packages/shared/widget-plugin-grid/src/query/DatasourceController.ts
```diff
@@ -105,6 +114,11 @@ export class DatasourceController implements ReactiveController, QueryController
     }
 
     setup(): () => void {
+        when(
```
iobuhov, line 117: I think to keep consistency, we should use `disposeBatch` here

iobuhov (approved):

leonardomendix (dismissed):

leonardomendix (approved):

gjulivan (approved):

PR #1771, opened 2025-07-14
[WC-2984] Events: Parameter type support

yordan-st:
### Pull request type

---

### Description
When configuring the Events widget to run specific action on load or on attribute change, previously user only could set a static value when setting the delay value or repeat interval.

I enhanced this by adding a the possibility to set these value with expressions. This way users have more flexibility when configuring their events.

- Toggling 'Value' or 'Expression' should display the correct property
- When a user configures expressions to calculate the delay or paramater value, this should be correctly set to the expected value


packages/pluggableWidgets/events-web/src/Events.tsx
```diff
@@ -9,20 +9,52 @@ export default function Events(props: EventsContainerProps): ReactElement {
...
+    let repeatInterval;
+    let onEventChangeDelay;
+    if (componentLoadDelayParameterType === "number") {
+        loadDelay = componentLoadDelayInteger;
+    } else {
+        loadDelay =
+            componentLoadDelayExpression.status === "available"
+                ? componentLoadDelayExpression.value.toNumber()
+                : undefined;
+    }
```
gjulivan, line 36: duplication

packages/pluggableWidgets/events-web/src/__tests__/AppEvents.spec.tsx
```diff
@@ -12,10 +13,18 @@ describe("App events (load)", () => {
             name: "app events",
             class: "app-events",
             onComponentLoad: actionValue(),
-            componentLoadDelay: 0,
-            onEventChangeDelay: 0,
+            componentLoadDelayParameterType: "number",
+            componentLoadDelayInteger: 0,
+            componentLoadDelayExpression: { value: undefined, status: ValueStatus.Unavailable },
```
r0b1n, line 18: There is `dynamicValue` helper for this, in the same place where `actionValue` is coming from.

packages/pluggableWidgets/events-web/src/Events.xml
```diff
@@ -11,18 +11,51 @@
...
+                    </enumerationValues>
+                </property>
+
                 <property key="componentLoadRepeatInterval" type="integer" required="true" defaultValue="30000">
-                    <caption>Delay</caption>
+                    <caption>Interval</caption>
                     <description>Interval between repeat action execution. Value is in milliseconds.</description>
                 </property>
+
+                <property key="componentLoadRepeatExpression" type="expression" required="false">
```
r0b1n, line 54: As originally property was called `componentLoadRepeatInterval` (and should stay this way, old names can't be changed), isn't `componentLoadRepeatIntervalExpression` is a better name? 
Same for `componentLoadRepeatParameterType` -> `componentLoadRepeatIntervalParameterType`. 


Previously we had `componentLoadRepeatInterval` - componentLoad is prefix, forget about that for a moment. So we have 
`RepeatInterval` as number, and we need to extend the properties, so we need following properties:

(Repeat interval) Parameter Type (switch between types)
(Repeat interval) Static (for static number )
(Repeat interval) Expression (for expression)

So logical names would be 
`[prefix]RepeatIntervalParameterType`
`[prefix]RepeatIntervalStatic`
`[prefix]RepeatIntervalExpression`

So far so good, but because static property already existed there under name `[prefix]RepeatInterval`, we have to keep it, so even though it is a bit inconvenient naming we have to do the following:

`[prefix]RepeatIntervalParameterType`
`[prefix]RepeatInterval` (old name, so that property value is not getting lost)
`[prefix]RepeatIntervalExpression`


Hope this makes sense.

Check other names and standardize the convention behind those names.

packages/pluggableWidgets/events-web/src/Events.tsx
```diff
@@ -42,7 +60,7 @@ export default function Events(props: EventsContainerProps): ReactElement {
                 }
             }
         },
-        delay: onEventChangeDelay,
+        delay: delayValue,
```
gjulivan, line 63: and this should not be the same as component load, right?

gjulivan (approved):

PR #1777, opened 2025-07-18
[WC-2845] File Uploader - Limit total number of files, and make it dynamic

gjulivan:
### Pull request type

---

### DIFF
--- a/packages/pluggableWidgets/file-uploader-web/src/FileUploader.editorConfig.ts
+++ b/packages/pluggableWidgets/file-uploader-web/src/FileUploader.editorConfig.ts
@@ -40,6 +40,12 @@ export function getProperties(
         hidePropertiesIn(properties, values, ["customButtons"]);
     }
 
+    if (values.maxFilePerUploadType === "expression") {
+        hidePropertiesIn(properties, values, ["maxFilesPerUpload"]);
+    } else {
+        hidePropertiesIn(properties, values, ["maxFilesPerUploadExpression"]);
+    }
+
     return properties;
 }
 
--- a/packages/pluggableWidgets/file-uploader-web/src/stores/FileUploaderStore.ts
+++ b/packages/pluggableWidgets/file-uploader-web/src/stores/FileUploaderStore.ts
@@ -1,6 +1,7 @@
-import { ListValue, ObjectItem } from "mendix";
-import { FileUploaderContainerProps, UploadModeEnum } from "../../typings/FileUploaderProps";
+import { DynamicValue, ListValue, ObjectItem } from "mendix";
+import { FileUploaderContainerProps, MaxFilePerUploadTypeEnum, UploadModeEnum } from "../../typings/FileUploaderProps";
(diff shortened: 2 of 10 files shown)


packages/pluggableWidgets/file-uploader-web/src/stores/FileUploaderStore.ts
```diff
@@ -113,6 +119,25 @@ export class FileUploaderStore {
             .join(", ");
     }
 
+    get maxFilesPerUpload(): number {
+        const expressionValue = this._maxFilesPerUpload.value;
+        if (expressionValue && !isNaN(Number(expressionValue))) {
+            return Number(expressionValue);
+        }
+        // Fallback to unlimited
+        return 0;
```
r0b1n, line 128: Big.js has [toNumber](https://mikemcl.github.io/big.js/#toN) which we can use. And looks like it can never return NaN, so we can exclude NaN check.

packages/pluggableWidgets/file-uploader-web/src/stores/FileUploaderStore.ts
```diff
@@ -128,7 +153,7 @@ export class FileUploaderStore {
 
         if (fileRejections.length && fileRejections[0].errors[0].code === "too-many-files") {
             this.setMessage(
-                this.translations.get("uploadFailureTooManyFilesMessage", this._maxFilesPerUpload.toString())
+                this.translations.get("uploadFailureTooManyFilesMessage", this.maxFilesPerUpload?.toString() ?? "")
```
r0b1n, line 156: It looks like `maxFilesPerUpload` never meant to return undefined, so this fallback is probably redundant.

packages/pluggableWidgets/file-uploader-web/src/FileUploader.editorConfig.ts
```diff
@@ -89,12 +89,7 @@ export function check(values: FileUploaderPreviewProps): Problem[] {
             }
         }
 
-        if (!values.maxFilesPerUpload || values.maxFilesPerUpload < 1) {
-            errors.push({
-                property: "maxFilesPerUpload",
-                message: "There must be at least one file per upload allowed."
-            });
-        }
+        // Note: maxFilesPerUpload validation is handled at runtime since it's an expression
```
r0b1n, line 92: I don't think this comment adds any value.

packages/pluggableWidgets/file-uploader-web/src/stores/FileUploaderStore.ts
gjulivan: fixed. updated

packages/pluggableWidgets/file-uploader-web/src/FileUploader.editorConfig.ts
gjulivan: removed.

r0b1n (approved):

PR #1821, opened 2025-08-11
[WC-349]: Prevent scroll on column selector focus

rahmanunver:
### Pull request type

---

### Description
The referenced column selector element was being focused after loading DG2 as a solution for focus behavior, it was causing a horizontal scroll issue on initial load. We are keeping the focus behavior same but preventing the scroll now.

A dg2 nested in container with column selector enabled. See the test project.


gjulivan:
need rebase @rahmanunver

leonardomendix (dismissed):

gjulivan (approved):

leonardomendix (approved):

PR #1827, opened 2025-08-12
[WC-2481] Image: Enhance dimension properties

yordan-st:
### Pull request type

---

### Description

Added support to configure a `min-height` and or `max-height` for the image when the height unit is set to `auto`
Added the viewport unit (`vh`) while setting the height, min-height and max-height values

- Go to the dimensions tab
- Set a different unit & value for every dimension related to height and test if it works in runtime

### DIFF
--- a/packages/pluggableWidgets/image-web/src/Image.tsx
+++ b/packages/pluggableWidgets/image-web/src/Image.tsx
@@ -2,6 +2,7 @@ import { ValueStatus } from "mendix";
 import { createElement, FunctionComponent, useCallback } from "react";
 import { ImageContainerProps } from "../typings/ImageProps";
 import { Image as ImageComponent, ImageType } from "./components/Image/Image";
+import { constructStyleObject } from "./utils/helpers";
 
 function getImageProps({
     datasource,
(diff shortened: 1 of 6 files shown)


packages/pluggableWidgets/image-web/src/Image.tsx
```diff
@@ -60,11 +61,22 @@ export const Image: FunctionComponent<ImageContainerProps> = props => {
     const { type, image } = getImageProps(props);
 
     const altText = props.alternativeText?.status === ValueStatus.Available ? props.alternativeText.value : undefined;
+    const styleObject = constructStyleObject(props);
+
+    const imageStyle = { ...props.style, ...styleObject };
+    console.warn("Image styling debug:", {
```
gjulivan, line 67: remove console

packages/pluggableWidgets/image-web/src/utils/helpers.ts
```diff
@@ -0,0 +1,30 @@
...
+    const { widthUnit, heightUnit, minHeightUnit, maxHeightUnit, width, height, minHeight, maxHeight } = props;
+
+    const imageStyle: Pick<CSSProperties, "width" | "height" | "minHeight" | "maxHeight" | "maxWidth"> = {};
+
+    imageStyle.width = `${width}${widthUnit === "pixels" ? "px" : "%"}`;
+    if (heightUnit === "auto") {
+        imageStyle.height = "auto";
+
+        if (minHeightUnit !== "none" && minHeight > 0) {
+            console.warn(minHeight);
```
leonardomendix, line 18: Hello,

Do we need this `console.warn`?

PR #1828, opened 2025-08-13
[WC-2961] Selection items count

iobuhov:
### Pull request type

---

### Description
Add new store that counts items in selection. New settings to customize text and layout changes in the footer.

- Pagination should remain unchanged
- New counter in bottom-left corner
- Should be visible only if 1 or more items selected

### DIFF
--- a/packages/pluggableWidgets/gallery-web/typings/GalleryProps.d.ts
+++ b/packages/pluggableWidgets/gallery-web/typings/GalleryProps.d.ts
@@ -55,8 +55,8 @@ export interface GalleryContainerProps {
     emptyMessageTitle?: DynamicValue<string>;
     ariaLabelListBox?: DynamicValue<string>;
     ariaLabelItem?: ListExpressionValue<string>;
-    sCountFmtSingular?: DynamicValue<string>;
-    sCountFmtPlural?: DynamicValue<string>;
+    selectedCountTemplateSingular?: DynamicValue<string>;
+    selectedCountTemplatePlural?: DynamicValue<string>;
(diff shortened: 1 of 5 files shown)


packages/pluggableWidgets/datagrid-web/src/components/WidgetFooter.tsx
```diff
@@ -14,16 +16,33 @@ export function WidgetFooter(props: WidgetFooterProps): ReactElement | null {
...
+                    {(pagingPosition === "bottom" || pagingPosition === "both") && pagination}
+                </div>
+            </div>
         </div>
     );
 }
+
+const SelectionCounter = observer(function SelectionCounter() {
+    const { selectionCountStore } = useDatagridRootScope();
+    if (selectionCountStore.displayCount) {
```
gjulivan, line 44: we can use https://github.com/mendix/web-widgets/blob/main/packages/shared/widget-plugin-component-kit/src/If.tsx

packages/pluggableWidgets/gallery-web/typings/GalleryProps.d.ts
```diff
@@ -97,4 +101,6 @@ export interface GalleryPreviewProps {
     emptyMessageTitle: string;
     ariaLabelListBox: string;
     ariaLabelItem: string;
+    sCountFmtSingular: string;
```
r0b1n, line 104: Maybe give them a bit more readable name that explains their purpose.

r0b1n (approved):

PR #1869, opened 2025-08-29
[Crafting] Pluggable Widgets MCP

rahmanunver:
It's a WIP.


automation/mendix-widgets-copilot/src/property-engine.ts
```diff
@@ -0,0 +1,508 @@
...
+            description: `Rename property key from '${oldKey}' to '${newKey}' in XML`
+        };
+    }
+
+    private async renamePropertyInTypeScript(
+        packagePath: string,
+        oldKey: string,
+        newKey: string
+    ): Promise<Omit<FileChange, "originalContent"> | null> {
+        const typingsPath = join(packagePath, "typings", `${this.getWidgetName(packagePath)}Props.d.ts`);
```
r0b1n, line 332: Typings are auto generated, so a safer way would be to run the generation using pwt when we add/remove properties.

automation/mendix-widgets-copilot/src/helpers.ts
```diff
@@ -0,0 +1,221 @@
...
+            inspection.packageXml = parser.parse(packageXmlContent);
+        } catch (err) {
+            inspection.errors.push(`Could not read package.xml: ${err}`);
+        }
+
+        // Find editor config
+        try {
+            const srcFiles = await readdir(srcPath);
+            for (const file of srcFiles) {
+                if (file.includes("editorConfig")) {
```
r0b1n, line 133: You can guess the name if you parse package json and get `widgetName` field + editorConfig + ".ts", same with preview files, widget xml, they all have the same pattern.

automation/mendix-widgets-copilot/src/property-engine.ts
```diff
@@ -0,0 +1,508 @@
...
+                operation: "update",
+                description: `Add property '${property.key}' usage to runtime component`
+            };
+        } catch (error) {
+            return null;
+        }
+    }
+
+    // Rename operations (similar structure but with find/replace logic)
+    private async renamePropertyInXml(
```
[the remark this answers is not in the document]
rahmanunver: I think we can modify and keep it as a guardrail, this can trigger a "Don't do it" logic we would like to provide. 
That guardrail can also be added into PRD, but doesn't an mcp level intervention sound safer? It is indeed a critical knowledge not to rename properties.

PR #1870, opened 2025-09-01
[WC-2552] Switch: Improve consistent toggling with label

yordan-st:
### Pull request type

---

### Description

- Added a visually hidden native <input type="checkbox"> to the Switch component for accessibility and label support.
- Ensured the Mendix-rendered label with for attribute toggles the switch as expected.

Related branch: https://github.com/mendix/atlas/pull/269

- If clicking the label toggles the Switch

### DIFF
--- a/packages/pluggableWidgets/switch-web/src/components/Switch.tsx
+++ b/packages/pluggableWidgets/switch-web/src/components/Switch.tsx
@@ -38,7 +38,7 @@ export default function Switch(props: SwitchProps): ReactElement {
                 role="switch"
                 aria-checked={props.isChecked}
                 aria-labelledby={`${props.id}-label`}
-                aria-readonly={!props.editable}
+                aria-disabled={!props.editable}
             >
                 <div
                     className={classNames("widget-switch-btn", {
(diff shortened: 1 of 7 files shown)


packages/pluggableWidgets/switch-web/src/components/Switch.tsx
```diff
@@ -14,8 +14,16 @@ export interface SwitchProps extends Pick<SwitchContainerProps, "id" | "tabIndex
 export default function Switch(props: SwitchProps): ReactElement {
     return (
         <div className="widget-switch">
-            <div
+            <input
+                type="checkbox"
                 id={props.id}
+                onClick={props.onClick}
+                style={{ opacity: "0", position: "absolute", width: "1px", height: "1px" }}
```
r0b1n, line 21: I am thinking if we can use readily available `sr-only` class to visually hide it instead of using inline styling.

gjulivan:
add changelog and increase minor version

r0b1n (dismissed):

gjulivan (approved):

PR #1872, opened 2025-09-01
Package xml improvements

r0b1n:


packages/pluggableWidgets/accessibility-helper-web/src/package.xml
```diff
@@ -5,7 +5,7 @@
             <widgetFile path="AccessibilityHelper.xml" />
         </widgetFiles>
         <files>
-            <file path="com/mendix/widget/web/accessibilityhelper" />
+            <file path="com/mendix/widget/web/accessibilityhelper/" />
```
iobuhov, line 8: I think we should avoid end slashes.

PR #1876, opened 2025-09-02
[WC-3032] Add logic to use barcodedetector API for barcode scanner

gjulivan:
- add logic to use BarcodeDetector API
- increase ideal width and height camera resolution for higher end devices

note: add Class "debug" on Properties -> Styling : Class in studio pro to check if BarcodeDetector API is being used.
if BarcodeDetector API is used, green border will be shown, otherwise, it is normal zxing/library.


samuelreichert (approved):
LGTM

PR #1882, opened 2025-09-08
[WC-3081] Combobox: Debounce datasource filter

yordan-st:
### Pull request type

---

### Description

This PR introduces a new property, allowing users to control the debounce interval (in ms) for datasource filter operations. This improves performance and user experience for large datasets and rapid input.

- Go to Advanced tab, set the debounce interval to 0 and test in runtime if it filters instantly
- Go to the Advanced tab, set the debounce interval to a high number e.g. 4000ms, and test in runtime if the filter is delayed.

### DIFF
--- a/packages/pluggableWidgets/combobox-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/combobox-web/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Changed
+
+- Moved the debounce interval for filtering operations from the Events tab to Advanced tab.
+
 ## [2.5.1] - 2025-09-19
 
 ### Fixed
--- a/packages/pluggableWidgets/combobox-web/src/Combobox.xml
+++ b/packages/pluggableWidgets/combobox-web/src/Combobox.xml
@@ -336,10 +336,6 @@
                     <actionVariable key="filterInput" caption="Filter Input" type="String" />
                 </actionVariables>
             </property>
-            <property key="filterInputDebounceInterval" type="integer" required="true" defaultValue="200">
-                <caption>Debounce interval</caption>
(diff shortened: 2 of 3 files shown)


packages/pluggableWidgets/combobox-web/src/__tests__/SingleSelection.spec.tsx
```diff
@@ -80,7 +80,8 @@ describe("Combo box (Association)", () => {
             selectedItemsSorting: "none",
             customEditability: "default",
             customEditabilityExpression: dynamic(false),
-            filterInputDebounceInterval: 200
+            filterInputDebounceInterval: 200,
+            datasourceFilterDebounceInterval: 200
```
r0b1n, line 84: Can't we reuse already existing value? Arguably they should always be the same, like applying filters and calling onFilterChange action should be hand in hand.

packages/pluggableWidgets/combobox-web/src/helpers/BaseDatasourceOptionsProvider.ts
```diff
@@ -50,11 +68,14 @@ export class BaseDatasourceOptionsProvider extends BaseOptionsProvider<ObjectIte
 
     getAll(): string[] {
         if (this.lazyLoading && this.attributeId) {
+            if (!this.debouncedSetFilter) {
+                this.createDebouncedSetFilter();
+            }
```
r0b1n, line 73: Why do we need to recreate it every time `getAll` is called?

packages/pluggableWidgets/combobox-web/src/__tests__/SingleSelection.spec.tsx
yordan-st: Okay but I will detach it from the 'On filter input change' action so it's always visible instead of only when there is an action configured.

leonardomendix (approved):

PR #1899, opened 2025-09-16
[WC-3050] allow self hosted pdfjs worker file

gjulivan:
### Pull request type

---

### DIFF
--- a/packages/pluggableWidgets/document-viewer-web/components/PDFViewer.tsx
+++ b/packages/pluggableWidgets/document-viewer-web/components/PDFViewer.tsx
@@ -6,7 +6,7 @@ import { downloadFile } from "../utils/helpers";
 import { useZoomScale } from "../utils/useZoomScale";
 import BaseViewer from "./BaseViewer";
 import { DocRendererElement, DocumentRendererProps, DocumentStatus } from "./documentRenderer";
-
+import { If } from "@mendix/widget-plugin-component-kit/If";
 const options = {
     cMapUrl: "/widgets/com/mendix/shared/pdfjs/cmaps/",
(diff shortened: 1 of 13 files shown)


packages/pluggableWidgets/document-viewer-web/components/PDFViewer.tsx
```diff
@@ -7,18 +7,20 @@ import { useZoomScale } from "../utils/useZoomScale";
...
 const PDFViewer: DocRendererElement = (props: DocumentRendererProps) => {
-    const { file, setDocumentStatus } = props;
+    const { file, setDocumentStatus, pdfjsWorkerUrl } = props;
+    pdfjs.GlobalWorkerOptions.workerSrc = useMemo(() => {
+        if (pdfjsWorkerUrl?.status === "available" && pdfjsWorkerUrl?.value) {
+            return pdfjsWorkerUrl.value;
+        }
+        return `//unpkg.com/pdfjs-dist@${pdfjs.version}/build/pdf.worker.min.mjs`;
+    }, [pdfjsWorkerUrl]);
+
```
r0b1n, line 23: To be on the safe side, we probably should delay the rendering until this value if correctly resolved, we don't what to process with loading from cdn while it is being loade. Also if this url is unavailable we probably should not render at all. Unexpected loading from CDN can be seen as security issues.
So I guess we should load from cdn only if the value is available and is explicitly empty string. Other than that - wait for it. If value goes to unavailable - show an error message.

leonardomendix (dismissed):

r0b1n (approved):

leonardomendix (approved):

PR #1901, opened 2025-09-17
[WC-3096] Gallery: fix divider gap spacing

yordan-st:
### Pull request type

---

### Description

Fixed an issue where setting the gallery gap to 0 would cause an offset, which made the bottom border of items to dissapear.

Set the gap to 0 and check if gallery items are spaced evenly and all border are visible

### DIFF
--- a/packages/modules/data-widgets/src/themesource/datawidgets/web/_gallery-design-properties.scss
+++ b/packages/modules/data-widgets/src/themesource/datawidgets/web/_gallery-design-properties.scss
@@ -43,7 +43,7 @@
             left: 0;
             right: 0;
             border-bottom: 1px solid var(--grid-border-color, $grid-border-color);
-            margin-top: calc(var(--gallery-gap) / 2 - 1px);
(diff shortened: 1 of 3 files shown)


packages/modules/data-widgets/src/themesource/datawidgets/web/_gallery-design-properties.scss
```diff
@@ -43,7 +43,7 @@
             left: 0;
             right: 0;
             border-bottom: 1px solid var(--grid-border-color, $grid-border-color);
-            margin-top: calc(var(--spacing-small, $spacing-small) / 2);
+            margin-top: calc(var(--gallery-gap) / 2 - 1px);
```
gjulivan, line 46: `var(--gallery-gap, var(--spacing-small, $spacing-small))`

put a fallback value.

samuelreichert (approved):

leonardomendix (approved):

PR #1906, opened 2025-09-19
[continued as #1953]

rahmanunver:
### Pull request type

---

### Description
This PR implements multi-page select all functionality for the Datagrid2 widget, allowing users to select all items across multiple pages based on the current filter. The feature includes configurable batch processing, progress tracking, and cancellation support, similar to the existing Excel export functionality.

Small dataset (< buffer size): Should use immediate selection
Large dataset (>= buffer size): Should show progress dialog and batch processing
Unknown total count: Should fall back to per-page selection
Cancellation: Should properly cancel and restore state
Page restoration: Should return user to original page after selection
Filtered data: Should only select items matching current filter


packages/pluggableWidgets/datagrid-web/src/components/loader/SpinnerLoader.tsx
```diff
@@ -4,12 +4,20 @@ import { createElement, ReactElement } from "react";
 type SpinnerLoaderProps = {
     size?: "small" | "medium" | "large";
     withMargins?: boolean;
+    fullWidth?: boolean;
```
r0b1n: Looks like we never use this prop anywhere

packages/pluggableWidgets/datagrid-web/src/features/data-export/DSExportRequest.ts
```diff
@@ -230,6 +231,9 @@ export class DSExportRequest {
         this.emitEnd();
         this.emitLoadEnd();
         this.dispose();
+        performance.mark("DSExportRequest_end");
+        const measure = performance.measure("DSExportRequest", "DSExportRequest_send", "DSExportRequest_end");
+        console.debug(`DSExportRequest: export took ${(measure.duration / 1000).toFixed(2)} seconds`);
```
r0b1n: Some leftovers

packages/pluggableWidgets/gallery-web/src/Gallery.xml
```diff
@@ -198,6 +198,13 @@
                     <caption>Item count plural</caption>
                     <description>Must include '%d' to denote number position ('%d items selected')</description>
                 </property>
+                <property key="clearSelectionCaption" type="textTemplate">
+                    <caption>Clear selection caption</caption>
+                    <description />
+                    <translations>
+                        <translation lang="en_US">Clear selection</translation>
+                    </translations>
+                </property>
```
r0b1n: Gallery probably needs a changlog entry about this change

packages/shared/widget-plugin-grid/src/select-all/SelectAllController.ts
```diff
@@ -0,0 +1,184 @@
...
+
+            // const selectionBeforeReload = this.selection?.selection ?? [];
+            // Reload selection to make sure setSelection is working as expected.
+            this.selection?.setSelection(success ? allItems : initSelection);
+            this.locked = false;
+            this.#abortController = undefined;
+
+            performance.mark("SelectAll_End");
+            const measure1 = performance.measure("Measure1", "SelectAll_Start", "SelectAll_End");
+            console.debug(`Data grid 2: 'select all' took ${(measure1.duration / 1000).toFixed(2)} seconds.`);
```
r0b1n: Do we want to keep this in production code? Likely useless for production, but if we want to keep this for dev build mayse there is a way to strip this away via rollup for production.

iobuhov: I always see debug info for nanoflows. So I think it's fine to keep this one as well

PR #1912, opened 2025-09-23
[WC-3097] Image: Fix icon scaling and alignment

yordan-st:
### Pull request type

---

### Description

Fixed an issue where selecting an icon as data source would cause layout, icon dimensions and alignment to break.

- Select icon as data source,
- Observe dimension changes
- Only font size should change icon dimensions
- Use alignment design properties

### DIFF
--- a/packages/pluggableWidgets/image-web/src/Image.editorConfig.ts
+++ b/packages/pluggableWidgets/image-web/src/Image.editorConfig.ts
@@ -91,7 +91,14 @@ export function getProperties(
     }
 
     if (values.datasource === "icon" && (values.imageIcon?.type === "glyph" || values.imageIcon?.type === "icon")) {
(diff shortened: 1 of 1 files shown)


packages/pluggableWidgets/image-web/src/components/Image/ui.tsx
```diff
@@ -67,7 +69,7 @@ function ContentIcon(props: ImageContentIcon): ReactElement {
     return (
         <span
             className={classNames(props.icon, { glyphicon: props.isGlyph })}
-            style={{ ...props.style, fontSize: `${props.size}px` }}
```
gjulivan, line 70: this is removing style props being set from studio pro. we can't do this.

image

gjulivan (approved):

leonardomendix (approved):

PR #1928, opened 2025-10-07
[WC-3139] Change custom chart event data to return more properties

leonardomendix:
### Pull request type

---

### Description

Change custom chart event data to return more properties.

### DIFF
--- a/packages/pluggableWidgets/custom-chart-web/src/controllers/ChartPropsController.ts
+++ b/packages/pluggableWidgets/custom-chart-web/src/controllers/ChartPropsController.ts
@@ -126,19 +126,31 @@ export class ChartPropsController implements ReactiveController {
         return (data: any): void => {
             if (this.props.eventDataAttribute && data.points && data.points.length > 0) {
                 const point = data.points[0];
-                const eventData = {
-                    curveNumber: point.curveNumber,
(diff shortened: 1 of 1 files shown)


packages/pluggableWidgets/custom-chart-web/src/controllers/ChartPropsController.ts
```diff
@@ -124,8 +124,21 @@ export class ChartPropsController implements ReactiveController {
 
     private get chartOnClick(): (data: any) => void {
         return (data: any): void => {
-            if (this.props.eventDataAttribute) {
-                this.props.eventDataAttribute.setValue(JSON.stringify(data.points[0].bbox));
+            if (this.props.eventDataAttribute && data.points && data.points.length > 0) {
+                const point = data.points[0];
+                const eventData = {
```
gjulivan, line 129: need to have null pointer check here.
`if(point)`

then do destructure
`const {curveNumber, pointNumber, pointIndex,x , y, z, text, hovertext,customdata, bbox} = point;`

leonardomendix: Thanks, good one! I did that and added extra properties for the maps chart and others. ✅

PR #1941, opened 2025-10-15
[WC-2826]: Export to excel types and formats

samuelreichert:
### Pull request type

---

### Description

Adding possibility to set type and format (masks) to export to excel in datagrid2.

### DIFF
--- a/packages/pluggableWidgets/datagrid-web/src/features/data-export/DSExportRequest.ts
+++ b/packages/pluggableWidgets/datagrid-web/src/features/data-export/DSExportRequest.ts
@@ -304,7 +304,7 @@ const readers: ReadersByType = {
                     exportNumberFormat: props.exportNumberFormat
                 });
 
-                return excelStringFormat(data.value ?? "", format);
+                return excelString(data.value ?? "", format);
             case "unavailable":
                 return excelString("n/a");
             default:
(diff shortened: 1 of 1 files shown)


packages/pluggableWidgets/datagrid-web/src/features/data-export/DSExportRequest.ts
```diff
@@ -253,48 +265,107 @@ export class DSExportRequest {
...
 
-        if (data.value instanceof Big) {
-            return data.value.toNumber();
+        if (value instanceof Big || typeof value === "number") {
+            const num = value instanceof Big ? value.toNumber() : value;
+            return {
+                t: "n",
+                v: num,
+                z: format
+            };
```
r0b1n: Can we make some helpers out of those returns, so we have the complexity away and use something like:

return excelNumber(num, format);
return excelBoolean(value);
return excelString("");
return excelDate(data, format);

packages/pluggableWidgets/datagrid-web/src/features/data-export/DSExportRequest.ts
```diff
@@ -252,49 +264,126 @@ export class DSExportRequest {
...
+    };
+}
+
+function excelStringFormat(value: string, format?: string): ExcelCell {
+    return {
+        t: "s",
+        v: value,
+        z: format
+    };
+}
```
r0b1n: It can be one method, I think

r0b1n (approved):

leonardomendix (approved):

PR #1950, opened 2025-10-23
[WC-3158] Maps: mitigate not rendering

r0b1n:
### Pull request type

---

### Description

Move styling to inline so they size is immediately available and Leaflet is not distorted 🙈 
This should make Maps render correctly without distortion. As the issue is likely CSS not being loaded at the moment the size calculation happens.


iobuhov (approved):
Fixing stuff with delays... LGTM

leonardomendix (dismissed):

leonardomendix (dismissed):

gjulivan (dismissed):

PR #1964, opened 2025-11-04
[WC-3157] Add event handlers for file uploader

r0b1n:
### Pull request type

### Description

We added possibility to configure actions that will be executed on succeeded or failed file upload. This gives a possibility to handle some corner cases that may occur, so example clean up hanging object if upload was not successful. Object in question is being passed to those actions.

Configured actions work as expected for respective events, and also they are getting the object passed.


iobuhov (approved):
LGTM

PR #1965, opened 2025-11-05
[WC-3162] DG: Add auto-select first row

yordan-st:
### Pull request type

---

### Description

- We added a feature where developers can configure the first item of a DataGrid to be automatically selected when it first loads.
- This feature is intended for master-detail views, where the user wants the dataview not to be empty on first load.

- Configure 'auto select first row'
- Observe if the first item is selected
- Check if this selection persists when you move to another page of the DataGrid

### DIFF
--- a/packages/shared/widget-plugin-grid/src/selection/createSelectionHelper.ts
+++ b/packages/shared/widget-plugin-grid/src/selection/createSelectionHelper.ts
@@ -58,7 +58,7 @@ export function createSelectionHelper(
             );
             add(cleanup);
         }
-        if (helper.type !== null && config.autoSelect) {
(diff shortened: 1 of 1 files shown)


packages/pluggableWidgets/datagrid-web/src/Datagrid.xml
```diff
@@ -37,6 +37,10 @@
                         <enumerationValue key="rowClick">Row click</enumerationValue>
                     </enumerationValues>
                 </property>
+                <property key="selectFirstRow" type="boolean" defaultValue="false">
```
iobuhov: let's rename this to `autoSelect`

PR #1966, opened 2025-11-06
fix: resolving react unique key prop warning error

DiljohnSingh:
### Pull request type

Screenshot 2025-11-06 at 11 36 16

### Description

- Added index as fallback key to avoid react warnings in case of non repeated element.
- Fixed the typo in test cases
- Updated the Changelog
- Updated package.xml and package.json with new version 1.2.3

- Try to use the current version of HTML element widget on page as non repeated element.
- Check in console both mendix studio pro and browser to see React warning.
- Use the widget in your project after building the widget from branch, warning should not be removed.

### DIFF
--- a/packages/pluggableWidgets/html-element-web/src/components/__tests__/HTMLTag.spec.tsx
+++ b/packages/pluggableWidgets/html-element-web/src/components/__tests__/HTMLTag.spec.tsx
@@ -37,7 +37,7 @@ describe("HTMLTag", () => {
     });
 
     it("with innerHTML apply html sanitizing", () => {
-        const checkSapshot = (html: string): void => {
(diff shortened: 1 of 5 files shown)


packages/pluggableWidgets/html-element-web/src/components/__tests__/HTMLTag.spec.tsx
```diff
@@ -37,7 +37,7 @@ describe("HTMLTag", () => {
     });
 
     it("with innerHTML apply html sanitizing", () => {
-        const checkSapshot = (html: string): void => {
+        const checkSnapshot = (html: string): void => {
```
DiljohnSingh: Noticed the typo

PR #1968, opened 2025-11-07
fix: html element id on single element

Andries-Smit:
HTML Element widget cause error in the log
>Warning: Each child in a list should have a unique "key" prop.%s%s See https://reactjs.org/link/warning-keys for more information.%s
>    at HTMLTag (http://naviva.local:8080/dist/HTMLElement-JMsVPxAr.js:1503:40)

When not using a tagUseRepeat the there is no key set, as `item?.id ` will be empty
This will cause an issues when the the HTML Element widget is placed multiple times in the same parent.

Solution set `key` to the name of the widget.


gjulivan:
related https://github.com/mendix/web-widgets/pull/1966/files

PR #1969, opened 2025-11-07
[WC-2942]: Fix carousel focus indicator

HedwigAR:
### Pull request type

---

### Description

Added focus indicators to the navigation buttons in the carousel, disabled autoplay by default since it's not accessible. 

- Autoplay should be disabled in settings by default, 
- When tabbing to the navigation buttons on the carousel (the left and right arrows) a focus indicator should appear. The focus indicator should be blue and with a 2px offset, to match our other focus indicators. 
- The navigation buttons should also have a default white semi transparent circular background to make them stand out more.

### DIFF
--- a/packages/pluggableWidgets/carousel-web/src/ui/Carousel.scss
+++ b/packages/pluggableWidgets/carousel-web/src/ui/Carousel.scss
@@ -1,10 +1,12 @@
 // Atlas UI compatible variable naming for data widgets
 $gray-200: #e3e3e5 !default;
-$spacing-large: 24px !default;
+$spacing-small: 8px !default;
 $focus-outline: 2px solid #264ae5 !default;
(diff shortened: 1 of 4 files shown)


packages/pluggableWidgets/carousel-web/src/ui/Carousel.scss
```diff
@@ -23,6 +23,20 @@
...
+.swiper-button-prev {
+    --swiper-navigation-size: 24px;
+    border-radius: 50%;
+    background-color: rgba(245, 245, 245, 0.7);
+    padding: 2em;
+}
+
+.swiper-button-next:focus,
+.swiper-button-prev:focus {
+    outline: 2px solid #264ae5;
```
gjulivan: is there a css variables for this?
i think it should be `var(--focus-outline)`

PR #1987, opened 2025-11-19
Update Export_To_Excel.js

kylekunzer-kt:
### Pull request type

---

### Description

Fixes an issue where Date attributes become Strings during the export and therefore do not behave as Dates in the resultant Excel file and cannot have Date formatting applied to them.

Export from a DataGrid 2 widget that has at least one column where the attribute is a DateTime attribute. Test that dates in the exported file can be formatted (e.g. short date to long date, or change formatting to yyyy-MM-dd, MM/dd/yyyy, etc.) and that the change is applied.

Prior to this change, exported columns that are a datetime attribute should not change when formatting is applied because they are not true dates in the file.


gjulivan:
https://github.com/mendix/web-widgets/pull/1941/files

in this PR (unreleased), we are already implemented export types format

PR #1992, opened 2025-11-25
[WC-3186]: Combobox(Database) onChange event firing unexpectedly

rahmanunver:
### Pull request type

---

### Description
Changed onChange binding to the Database selector.
Should resolve the issue with onChange firing on initial mount.

Values getting properly set on initial run.

### DIFF
--- a/packages/pluggableWidgets/combobox-web/src/Combobox.xml
+++ b/packages/pluggableWidgets/combobox-web/src/Combobox.xml
@@ -50,7 +50,7 @@
                     <caption>Selectable objects</caption>
                     <description />
                 </property>
-                <property key="optionsSourceDatabaseItemSelection" type="selection" dataSource="optionsSourceDatabaseDataSource" onChange="onChangeDatabaseEvent">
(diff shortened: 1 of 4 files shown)


packages/pluggableWidgets/combobox-web/src/Combobox.xml
```diff
@@ -50,7 +50,7 @@
                     <caption>Selectable objects</caption>
                     <description />
                 </property>
-                <property key="optionsSourceDatabaseItemSelection" type="selection" dataSource="optionsSourceDatabaseDataSource" onChange="onChangeDatabaseEvent">
+                <property key="optionsSourceDatabaseItemSelection" type="selection" dataSource="optionsSourceDatabaseDataSource">
```
gjulivan: the reason why we add `onChangeDatabaseEvent` in addition to normal `onChangeEvent` event is because the on-change was handled by framework due to selection api..
and if we remove this, means that we are handling the onchange ourself. (again).
and if we do that, why do we still keep the `onChangeDatabaseEvent` separate from the `onChangeEvent`?
why not just use `onChangeEvent`?

PR #2122, opened 2026-03-03
[CRAFTING] Use Vite to build widgets

samuelreichert:
### Pull request type

---

### Description

Adding vite and scripts on all widgets for building with vite, and generating the MPK using vite.
Also adding a benchmark script to compare current build and vite build (Vite + Rollup)


packages/shared/vite-web-widgets/vite.config.ts
iobuhov: Which model created this file?

iobuhov:
Did you tried to rune `dev`? I'm curious as it should work, but I'm not sure how good. We might need spend some time on merging this code into widget tools + we need time to build "runtime" code to mock PW API. This way we have "sandbox" where widget can be fully rendered and tested.

PR #2125, opened 2026-03-04
fix: letters OL are converted to UL

MikeGogol:
Hello, 

This PR is created according to ticket number 273055.

Please note that solution which we proposed is based on [this resource](https://github.com/attoae/quill-table-better/pull/146).

Looks like it's known issue in the library which you guys using. 

Please note that we didn't perform full regression tests on the widget.


gjulivan (approved):

gjulivan:
fixed in https://github.com/mendix/web-widgets/pull/2143/changes/36bcd980e135f6d82719236ba0afc987d32913ad

PR #2136, opened 2026-03-11
fix(combobox): correct filtering when only one character is typed

msoaresms:
Issue: filtering did not work correctly when only one character was entered.

Root cause: startsWith search option was using WORD_STARTS_WITH.

Solution: replaced WORD_STARTS_WITH with STARTS_WITH.

### Pull request type

---


iobuhov (approved):

gjulivan:
https://mendix.atlassian.net/browse/WC-3456

yordan-st (approved):

leonardomendix:
Thanks for the PR. However, we will close this one for now, as the behaviour tested is effectively the same: with one-character input and single-word options, both versions produce the same results.

The threshold change (from WORD_STARTS_WITH to STARTS_WITH) only affects multi-word options, which weren't included in this validation.

PR #2161, opened 2026-04-03
[WC-3259] DG2: fix focus jump on filter controls

yordan-st:
### Pull request type

---

### Description

Selecting Empty or Not empty in a DG2 text or number filter caused keyboard focus to jump to the top of the page, forcing keyboard users to re-orient after every operator change. This happened because the filter tried to focus a text input that was disabled in those cases. The fix prevents that focus attempt, so focus stays on the operator dropdown where the user already is.

- Verify that selecting Empty and Not empty in both text and number filters does not move focus away from the filter controls.
- Confirm keyboard-only flow works smoothly: open operator dropdown, choose operator, continue tabbing without losing place.
- Check that the value input is disabled for Empty/Not empty and enabled again for operators that require a value.
- Validate behavior in both directions: switch from value-based operator to Empty/Not empty and back again.


iobuhov (approved):
LGTM 👏

leonardomendix (approved):

PR #2162, opened 2026-04-03
[WC-3394] virtual scrolling sticky, make it stick again

r0b1n:
### Pull request type

---


packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid.scss
```diff
@@ -681,3 +633,7 @@ $root: ".widget-datagrid";
         }
     }
 }
+
+[data-overlay-content] {
```
iobuhov: We need prefix this selector with $root so it not "global"

iobuhov (commented):
So we just revert back to "all" in one css grid mode?

r0b1n:
Yes, pretty much, standard grid and sticky header. No subgrid, no JS magic.

packages/modules/data-widgets/src/themesource/datawidgets/web/_datagrid.scss
r0b1n: ✅

leonardomendix (approved):

gjulivan (approved):

PR #2166, opened 2026-04-08
[WC-3163] Treenode v2

gjulivan:
### Pull request type

---


packages/pluggableWidgets/tree-node-web/src/components/v2/hooks/useIncrementalTreeData.ts
iobuhov, line 1: This file needs bit of refactoring.
- We need extract all logic (useEffect) to function, so we can test it.
- We should create new type for tree, something like TreeState that should replace multiple useRefs
- Line 26 -> let's use type literlas

Also, how we prevent recursion?

packages/pluggableWidgets/tree-node-web/src/components/v2/hooks/useInfiniteTreeNode.ts
iobuhov, line 1: Same here, we need extract logic to function so we can write unit tests

packages/pluggableWidgets/tree-node-web/src/components/v2/TreeNode.tsx
```diff
@@ -0,0 +1,120 @@
...
+        }),
+        [props.headerCaption, props.headerContent, props.headerType, props.parentAssociation, props.startExpanded]
+    );
+
+    const treeData = useIncrementalTreeData(items, treeConfig);
+    const onNodeClick = useCallback(
+        (node: TreeNodeV2DataItem) => {
+            if (node.treeNodeState === TreeNodeState.EXPANDED) {
+                node.treeNodeState = TreeNodeState.COLLAPSED_WITH_CSS;
+                forceRender(version => version + 1);
```
iobuhov, line 127: forceRender uses `useReducer` with default n => n + 1 reducer. This removes need to always pass function to forceRender

iobuhov (dismissed):

packages/pluggableWidgets/tree-node-web/src/components/v2/hooks/useIncrementalTreeData.ts
gjulivan: @iobuhov this whole widgets idea is recursive. what do you mean by prevent recursion?

iobuhov: I was about infinite recursion, lol

iobuhov (dismissed):
Approve, but not all changes requested have been implemented.

iobuhov (approved):

r0b1n (approved):

PR #2167, opened 2026-04-08
[WC-3333]- DG2 virtual scroll loses scrollbar after column hide

rahmanunver:
### Pull request type

Bug fix

---

### Description

**What:** `lockGridBodyHeight()` in `GridSizeStore` locked the virtual scroll body height once and only reset it when the page size changed. When a column is hidden, row heights can decrease (text no longer wraps), so the content height drops below the locked `max-height` — the scrollbar disappears.

**Why:** The reset condition was incomplete. The existing `lockedAtPageSize` guard pattern was correct but not applied to column-count changes. The fix adds a parallel `lockedAtColumnCount` guard: the locked height is now invalidated whenever the visible column count changes, mirroring the existing page-size reset exactly.

**Files changed:**
- `src/model/stores/GridSize.store.ts` — added `lockedAtColumnCount` guard
- `src/model/containers/Datagrid.container.ts` — wired `visibleColumnsCountAtom` into `GridSizeStore` via DI
- `src/components/Pagination.tsx`, `src/components/WidgetFooter.tsx` — pre-existing TS7006 implicit-any fixes (were already failing on `main`)
- `e2e/DataGrid.spec.js` — regression test: virtual scroll + column hide
- `AGENTS.md` — architecture reference for the datagrid-web package

1. Open a page with a Data Grid configured for **virtual scrolling** and enough rows to produce a scrollbar
2. Confirm the scrollbar is visible in the grid body
3. Open the column selector and hide any column
4. Confirm the scrollbar is **still visible** after hiding — this was the bug
5. Show the column again and confirm the scrollbar remains consistent


packages/pluggableWidgets/datagrid-web/AGENTS.md
```diff
@@ -0,0 +1,196 @@
+# Datagrid Web (`@mendix/datagrid-web`)
+
+Data Grid 2 — the primary data table widget for Mendix web apps. Sorting, filtering,
+pagination (buttons/virtual scroll/load more), column resize/reorder/hide, row selection
+(single/multi, checkbox/click), data export, and personalization (localStorage/attribute).
+
+## Commands
+
+- Test: `cd packages/pluggableWidgets/datagrid-web && pnpm run test`
+- Build: `pnpm --filter @mendix/datagrid-web run build`
```
r0b1n: Probably `pnpm turbo build` from the widget directory is better.

packages/pluggableWidgets/datagrid-web/AGENTS.md
```diff
@@ -0,0 +1,196 @@
...
+- Lint: `cd packages/pluggableWidgets/datagrid-web && pnpm run lint`
+- E2E: `cd packages/pluggableWidgets/datagrid-web && pnpm run e2e`
+- Dev: set `MX_PROJECT_PATH`, then `pnpm run start` inside package dir
+
+**Shared package changes:** When modifying a workspace dependency (e.g., `widget-plugin-grid`,
+`widget-plugin-mobx-kit`), you must build the shared package first, then rebuild datagrid-web:
```
r0b1n: is used `pnpm turbo build` this is not needed.

PR #2170, opened 2026-04-13
[WC-3336] Barcode generator

gjulivan:
-  generate file name on downclick instead of on initialize to keep creating different file names.

### DIFF
--- a/packages/pluggableWidgets/barcode-generator-web/src/utils/download-utils.ts
+++ b/packages/pluggableWidgets/barcode-generator-web/src/utils/download-utils.ts
@@ -5,6 +5,49 @@ const NAMESPACES = {
     XLINK: "http://www.w3.org/1999/xlink"
 } as const;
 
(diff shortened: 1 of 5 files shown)


packages/pluggableWidgets/barcode-generator-web/src/utils/download-utils.ts
```diff
@@ -5,6 +5,49 @@ const NAMESPACES = {
...
+    const hash = hashCode(codeValue);
+    return `${prefix}_${hash}_${timestamp}.png`;
+}
+
+function generateTimestamp(): string {
+    // Get current date and time
+    const now = new Date();
+
+    // Format: YYYYMMDD_HHMMSS
+    const timestamp =
```
iobuhov, line 20: Can we use literal template?

iobuhov (approved):

leonardomendix (approved):

PR #2173, opened 2026-04-13
[WC-3355]: fix combobox not updating displayed value on external attribute change

rahmanunver:
### Pull request type

---

### Description

In database mode, the combobox displayed a stale value when the target attribute was changed externally (e.g. by a popup combobox writing to the same attribute).

**Root cause:** `DatabaseSingleSelectionSelector.updateProps()` only synced `currentId` on initial load (`!this.currentId`). When the attribute changed from value A to value B while `currentId` was already set to A, the condition was false so `currentId` stayed at A and the display never updated.

**Fix:** Added a second condition — also enter the sync branch when the current option's stored value no longer matches the incoming `targetAttribute.value`, using the existing `_valuesIsEqual` helper for type-safe comparison. A matching value short-circuits the scan, so there is no performance impact on the common (no-change) case.

**Files changed:**
- `src/helpers/Database/DatabaseSingleSelectionSelector.ts` — core fix (one condition added)
- `CHANGELOG.md` — `### Fixed` entry under `## [Unreleased]`
- `package.json` + `src/package.xml` — version bump 2.8.0 → 2.8.1
- `AGENTS.md` — architecture reference for future debugging sessions

1. Open a page with two comboboxes in database mode bound to the same attribute on the same object.
2. Select a value in combobox A (e.g. "Option 1").
3. Open a popup that contains combobox B; select a different value (e.g. "Option 2") and close the popup.
4. **Expected:** Combobox A now displays "Option 2".
5. **Before fix:** Combobox A still shows "Option 1" (stale).
6. Verify the reverse direction: change in A after changing via B also works correctly.

### DIFF
--- a/packages/pluggableWidgets/combobox-web/src/package.xml
+++ b/packages/pluggableWidgets/combobox-web/src/package.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <package xmlns="http://www.mendix.com/package/1.0/">
-    <clientModule name="Combobox" version="2.8.0" xmlns="http://www.mendix.com/clientModule/1.0/">
+    <clientModule name="Combobox" version="2.8.1" xmlns="http://www.mendix.com/clientModule/1.0/">
         <widgetFiles>
             <widgetFile path="Combobox.xml" />
         </widgetFiles>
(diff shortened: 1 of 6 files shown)


packages/pluggableWidgets/combobox-web/src/package.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <package xmlns="http://www.mendix.com/package/1.0/">
-    <clientModule name="Combobox" version="2.8.0" xmlns="http://www.mendix.com/clientModule/1.0/">
+    <clientModule name="Combobox" version="2.8.1" xmlns="http://www.mendix.com/clientModule/1.0/">
```
r0b1n: We bump versions in release preparation step nowadays

r0b1n (dismissed):

gjulivan (approved):

leonardomendix (approved):

PR #2182, opened 2026-04-20
[WC-3334] Fix DG2 export type bugs

samuelreichert:
### Pull request type

### Description

Fix multiple data export bugs in DataGrid2's Excel export feature (WC-3334). 

### Problems                                                                                                                              
           
1. **customContent columns ignore exportType** — exportValue always exports as text regardless of exportType setting, making numeric values unusable for calculations in Excel
2. **Date exports leak time components** — Date-only formats (e.g., dd-MMM-yyyy) carry a hidden time component in the cell value, affecting sorting and calculations
3. **Boolean exports use string representation** — Boolean values exported as text strings instead of native Excel boolean cells (`t: "b"`)
4. **Long numbers lose precision** — Big.toNumber() loses precision for values >15 significant digits, causing scientific notation in Excel

### Changes

- Extracted cell-creation logic from DSExportRequest.ts into standalone cell-readers.ts module (pure functions, zero behavior change)
- customContent reader now branches on exportType: parses to `Number()` for "number", `new Date()` for "date", string boolean matching for "boolean", with graceful string fallback on parse failure
- Added `hasTimeComponent(format)` / `stripTime(date)` helpers — when format has no h/s tokens, date values are normalized to midnight UTC
- `excelDate` now always produces `t: "d"` cells with a format code; when no explicit format is configured, reads the app's locale date pattern from `window.mx.session.getConfig().locale.patterns.date` and converts it to an Excel-compatible format
- `excelBoolean` now returns `{ t: "b", v: boolean }` — native Excel boolean cells (TRUE/FALSE)
- customContent boolean parsing accepts case-insensitive: true/false, yes/no, 1/0
- Big values with >15 significant digits export as strings via `toFixed()` to preserve all digits
- Added `typings/global.d.ts` for `window.mx` type declarations
- 35 unit tests covering all reader paths, edge cases, and fallback behavior


packages/pluggableWidgets/datagrid-web/src/features/data-export/__tests__/cell-readers.spec.ts
```diff
@@ -0,0 +1,338 @@
...
+        });
+
+        it("exports date attribute without format as string cell (displayValue)", () => {
+            const col = column("Created", c => {
+                c.showContentAs = "attribute";
+                c.attribute = listAttribute(() => new Date("2024-06-15T10:30:00Z"));
+                c.exportType = "default";
+            });
+            const cell = readSingleCell(col);
+            expect(cell.t).toBe("s");
```
iobuhov: value test is missing

packages/pluggableWidgets/datagrid-web/src/features/data-export/cell-readers.ts
```diff
@@ -0,0 +1,200 @@
...
+    /** Cell type: 's' = string, 'n' = number, 'd' = date */
+    t: "s" | "n" | "d";
+    /** Underlying value */
+    v: string | number | Date;
+    /** Optional Excel number/date format, e.g. "yyyy-mm-dd" or "$0.00" */
+    z?: string;
+    /** Optional pre-formatted display text */
+    w?: string;
+}
+
```
iobuhov, line 16: Can we use "proper" names? (e.g. type, value, format?

packages/pluggableWidgets/datagrid-web/src/features/data-export/cell-readers.ts
```diff
@@ -0,0 +1,200 @@
...
+export function excelBoolean(value: boolean): ExcelCell {
+    return {
+        t: "s",
+        v: value ? "Yes" : "No"
+    };
+}
+
+function hasTimeComponent(format: string): boolean {
+    return /[hs]/i.test(format);
```
iobuhov: Which format schema we use in number/date cases? Is there ISO spec or something?

packages/pluggableWidgets/datagrid-web/src/features/data-export/cell-readers.ts
samuelreichert: Not if we want to keep it SheetJS compatible: https://docs.sheetjs.com/docs/csf/cell

packages/pluggableWidgets/datagrid-web/src/features/data-export/cell-readers.ts
samuelreichert: The `z` property follows ECMA-376 / OOXML number format strings, the same syntax Excel uses natively.
SheetJS reads them directly and applies them in the workbook.

Dates: `yyyy`, `mm`, `dd`, `mmm`, `h`, `s` tokens
e.g. "yyyy-mm-dd", "dd-mmm-yyyy hh:mm:ss"

Numbers: `#`, `0`, `,`, `.` tokens
e.g. "#,##0.00", "$0.00"

There's no ISO spec here; it's the Excel format code syntax defined in [ECMA-376 Part 1, §18.8.31](https://ecma-international.org/publications-and-standards/standards/ecma-376/).

The `hasTimeComponent` regex `/[hs]/i` keys off `h` (hours) and `s` (seconds) from that same token set to decide whether to strip the time part before writing the date cell.

iobuhov (dismissed):

rahmanunver (dismissed):

iobuhov (dismissed):
LGTM

leonardomendix (dismissed):

leonardomendix (dismissed):

rahmanunver (dismissed):

iobuhov (approved):

gjulivan (approved):

PR #2193, opened 2026-04-28
[WC-3390] DG2: improve virtual scroll behavior and height logic

yordan-st:
### Pull request type

---

### Description

- Virtual scrolling got stuck showing only the initial page of rows when the grid had many columns. The overflow detection compared measurements from different sources, incorrectly skipping the offset needed to make the body scrollable. Fixed by using consistent measurements for the overflow check and adding a fallback that loads more rows when the grid has no scrollbar but more data is available.


packages/pluggableWidgets/datagrid-web/src/model/hooks/useInfiniteControl.tsx
```diff
@@ -42,7 +42,22 @@ export function useInfiniteControl(): [trackBodyScrolling: ((e: any) => void) |
     );
 
     useEffect(() => {
```
iobuhov, line 44: This use effect runs on every rerender. That's by design?

iobuhov: Maybe this should be fixed in GridSizeStore?

yordan-st: The no-dep-array was pre-existing. But I moved new logic to GridSizeStore

r0b1n:
It feels this has to be consolidated with https://github.com/mendix/web-widgets/pull/2167/ and addressed together, while those might be different issues, they are caused by the came code which needs an overhaul/thinking though session.

PR #2196, opened 2026-04-30
[WC-3353][WC-3401]RichText: strict CSP support

gjulivan:
use class attributor instead of style attributor to avoid inline styling


samuelreichert (dismissed):

samuelreichert (dismissed):

gjulivan:
closing due to architectural changes

PR #2201, opened 2026-05-05
E2e reliability and performance

samuelreichert:
### Pull request type

No code changes (changes to documentation, CI, metadata, etc.)

---

### Description

This PR improves E2E test reliability and reduces nightly run duration through a set of infrastructure, stability, and parallelization improvements.

**Shared fixtures & Mendix helpers (`run-e2e`)**

- Added worker-scoped session fixture — each Playwright worker holds one Mendix session, eliminating per-test login/logout overhead and staying within the 5-session developer license limit
- Added `waitForMendixApp(page)` helper to replace `waitForLoadState("networkidle")`, which was unreliable and caused flaky tests
- Added `checkAccessibility` shared helper
- Added smoke suite support via `E2E_SUITE=smoke` env var and `@smoke` tag

**Flakiness fixes**

- Replaced all `page.waitForTimeout()` fixed delays with event-based waits
- Removed `waitForLoadState("networkidle")` across all specs
- Fixed race conditions in datagrid filter specs
- Fixed racy `nth(1)` assertion in badge-button close page test
- Migrated remaining specs that the codemod missed

**Screenshot baseline hardening**

- Set global `threshold: 0.1` and `animations: "disabled"` in playwright config
- Removed per-test loose threshold/maxDiffPixels overrides (0.2–0.5, up to 15000px) that were masking real pixel differences

**ESLint rules**

- Added `eslint-plugin-playwright` rules to prevent flaky patterns from creeping back: `no-wait-for-timeout` (error), `no-networkidle` (warn), `prefer-web-first-assertions` (warn)

**Parallelization**

- Nightly workflow now runs 4 parallel matrix runners instead of a single sequential job
- Tests split using weighted bin-packing (FFD algorithm) for balanced runner load
- Expected wall-clock time: ~80 min → ~20–25 min

**Turbo**

- Fixed stale Cypress input globs (`cypress/**`, `cypress.config.*`) replaced with Playwright equivalents (`e2e/**`, `playwright.config.*`) for correct cache invalidation

- Trigger the nightly workflow manually and verify all 4 matrix runners complete without infrastructure errors


leonardomendix (dismissed):

docs/requirements/e2e-test-guidelines.md
```diff
@@ -0,0 +1,136 @@
...
+| `el.nth(1)` to disambiguate                                          | More specific selector or wait first                 | nth() fragile to render order                    |
+| `page.$$eval(...)` to extract data                                   | `expect(locator).toContainText()` or `.toHaveText()` | evaluate snapshots DOM; locator assertions retry |
+
+Preferred assertions: `toBeVisible`, `toHaveText`, `toHaveCount`, `toHaveCSS`, `toContainText`, `toHaveScreenshot`.
+
+## Locator Patterns
+
+| Don't                                   | Do Instead                                 | Why                                     |
+| --------------------------------------- | ------------------------------------------ | --------------------------------------- |
+| `.nth(N)` on ambiguous selectors        | `.mx-name-*` attribute selectors           | nth fragile to DOM order                |
```
r0b1n: Not sure how this works. `.mx-name-` is for widget, if we need to find nth option in a list of a row in the data grid, this won't help.

docs/requirements/e2e-test-guidelines.md
```diff
@@ -0,0 +1,136 @@
...
+| Complex CSS selectors                   | `page.locator(".mx-name-widgetName")`      | mx-name attributes are stable, semantic |
+| `page.click("text=...")` for navigation | `page.locator(".mx-name-navItem").click()` | Text fragile to i18n/copy changes       |
```
r0b1n: I believe having more human-visible selector is better, so we should encourage role/text based selectors. Am I wrong?

docs/requirements/e2e-test-guidelines.md
```diff
@@ -0,0 +1,136 @@
...
+## Spec File Template
+
+```javascript
+import { test, expect } from "@mendix/run-e2e/fixtures";
+import { waitForMendixApp } from "@mendix/run-e2e/mendix-helpers";
+
+test.describe("WidgetName", () => {
+    test.beforeEach(async ({ page }) => {
+        await page.goto("/");
+        await waitForMendixApp(page);
```
r0b1n: Doc says: "Auto-wraps `page.goto()` to call `waitForMendixApp()`", so is this wait redundant?

docs/requirements/e2e-test-guidelines.md
```diff
@@ -0,0 +1,136 @@
...
+## Code Review Checklist
+
+- [ ] Uses `@mendix/run-e2e/fixtures` import (not `@playwright/test`)
+- [ ] No `waitForTimeout` calls
+- [ ] No `waitForLoadState("networkidle")` without explicit justification
+- [ ] All assertions use web-first Playwright assertions
+- [ ] No per-test screenshot threshold overrides
+- [ ] No manual `afterEach` logout
+- [ ] Locators use `.mx-name-*` attributes where possible
+- [ ] Tests tagged `@smoke` if they cover critical paths
```
r0b1n: How is this being used?

docs/requirements/e2e-test-guidelines.md
leonardomendix: I don't think using text is a good approach. The class or role is a more stable selector. In most of the tests, we try to use classes. I'm not against roles, but the class is more unique, so that avoids finding more than one on the page.

AGENTS.md
```diff
@@ -33,6 +33,17 @@ docs/requirements/                -> Detailed technical requirements
...
+
+- Import from `@mendix/run-e2e/fixtures` (not `@playwright/test`)
+- Wait with `waitForMendixApp(page)`, never `waitForTimeout` or `networkidle`
+- Use web-first assertions: `toBeVisible`, `toHaveCount`, `toContainText`, `toHaveCSS`
+- Locators: prefer `.mx-name-*` attributes over nth() or text selectors
+- Screenshots: no per-test threshold overrides, ensure element visible first
+- No manual afterEach logout — fixture handles session lifecycle
+- Tag critical-path tests with `@smoke`
+- See `docs/requirements/e2e-test-guidelines.md` for full rules + template
+
```
r0b1n, line 39: This feels like duplication, should we just point it out to the document

docs/requirements/e2e-test-guidelines.md
leonardomendix: I think that's fine, the don't is more to avoid ambiguous selectors. For a list row in the datagrid2 will still need a widget class name and a more complex CSS selector.

AGENTS.md
leonardomendix: Indeed!

docs/requirements/e2e-test-guidelines.md
leonardomendix: That was introduced in this PR. We're not using smoke tags.

docs/requirements/e2e-test-guidelines.md
r0b1n: I think we should guide it more towards using it when needed, `css selector to select widget` + `text selector` to select an element in it is better than a `complex CSS selector` to do the same. For example guiding towards a widget with css selector, and then selecting a dropdown option based on text. And we control the tested environment and if texts change, there is a reason behind it, and we probably should catch it.

leonardomendix: Actually, I think it's the opposite because this way we can see when the order of the DOM changes. Looking only for a text can lead to a false positive. To be honest, ideally, that should be checked in unit snapshot testing.  I liked the idea of the compose selector. We can suggest, as you mentioned, a CSS selector/ role + text/caption/label. It can use the playwright docs as a reference: https://playwright.dev/docs/locators.

leonardomendix (dismissed):

r0b1n (dismissed):

r0b1n (approved):

leonardomendix (approved):

PR #2208, opened 2026-05-11
[WC-3361] FileUploader: enhance limits and feedback

yordan-st:
### Pull request type

Bug fix + Feature (behavior change — non-breaking)

### Description

When the file upload limit was reached, the dropzone silently turned grey with no explanation. Dropping more files than the limit rejected the entire batch. There was no proper upload queue — excess files were marked as errors and "retried" via a promotion mechanism, conflating queueing with error state.

This PR addresses:

- Silent disabled dropzone → now shows "Maximum file count of X reached."
- Drop overflow now splits on remaining capacity. Only excess files are rejected; the rest upload normally.
- Replaced the error/retry approach with a proper upload queue. Files waiting for a concurrent slot show a "Waiting..." state, not an error state.
- New "Maximum concurrent uploads" property (XML key: `maxFilesPerBatch`) controls how many files upload simultaneously. Excess files queue and drain automatically.
- New "Upload queued" text property for the waiting state message.
- New "File limit reached" text property for the limit-reached message.
- Rejected files (over total limit) now show a **retry button**. The button is enabled when capacity is available and greyed out when the limit is still full. Files do NOT auto-promote — the user must click retry explicitly (confirmed with PM).
- Upload errors free a concurrent slot, allowing queued files to start uploading automatically.
- `maxFilesPerUpload` property is now optional — empty or 0 means unlimited.
- File list reordered: successful uploads shown above rejected files.

**No longer depends on WC-3363.** The `dismissValidationErrors` method was removed as part of this PR — rejected files are a distinct state from validation errors and are not dismissable. WC-3363 can merge independently.

**Setup:** File Uploader widget, files mode. Start with "Maximum number of files" = 5, "Maximum concurrent uploads" = 2.

**File limit feedback:**
1. Upload 1 file → dropzone stays active, no warning shown
2. Upload until total = 5 → dropzone turns grey AND message appears: "Maximum file count of 5 reached."
3. Remove 1 file while at limit → dropzone re-enables, message disappears

**Unlimited behavior:**
4. Set total limit = 0 → dropzone never disables regardless of file count
5. Leave total limit empty → same as 0, unlimited

**Concurrent upload queue:**
6. Set concurrent uploads = 2, drop 5 files → exactly 2 show "Uploading..." with progress bar, remaining 3 show "Waiting..." without progress bar
7. Wait for 1 upload to complete → next queued file automatically starts uploading (now shows "Uploading...")
8. Leave concurrent uploads empty → all dropped files start uploading simultaneously

**Total limit with manual retry:**
9. Set total limit = 3, concurrent uploads = unlimited. Drop 5 files → 3 upload, 2 show as rejected with "Maximum file count of 3 reached." and a retry button
10. While at limit (3 active files): retry button on rejected file is greyed out (disabled)
11. Remove 1 uploaded file → retry button on rejected file becomes enabled
12. Click retry → file starts uploading; remaining rejected file still shows greyed retry button
13. Let an upload fail → that file shows upload error; retry buttons on rejected files remain unchanged (no auto-promotion)

**List ordering:**
14. Upload a mix of valid and invalid files → successful uploads appear above rejected files

**Other:**
15. Read-only mode → dropzone not rendered, no regression
16. Image upload mode → same queue and limit behavior applies


packages/pluggableWidgets/file-uploader-web/src/stores/FileUploaderStore.ts
```diff
@@ -75,24 +78,45 @@ export class FileUploaderStore {
...
+                        f.fileStatus !== "removedFile" &&
+                        f.fileStatus !== "validationError"
+                ).length,
+            (count, prevCount) => {
+                if (count < prevCount) {
+                    this.retryLimitExceededFiles();
+                }
+            }
+        );
     }
```
r0b1n, line 111: It looks really cumbersome to have essentially upload-throttling mechanism implemented like this. If we want upload to happen automatically in batches, we should make a proper uploading logic that keeps files in the queue and uploads max X files at a time and files should be in a queueing state, not in the error/warning state then.

packages/pluggableWidgets/file-uploader-web/src/stores/FileUploaderStore.ts
```diff
@@ -116,11 +140,18 @@ export class FileUploaderStore {
...
 
     get maxFilesPerUpload(): number {
```
r0b1n: It is still called PerUpload, this is confusing, rename it to better match the purpose

r0b1n: THe XML we can't rename unfortunately, but this we def can

packages/pluggableWidgets/file-uploader-web/src/components/FileUploaderRoot.tsx
```diff
@@ -26,16 +26,15 @@ export const FileUploaderRoot = observer((props: FileUploaderContainerProps): Re
             {!rootStore.isReadOnly && (
                 <Dropzone
                     onDrop={onDrop}
-                    warningMessage={rootStore.errorMessage}
```
r0b1n, line 29: We probably don't use computed `errorMessage` anymore.

packages/pluggableWidgets/file-uploader-web/src/stores/FileUploaderStore.ts
```diff
@@ -75,24 +79,52 @@ export class FileUploaderStore {
...
+
+        // Reaction 1: active count drops → promote rejected files into queue
+        this._disposePromoteRejectedReaction = reaction(
+            () => this.activeCount,
+            (count, prevCount) => {
+                if (count < prevCount) {
+                    this.promoteRejectedFiles();
+                }
+            }
+        );
```
r0b1n: Not sure if we want this to happen automatically, it might be confusing (like files were not waiting, they were rejected, but suddenly started uploading). Maybe letting users manually click "retry" (if there is place) is a better option. Discuss with PM

packages/pluggableWidgets/file-uploader-web/src/components/FileUploaderRoot.tsx
r0b1n: And I guess it is time to move away from storing messages in the store, so we don't mix logic and representation, let the component show needed error message instead based on the [calculated] state of the root store.

packages/pluggableWidgets/file-uploader-web/src/stores/FileUploaderStore.ts
```diff
@@ -182,24 +277,23 @@ export class FileUploaderStore {
...
             );
+            this.files.unshift(newFileStore);
+        }
 
+        for (const file of capacityExcess) {
+            const newFileStore = FileStore.newRejectedFile(
+                file,
+                this.translations.get("uploadLimitReachedMessage", this.maxTotalFiles.toString()),
+                this
+            );
```
r0b1n: Looks like it is always the same error message for `rejected` files, so we likely don't need to carry the text itself in the store, the react components can show this message instead then when status is `rejected`.

packages/pluggableWidgets/file-uploader-web/src/components/FileEntry.tsx
```diff
@@ -37,8 +48,21 @@ export const FileEntryContainer = observer(({ store, actions }: FileEntryContain
...
+                    store={store}
+                    fileStatus={store.fileStatus}
+                    canRetry={store.canRetry}
+                    canDownload={store.canDownload}
+                    canRemove={store.canRemove}
+                    onRetry={onRetry}
+                    onRemove={onRemove}
+                    onDownload={onDownload}
+                />
+            }
```
r0b1n: Why exposing all of this here, can't `ActionsBar` do it itself based on store, it has it already?

packages/pluggableWidgets/file-uploader-web/src/stores/FileStore.ts
```diff
@@ -156,6 +159,18 @@ export class FileStore {
...
+    get canRetry(): boolean {
+        return this.fileStatus === "rejected" && !this._rootStore.isFileUploadLimitReached;
+    }
+
+    retry(): void {
+        if (!this.canRetry) {
+            return;
+        }
+        this.setQueued();
+        this._rootStore.promoteQueuedFiles();
```
r0b1n: If we go reactive route, this also shouldn't be needed, rootStore must pick up the fileStatus change automatically and do the right thing.

packages/pluggableWidgets/file-uploader-web/src/components/FileEntry.tsx
```diff
@@ -37,8 +48,21 @@ export const FileEntryContainer = observer(({ store, actions }: FileEntryContain
             mimeType={store.mimeType}
             fileStatus={store.fileStatus}
             errorMessage={store.errorDescription}
+            maxTotalFiles={store._rootStore.maxTotalFiles}
```
r0b1n: It should not be used from components, make `_rootStore` private in `FileStore`. Adjust this code accordingly.

packages/pluggableWidgets/file-uploader-web/src/components/ActionButton.tsx
```diff
@@ -22,7 +23,8 @@ export function ActionButton({ action, icon, title, isDisabled }: ActionButtonPr
         <button
             role={"button"}
             className={classNames("action-button", {
-                disabled: isDisabled
+                disabled: isDisabled,
+                hidden: isHidden ?? isDisabled
```
r0b1n: So when button is not passing this flag, it starts hiding when disabled? This changes default behavior, iiuc. 

And this code is hard to grasp, make it better and easier to read, if this is really a thing to switch between those two behaviors.

packages/pluggableWidgets/file-uploader-web/src/components/FileEntry.tsx
r0b1n: And other fields starting with `_` as well. In all stores.

packages/pluggableWidgets/file-uploader-web/src/stores/FileStore.ts
```diff
@@ -51,29 +51,54 @@ export class FileStore {
...
+
+    get statusMessage(): string | undefined {
+        if (this.fileStatus === "rejected") {
+            return this._rootStore.translations.get(
+                "uploadLimitReachedMessage",
+                this._rootStore.maxTotalFiles.toString()
+            );
+        }
+        return this.errorDescription;
+    }
```
r0b1n: Message in case of `rejected` is the same for all files, there is no need to compute it per file, let view component render it when status is `rejected`.

packages/pluggableWidgets/file-uploader-web/src/stores/FileUploaderStore.ts
```diff
@@ -22,21 +22,25 @@ export class FileUploaderStore {
...
+    private _maxFileSize = 0;
+    private _maxTotalFiles: DynamicValue<Big> | undefined;
+    private _maxConcurrentUploads: DynamicValue<Big> | undefined;
+    private _disposePromoteQueuedReaction: (() => void) | undefined;
+    private _disposeQueuedCountReaction: (() => void) | undefined;
 
-    errorMessage?: string = undefined;
+    createActionFailed = false;
 
     translations: TranslationsStore;
```
r0b1n: Make this private as well.

packages/pluggableWidgets/file-uploader-web/src/stores/FileUploaderStore.ts
```diff
@@ -71,28 +75,56 @@ export class FileUploaderStore {
...
+            isFileUploadLimitReached: computed,
+            sortedFiles: computed,
+            activeCount: computed,
+            uploadingCount: computed,
+            queuedCount: computed
         });
 
         this.updateProps(props);
+
+        // Reaction 1: uploading count drops → promote queued files to uploading
```
r0b1n: They are almost identical. Can we make into one reaction?

packages/pluggableWidgets/file-uploader-web/src/stores/FileStore.ts
```diff
@@ -84,14 +83,8 @@ export class FileStore {
...
-        if (this.fileStatus === "rejected") {
-            return this._rootStore.translations.get(
-                "uploadLimitReachedMessage",
-                this._rootStore.maxTotalFiles.toString()
-            );
-        }
-        return this.errorDescription;
+    get maxTotalFiles(): number {
+        return this._rootStore.maxTotalFiles;
     }
```
r0b1n: File store should not be responsible for passing parameters it doesn't care about to view components. View components should find their own way.

leonardomendix (dismissed):

r0b1n (dismissed):

leonardomendix (approved):

samuelreichert (approved):

PR #2220, opened 2026-05-19
[WC-3322]: Fix slider and range-slider decimal places formatting

samuelreichert:
### Pull request type

---

### Description

When `decimalPlaces` is configured on the **Slider** or **Range Slider** widget, displayed values were silently stripping trailing zeros and ignoring the user's locale:

- Mark labels: `10.00` rendered as `10`, `9.20` rendered as `9.2`
- Tooltip: raw RC Slider number passed through with no formatting applied

**Root causes:**

1. `marks.ts` — `parseFloat(value.toFixed(n)).toString()` round-tripped through a number, discarding trailing zeros. Fixed by keeping the formatted string as the label and using `parseFloat` only for the numeric key (so rc-slider positions dots correctly even when the decimal separator is a comma).

2. `createHandleRender.tsx` (Slider) / `Container.tsx` (Range Slider) — tooltip overlay received `restProps.value` (raw RC Slider number). Fixed by passing a `ValueFormatter` derived from the attribute's own Mendix `NumberFormatter`, so the decimal separator and thousands grouping follow the user's session locale.

**Shared logic extracted to `widget-plugin-platform`:**

`createValueFormatter` and `createMarks` were duplicated across both widgets. They have been moved to `packages/shared/widget-plugin-platform/src/utils/` so both widgets consume a single implementation. Tests live in the shared package only.

**Range Slider specifics:**

Uses a single formatter derived from `lowerBoundAttribute` for both marks and tooltips (both handles). The attribute formatter already carries the correct locale settings so there is no need for a separate formatter per handle.

---

**Slider:**

1. Configure a Slider widget with `decimalPlaces = 2` and `Number of markers > 0`.
2. Set min/max so some markers land on whole numbers (e.g. min=0, max=10, markers=2).
3. **Mark labels** should show `0.00`, `5.00`, `10.00` — not `0`, `5`, `10`.
4. Drag the handle to a whole number — **tooltip** should show `10.00`, not `10`.
5. Drag to a value with one decimal (e.g. 9.5) — tooltip should show `9.50`.
6. Set `decimalPlaces = 0` — labels and tooltip show integers, no regression.
7. Switch app locale to one with comma decimal separator — marks and tooltip use `,` not `.`.

**Range Slider:**

1. Same mark label checks as Slider above.
2. Drag the **lower** handle to a whole number — tooltip shows `10.00`.
3. Drag the **upper** handle — tooltip also formats correctly (uses same formatter as lower handle).
4. Custom text tooltip type — still shows custom text unchanged.
5. Set `decimalPlaces = 0` — no regression.


iobuhov (dismissed):
LGTM

packages/pluggableWidgets/slider-web/src/utils/createHandleRender.tsx
```diff
@@ -29,7 +31,7 @@ export function createHandleRender({
                 getTooltipContainer={() => sliderRef.current ?? document.body}
                 defaultVisible
                 prefixCls="rc-slider-tooltip"
-                overlay={isCustomText ? overlay : restProps.value}
+                overlay={isCustomText ? overlay : restProps.value.toFixed(decimalPlaces)}
```
r0b1n: I am thinking, default `toFixed` doesn't respect the user settings, like if current Mendix locale is using `,` as decimal separator, toFixed will still be `.` , right? Same with thousands separator.

packages/pluggableWidgets/slider-web/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+- We fixed mark labels and tooltip values not respecting the value attribute's number formatting. They now use the attribute's own formatter with the configured number of decimal places, so the decimal separator and thousands grouping follow the current locale and the attribute's settings (e.g., `10` displays as `10.00` and `9.2` as `9.20` with two decimal places, and grouping like `1,000,000` is preserved when enabled).
```
r0b1n: Remove implementation details, describe user facing behavior.

packages/pluggableWidgets/slider-web/src/utils/__tests__/marks.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    (value: number): string => {
+        const fixed = value.toFixed(decimalPlaces);
+        return decimalSeparator === "." ? fixed : fixed.replace(".", decimalSeparator);
+    };
+
+describe("createMarks", () => {
+    it("forces trailing zeros when decimalPlaces > 0 and value is whole number", () => {
+        const marks = createMarks({ numberOfMarks: 2, decimalPlaces: 2, format: formatWith(2), min: 0, max: 10 });
+        expect(marks).toBeDefined();
+        expect(marks![0]).toBe("0.00");
```
r0b1n, line 14: nit: Why this pattern, and not `marks?.[0]`?

samuelreichert: On line 14 we already check if marks is defined

gjulivan (approved):

leonardomendix (approved):

PR #2221, opened 2026-05-19
[WC-3398] popup menu hover issue

r0b1n:
### Pull request type

---

- Click outside closes menu that opened on hover (though, it is a bit of a niche use-case?)
- Nested menus are working as expected when "Close on" set to "Hover leave" and it is possible to navigate to the submenu.


packages/pluggableWidgets/popup-menu-web/src/__tests__/usePopup.spec.tsx
rahmanunver, line 1: Let's populate this test

r0b1n: added some

iobuhov (approved):
LGTM

leonardomendix (approved):

PR #2222, opened 2026-05-20
DG2: consolidate virtual scroll height lock (WC-3333, WC-3390)

yordan-st:
### Pull request type

Bug fix and refactor

---

### Description

Two virtual scroll bugs in `GridSizeStore` shared the same root cause — incomplete height cache invalidation and incorrect overflow detection. This PR consolidates both fixes instead of band-aiding each symptom separately.

**What changed:**

- Replaced parallel `lockedAtPageSize` / `lockedAtColumnCount` guard fields with a single `layoutKey` private getter (`pageSize-columnCount`). One place to invalidate, one place to extend.
- Height is no longer cleared to `undefined` on invalidation — the old value stays until overwritten. Prevents CSS snap-to-full-height flicker and eliminates false scroll-bottom triggers during column toggle.
- Guarded `bumpPage` in scroll handler behind `scrollTop > 0` — prevents page loading on content-resize events when the user is at the top.

Closes #2167, closes #2193.

---

**Bug 1 — WC-3333: Scrollbar disappears after hiding a column**

1. Open a page with Data Grid, virtual scrolling enabled, enough rows to produce a scrollbar. Include at least one column with wide/wrapping text.
2. Confirm vertical scrollbar is visible.
3. Open column selector, hide the wide text column.
4. **Should:** scrollbar stays visible, row count unchanged.
5. **Should not:** scrollbar disappears, or extra rows load without scrolling.
6. Show the column again — scrollbar should adjust correctly.

**Bug 2 — WC-3390: Many columns, only first page loads**

1. Open a page with Data Grid, virtual scrolling enabled, 50+ row dataset, enough columns to require horizontal scrolling.
2. On load, confirm ~pageSize rows are visible and a vertical scrollbar is present.
3. Scroll down — more rows should load.
4. **Should not:** only 4–5 rows visible with no vertical scrollbar, loading blocked until zoom or horizontal scroll.

### DIFF
--- a/packages/pluggableWidgets/datagrid-web/src/model/stores/GridSize.store.ts
+++ b/packages/pluggableWidgets/datagrid-web/src/model/stores/GridSize.store.ts
@@ -13,23 +13,28 @@ export class GridSizeStore {
 
     gridContainerHeight?: number;
 
-    private lockedAtPageSize?: number;
+    private lockedAtLayoutKey?: string;
 
     constructor(
         private readonly hasMoreItemsAtom: ComputedAtom<boolean | undefined>,
         private readonly paginationConfig: PaginationConfig,
         private readonly setPageAction: SetPageAction,
-        private readonly pageSizeAtom: ComputedAtom<number>
+        private readonly pageSizeAtom: ComputedAtom<number>,
+        private readonly visibleColumnsCountAtom: ComputedAtom<number>
     ) {
-        makeAutoObservable<GridSizeStore, "lockedAtPageSize">(this, {
+        makeAutoObservable<GridSizeStore, "lockedAtLayoutKey">(this, {
             gridContainerRef: false,
             gridBodyRef: false,
             gridHeaderRef: false,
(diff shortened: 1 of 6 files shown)


r0b1n (dismissed):

packages/pluggableWidgets/datagrid-web/src/model/stores/GridSize.store.ts
```diff
@@ -111,6 +114,6 @@ export class GridSizeStore {
         // that we create a small synthetic overflow that makes the body scrollable.
         const overflows = gridContainer.scrollHeight > fullHeight;
```
leonardomendix, line 115: Do you have one example when that could be false? It seems this boolean will always be true.

leonardomendix: Already on main.

leonardomendix (approved):

r0b1n (approved):

PR #2224, opened 2026-05-21
[WC-3420] Pusher initial implementation

r0b1n:
### Pull request type

---


gjulivan (dismissed):

packages/pluggableWidgets/pusher-web/src/utils/fetchPusherConfig.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Fetches Pusher configuration from the backend.
+ * Returns a PusherConfig on success, or null on error / invalid response.
+ */
+export async function fetchPusherConfig(signal: AbortSignal): Promise<PusherConfig | null> {
+    const baseUrl = (window as any).mx?.remoteUrl || (window as any).mx?.appUrl || "";
+    const endpoint = `${baseUrl}rest/pusher/key`;
+    const csrfToken = (window as any).mx?.sessionData?.csrftoken || "";
+    const authEndpoint = `${baseUrl}rest/pusher/auth`;
```
iobuhov: Let's use URL constructor?

iobuhov (commented):
Where is module code?

packages/pluggableWidgets/pusher-web/src/utils/fetchPusherConfig.ts
r0b1n: ✅

packages/pluggableWidgets/pusher-web/src/hooks/useFetchPusherConfig.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Provides Pusher configuration fetched from the backend.
+ * Returns null while loading or on error.
+ */
+export function useFetchPusherConfig(): PusherConfig | null {
+    const [config, setConfig] = useState<PusherConfig | null>(null);
+
+    useEffect(() => {
+        let active = true;
```
iobuhov: I don't think we need `active` here. We can use `ctrl.signal.aborted` flag

packages/pluggableWidgets/pusher-web/src/hooks/usePusherListener.ts
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useRef, useState } from "react";
+import { useFetchPusherConfig } from "./useFetchPusherConfig";
+import { PusherListener } from "../utils/PusherListener";
+
+/**
+ * Creates and initializes a PusherListener
+ */
+export function usePusherListener(): PusherListener | null {
```
iobuhov: I think usePusherListener and useFetchPusherConfig should be merge into single factory. They have dependency that seems separated just for seek of hooks, don't think it adds lot of value. useFetchPusherConfig don't have any args. We can create one clean factory that fetches config, init the instance and returns it wrapped in promise. This way whole factory can be passed to useState and we have ether instance or null. This way we have less coupling with react.

packages/pluggableWidgets/pusher-web/src/utils/PusherListener.ts
```diff
@@ -0,0 +1,126 @@
...
+    private currentChannelName: string | null = null;
+    private currentEventName: string | null = null;
+
+    constructor(private config: PusherConfig) {}
+
+    /**
+     * Initialize Pusher connection
+     * Should be called once on widget mount
+     */
+    initialize(): void {
```
iobuhov: I think this code belongs to constructor. Init just adds extra step for no reason.

packages/pluggableWidgets/pusher-web/src/utils/PusherListener.ts
```diff
@@ -0,0 +1,126 @@
...
+        this.pusher.connection.bind("error", this.handleConnectionError);
+        this.pusher.connection.bind("state_change", this.handleStateChange);
+    }
+
+    /**
+     * Subscribe to channel for specific object and event
+     * Automatically unsubscribes from previous channel if different
+     */
+    subscribe(config: SubscriptionConfig): void {
+        if (!this.pusher) {
```
iobuhov: If we move init code to constructor this prop can be set to not-nullable

packages/pluggableWidgets/pusher-web/src/utils/PusherListener.ts
```diff
@@ -0,0 +1,126 @@
...
+    destroy(): void {
+        this.unsubscribe();
+        if (this.pusher) {
+            this.pusher.connection.unbind();
+            this.pusher.disconnect();
+            this.pusher = null;
+        }
+    }
+
+    private buildChannelName(entityName: string, guid: string): string {
```
iobuhov: I think this should be a getter

packages/pluggableWidgets/pusher-web/src/utils/getChannelName.ts
```diff
@@ -0,0 +1,32 @@
+import { DynamicValue, ObjectItem } from "mendix";
+import { useMemo } from "react";
+
+interface MxObjectInfo {
+    guid: string;
+    entityName: string;
+}
+
+export function useMxObjectInfo(objectSource: DynamicValue<ObjectItem>): MxObjectInfo | undefined {
+    const object = (objectSource as any)?.value as ObjectItem | undefined;
```
iobuhov: Are we going to keep this `any`?

packages/pluggableWidgets/pusher-web/src/Pusher.tsx
```diff
@@ -0,0 +1,50 @@
...
+        [notifyEventAction]
+    );
+
+    // Error callback
+    const handleError = useCallback((error: Error) => {
+        console.error("[Pusher] Subscription error:", error.message);
+    }, []);
+
+    // Setup stable subscription config
+    const subscription = useMemo(() => {
```
iobuhov, line 32: NIT: I would extract this block to small function for better testing

iobuhov (changes requested):

packages/pluggableWidgets/pusher-web/src/hooks/usePusherListener.ts
r0b1n: I reworked the flow to have less hooks

iobuhov (approved):
LGTM

leonardomendix (approved):

PR #2226, opened 2026-05-25
Modify post-edit.sh to include debugging output

moroq-h1:
### Pull request type

---


iobuhov:
This introduce risk of exposing private information. You can use this hook with your settings.local.json

PR #2232, opened 2026-05-26
test(video-player): wait for widget after tab clicks, fix poster race

samuelreichert:
### Pull request type

---

### Description

Fixes intermittent nightly failures in `video-player-web` and `accessibility-helper-web` E2E tests. Both share the same class of bug: actions fired before Mendix finishes re-rendering the DOM.

**video-player-web** — three root causes:

1. **Tab page tests** — after `.click()` on a tab, tests asserted `toBeVisible()` on iframes immediately. Mendix renders tab content asynchronously; the fixture's `waitForMendixApp` only patches `page.goto`, not tab navigation clicks. Added `waitForWidget` after each tab click.

2. **Aspect ratio test** — same tab-click race: `boundingBox()` called on elements in tabs 2 and 3 without waiting for them to be present in the DOM.

3. **Poster screenshot** — `page.evaluate` with a standalone `new Image()` resolves when the image loads into memory, not when the `<video>` element has painted the poster frame. Switched to `videoLocator.evaluate(el => ...)` to tie the wait to the actual element. Also added `scrollIntoViewIfNeeded` so the widget is in viewport before screenshotting.

**accessibility-helper-web** — one root cause, three affected tests:

4. **Double-click on `actionButton2`** — first click triggers a Mendix re-render that detaches the button from the DOM; second click immediately after races against re-attach and times out after 10s. Added `waitForMendixApp` between the two clicks in all three `with multiple targets` tests that used this pattern.

### DIFF
--- a/packages/pluggableWidgets/video-player-web/e2e/VideoPlayer.spec.js
+++ b/packages/pluggableWidgets/video-player-web/e2e/VideoPlayer.spec.js
@@ -1,5 +1,5 @@
 import { test, expect } from "@mendix/run-e2e/fixtures";
-import { waitForWidget } from "@mendix/run-e2e/mendix-helpers";
+import { waitForWidget, waitFrames } from "@mendix/run-e2e/mendix-helpers";
 
 test.describe("Video Player", () => {
     test.beforeEach(async ({ page }) => {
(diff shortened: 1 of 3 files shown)


r0b1n (dismissed):

packages/pluggableWidgets/video-player-web/e2e/VideoPlayer.spec.js
```diff
@@ -112,7 +116,21 @@ test.describe("External video", () => {
...
+                ? new Promise(r => {
+                      const i = new Image();
+                      i.onload = i.onerror = r;
+                      i.src = el.poster;
+                      if (i.complete) r();
+                  })
+                : Promise.resolve()
+        );
+        // Wait two animation frames so the browser flushes layout and paints the poster frame.
+        await page.evaluate(() => new Promise(r => requestAnimationFrame(() => requestAnimationFrame(r))));
```
iobuhov: Let's create utility for that? `waitFrames(n)`?

r0b1n (approved):

iobuhov (approved):

PR #2235, opened 2026-05-27
[WC-3337]: Pluggable Image Crop Widget

rahmanunver:
### Pull request type

---

### Description

Adds **Image Crop**, a new pluggable widget that lets end-users crop an image directly inside a Mendix page and saves the result back to an Image attribute.

- Shows the bound image inside a fixed crop area and lets the user drag a selection over it.
- Supports **rectangle** or **circle** crop shapes.
- Supports **aspect-ratio presets** (Free, 1:1, 16:9, 4:3, 3:4) plus a **Custom W:H** option.
- Optional **zoom** with a slider and mouse-wheel (always, only with Ctrl/Cmd, or off).
- Optional **live preview** at a configurable size, so users see the result before saving.
- On click of the crop button, the result is saved into the bound Image attribute and an optional action runs.
- Output size can match the visible crop area or the original image resolution.


packages/pluggableWidgets/image-cropper-web/src/ImageCropper.xml
```diff
@@ -0,0 +1,130 @@
...
+                <property key="wheelZoomMode" type="enumeration" defaultValue="onWithCtrl" required="true">
+                    <caption>Mouse wheel zoom</caption>
+                    <description>Whether the mouse wheel zooms the image. "On (hold Ctrl)" keeps page scroll working.</description>
+                    <enumerationValues>
+                        <enumerationValue key="off">Off</enumerationValue>
+                        <enumerationValue key="on">On</enumerationValue>
+                        <enumerationValue key="onWithCtrl">On (hold Ctrl)</enumerationValue>
+                    </enumerationValues>
+                </property>
+                <property key="minZoom" type="decimal" defaultValue="1" required="true">
```
r0b1n, line 102: Maybe just an integer?

packages/pluggableWidgets/image-crop-web/src/ImageCrop.xml
```diff
@@ -0,0 +1,130 @@
...
+                </property>
+                <property key="outputQuality" type="decimal" defaultValue="0.92" required="true">
+                    <caption>JPEG quality (0.0 - 1.0)</caption>
+                    <description>JPEG compression. Higher = sharper and larger. Ignored for PNG.</description>
+                </property>
+                <property key="outputSize" type="enumeration" defaultValue="original" required="true">
+                    <caption>Output size</caption>
+                    <description>Resolution of the saved crop. Original is sharpest; Viewport matches the on-screen canvas size.</description>
+                    <enumerationValues>
+                        <enumerationValue key="viewport">Viewport (canvas dims)</enumerationValue>
```
r0b1n: `dims` means dimensions?

packages/pluggableWidgets/image-crop-web/src/ImageCrop.xml
```diff
@@ -0,0 +1,130 @@
...
+                        <enumerationValue key="original">Original (source resolution)</enumerationValue>
+                    </enumerationValues>
+                </property>
+            </propertyGroup>
+            <propertyGroup caption="Action">
+                <property key="cropButtonCaption" type="textTemplate" required="false">
+                    <caption>Crop button caption</caption>
+                    <description>Label on the Crop button.</description>
+                    <translations>
+                        <translation lang="en_US">Crop</translation>
```
r0b1n: nl_NL translation missing.

packages/pluggableWidgets/image-cropper-web/src/utils/aspectRatio.ts
```diff
@@ -0,0 +1,29 @@
...
+        case "portrait3x4":
+            return 3 / 4;
+        case "custom":
+            if (customWidth > 0 && customHeight > 0) {
+                return customWidth / customHeight;
+            }
+            return undefined;
+        default: {
+            const _exhaustive: never = aspect;
+            return _exhaustive;
```
r0b1n, line 26: What is this?

packages/pluggableWidgets/image-crop-web/src/components/CropArea.tsx
```diff
@@ -0,0 +1,160 @@
...
+            imageRef.current = img;
+        },
+        [imageRef]
+    );
+
+    if (loadError) {
+        return (
+            <div className="widget-image-crop__error">
+                Image source does not allow cropping. Upload locally or configure CORS.
+            </div>
```
r0b1n: Not sure if this error message makes any sense.

packages/pluggableWidgets/image-crop-web/src/components/CropArea.tsx
```diff
@@ -0,0 +1,160 @@
...
+        },
+        [aspect, onImageLoad, boundaryWidth, boundaryHeight]
+    );
+
+    const setImageRef = useCallback(
+        (img: HTMLImageElement | null) => {
+            imageRef.current = img;
+        },
+        [imageRef]
+    );
```
r0b1n: Is it needed? Maybe we can pass `imageRef` as `ref` directly.

packages/pluggableWidgets/image-crop-web/src/components/CropArea.tsx
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    return (
+        <div
+            ref={containerRef}
+            className={classNames("widget-image-crop__canvas", {
+                "widget-image-crop__canvas--circle": props.circular
+            })}
+            style={{ maxWidth: props.boundaryWidth, maxHeight: props.boundaryHeight }}
+        >
```
r0b1n: We can decouple container wrapper with zoom logic to a separate component.

packages/pluggableWidgets/image-crop-web/src/components/CropArea.tsx
```diff
@@ -0,0 +1,160 @@
...
+            <ReactCrop
+                crop={props.crop}
+                onChange={(_pixel, percent) => props.onCropChange(percent)}
+                onComplete={pixel => props.onCropComplete(pixel)}
+                aspect={props.aspect}
+                circularCrop={props.circular}
+                disabled={!props.resizable}
+                locked={!props.resizable}
```
r0b1n: Should it really be locked if not resizable?

packages/pluggableWidgets/image-crop-web/src/components/CropArea.tsx
```diff
@@ -0,0 +1,160 @@
...
+                    style={{
+                        width: displaySize?.width,
+                        height: displaySize?.height,
+                        maxWidth: displaySize ? undefined : props.boundaryWidth,
+                        maxHeight: displaySize ? undefined : props.boundaryHeight,
+                        transform: `scale(${props.zoom})`,
+                        transformOrigin: "center"
+                    }}
+                    onLoad={handleImageLoad}
+                    onError={(_e: SyntheticEvent<HTMLImageElement>) => setLoadError(true)}
```
r0b1n: You likely can use empty signature `() => `

packages/pluggableWidgets/image-crop-web/src/hooks/useImageCropState.ts
```diff
@@ -0,0 +1,20 @@
...
+    zoom: number;
+    setZoom: Dispatch<SetStateAction<number>>;
+    imageRef: RefObject<HTMLImageElement | null>;
+}
+
+export function useImageCropState(initialZoom: number): ImageCropState {
+    const [crop, setCrop] = useState<Crop | undefined>(undefined);
+    const [completedCrop, setCompletedCrop] = useState<PixelCrop | undefined>(undefined);
+    const [zoom, setZoom] = useState<number>(initialZoom);
+    const imageRef = useRef<HTMLImageElement | null>(null);
```
r0b1n: `useRef` adds `| null` by default in this case. Check other places as well.

packages/pluggableWidgets/image-cropper-web/src/hooks/useImageCropperState.ts
```diff
@@ -0,0 +1,20 @@
+import { Dispatch, RefObject, SetStateAction, useRef, useState } from "react";
+import type { Crop, PixelCrop } from "react-image-crop";
+
+interface ImageCropState {
+    crop: Crop | undefined;
+    setCrop: Dispatch<SetStateAction<Crop | undefined>>;
+    completedCrop: PixelCrop | undefined;
+    setCompletedCrop: Dispatch<SetStateAction<PixelCrop | undefined>>;
```
r0b1n: What is the difference between those, can we use only one instead of two?

packages/pluggableWidgets/image-cropper-web/src/components/CropArea.tsx
```diff
@@ -0,0 +1,160 @@
...
+                    src={props.src}
+                    alt=""
+                    crossOrigin="anonymous"
+                    style={{
+                        width: displaySize?.width,
+                        height: displaySize?.height,
+                        maxWidth: displaySize ? undefined : props.boundaryWidth,
+                        maxHeight: displaySize ? undefined : props.boundaryHeight,
+                        transform: `scale(${props.zoom})`,
+                        transformOrigin: "center"
```
r0b1n, line 119: Do I understand it correctly that it only zooms into the center, like there is no way to zoom in and crop a corner of and image?

packages/pluggableWidgets/image-cropper-web/src/ImageCropper.xml
rahmanunver: Decimal is intentional here. Zoom is a multiplier (1× = image fits the canvas, 2× = double, 0.5× = half-size). Real-world configurations need fractional values:
- minZoom: 0.5 — lets users zoom out further than fit, useful when the source image is larger than the canvas and the user wants to see the whole thing before cropping.
- maxZoom: 2.5 — caps zoom mid-step so the slider doesn't overshoot detail.

Forcing integers would round these to the nearest whole multiple, breaking the slider's natural feel and the wheel-zoom step.

packages/pluggableWidgets/image-cropper-web/src/components/CropArea.tsx
rahmanunver: That's for the zoom in/out either with gesture or slider, we have drag-to-crop for corner edge cropping (ss is for circle crop).
Screenshot 2026-05-28 at 15 38 48

packages/pluggableWidgets/image-cropper-web/src/hooks/useImageCropperState.ts
rahmanunver: They fire on different events (onChange per pointer move vs onComplete on release), so collapsing them either makes the selection rectangle freeze during drag or makes the preview canvas re-render dozens of times per second.

packages/pluggableWidgets/image-crop-web/src/ImageCrop.xml
rahmanunver: Will rename to dimensions

packages/pluggableWidgets/image-crop-web/src/ImageCrop.xml
rahmanunver: Adding it.

packages/pluggableWidgets/image-crop-web/src/components/CropArea.tsx
rahmanunver: Will rephrase

packages/pluggableWidgets/image-crop-web/src/components/CropArea.tsx
rahmanunver: Updating it, stale change.

packages/pluggableWidgets/image-crop-web/src/components/CropArea.tsx
rahmanunver: refactoring into a ZoomContainer

packages/pluggableWidgets/image-crop-web/src/components/CropArea.tsx
rahmanunver: Will drop, keep only disabled

packages/pluggableWidgets/image-cropper-web/src/utils/aspectRatio.ts
rahmanunver: TypeScript exhaustiveness guard;
Imagine adding a new case, a new aspectRatio in the .xml, but don't add the new case in this switch;
After all existing cases are ruled out, since no case for the new aspect ratio exists, it will return undefined, causing the widget to act as if the setting was set to "free", and have no warnings.

A new "landscape21x9" is added as an option in the xml.
Adding a default case and setting the aspect to a value with type never:
```
error TS2322: Type '"landscape21x9"' is not assignable to type 'never'.
  src/utils/aspectRatio.ts:25  const _exhaustive: never = aspect;

```

The build fails with the error pointing at exactly this line, and the developer is forced to add the missing case "landscape21x9": return 21 / 9; before they can ship.

packages/pluggableWidgets/image-cropper-web/src/hooks/useImageCropperState.ts
r0b1n: Give them better names perhaps

r0b1n: And why do they use different types, is it necessary? (looks like one is subtype of another)

r0b1n (dismissed):

leonardomendix (dismissed):

leonardomendix (approved):

r0b1n (approved):

PR #2236, opened 2026-05-28
[WC-3417] Fix string empty/not-empty filters

r0b1n:
## Summary

Fixed an issue where "is empty" and "is not empty" string filters were not working correctly in some cases. 

The root cause was that string attributes can be in two different states: `empty` or `""` (empty string). Previously, the `empty()` filter matched both states, but due to new string behavior in recent Mendix versions, empty strings must now be matched explicitly.

## Changes

- Updated `BaseInputFilterStore.getFilterCondition()` to generate compound filter expressions for string attributes:
  - "is empty": `attr = empty OR attr = ""`
  - "is not empty": `attr != empty AND attr != ""`
- Added helper functions `isEmptyStringExp()` and `isNotEmptyStringExp()` to recognize these compound expressions
- Updated `inputStateFromCond()` to restore filter state from compound expressions
- Added comprehensive test coverage for the new expression pattern matching

## Test plan

- [x] Unit tests pass
- [x] E2E tests pass
- [x] Manual verification in test project with string filters

### DIFF
--- a/packages/shared/filter-commons/src/__tests__/condition-utils.spec.ts
+++ b/packages/shared/filter-commons/src/__tests__/condition-utils.spec.ts
@@ -1,8 +1,18 @@
 jest.mock("mendix/filters/builders");
 
 import { AndCondition, FilterCondition } from "mendix/filters";
-import { equals, literal } from "mendix/filters/builders";
-import { reduceArray, reduceMap, restoreArray, restoreMap, tag } from "../condition-utils";
+import { and, attribute, equals, literal, notEqual, or } from "mendix/filters/builders";
+import {
+    isEmptyExp,
+    isEmptyStringExp,
+    isNotEmptyExp,
+    isNotEmptyStringExp,
+    reduceArray,
+    reduceMap,
+    restoreArray,
+    restoreMap,
+    tag
+} from "../condition-utils";
 
 describe("condition-utils", () => {
     describe("reduceMap", () => {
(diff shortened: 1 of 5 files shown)


packages/shared/filter-commons/src/__tests__/condition-utils.spec.ts
```diff
@@ -412,4 +422,124 @@ describe("condition-utils", () => {
...
+    describe("isEmptyStringExp", () => {
+        it("identifies correct empty expression", () => {
+            const emptyStringExpression = or(
+                equals(attribute("testAttrId" as any), literal(undefined)),
+                equals(attribute("testAttrId" as any), literal(""))
+            );
+            expect(isEmptyStringExp(emptyStringExpression)).toBe(true);
+        });
+
+        it("ignores mismatching expressions", () => {
```
iobuhov: let's say explicitly "unequal" or "with different ids"

packages/shared/filter-commons/src/__tests__/condition-utils.spec.ts
```diff
@@ -412,4 +422,124 @@ describe("condition-utils", () => {
...
+            const mismatchedAttributes = or(
+                equals(attribute("testAttrId1" as any), literal(undefined)),
+                equals(attribute("testAttrId2" as any), literal(""))
+            );
+            expect(isEmptyStringExp(mismatchedAttributes)).toBe(false);
+
+            const mismatchedOrder = or(
```
iobuhov, line 444: nit: maybe let's split this into separate test?

packages/shared/filter-commons/src/__tests__/condition-utils.spec.ts
```diff
@@ -412,4 +422,124 @@ describe("condition-utils", () => {
...
+    describe("isNotEmptyStringExp", () => {
+        it("identifies correct not-empty expression", () => {
+            const notEmptyStringExpression = and(
+                notEqual(attribute("testAttrId" as any), literal(undefined)),
+                notEqual(attribute("testAttrId" as any), literal(""))
+            );
+            expect(isNotEmptyStringExp(notEmptyStringExpression)).toBe(true);
+        });
+
+        it("ignores mismatching expressions", () => {
```
iobuhov: Same as in empty teests - split and give clear diff between correct and "mismatched" exp

packages/shared/filter-commons/src/__tests__/condition-utils.spec.ts
```diff
@@ -412,4 +422,124 @@ describe("condition-utils", () => {
...
+        });
+    });
+
+    describe("isEmptyExp", () => {
+        it("returns true for equals expression with undefined literal", () => {
+            const emptyExp = equals(attribute("testAttrId" as any), literal(undefined));
+            expect(isEmptyExp(emptyExp)).toBe(true);
+        });
+
+        it("returns false for non-binary expression", () => {
```
iobuhov: nit: maybe worth to add a test where we use same exp to show that isEmptyExp returns false where isEmptyStringExp returns true

iobuhov (changes requested):

iobuhov (dismissed):
LGTM

iobuhov (approved):

leonardomendix (approved):

PR #2237, opened 2026-05-29
[WC-3409] Combobox: prevent 1px collapse

yordan-st:
## Summary

- Fixes WC-3409: Combobox collapses to 1px when placed in a flex row container with `align-items: center`
- Root cause: Mendix injects `.form-group.no-columns` (column flex, no explicit width) between the user's container and `.widget-combobox`. `flex-grow: 1` distributes height in a column container, not width. Children suppress `min-width` to ~0 via `min-width: 0` on `.widget-combobox-selected-items` (intentional, needed for multiselect tag wrapping)
- Fix: `min-width: 15ch` on `.widget-combobox` — explicit value does not ask children, breaking the collapse chain. `ch` scales with font-size rather than using px

## Changes

- `src/ui/Combobox.scss` — `min-width: 15ch` on `.widget-combobox` root rule
- `src/__tests__/ComboboxStyles.spec.ts` — unit test updated to assert `15ch`
- `e2e/Combobox.spec.js` — E2E regression tests added (pending Studio Pro test page at `/p/combobox-flex-layout`)
- `openspec/changes/fix-combobox-flex-collapse/` — design artifacts with corrected root cause analysis

## Test plan

1. Create a new page in Studio Pro with a container whose CSS class applies `display: flex; flex-flow: row; align-items: center` (e.g. add a class `row-center` and define it in a custom stylesheet).
2. Place a **Combobox** widget (single-select, connected to any enumeration attribute) inside that container.
3. Run the app locally and open the page.
4. **Before the fix**: the combobox is invisible (collapses to ~1px width). **After the fix**: the combobox renders with a visible width (≥ 15ch).
5. Repeat steps 1–4 with a **multi-select** Combobox to confirm the same result.
6. Verify that typing in the combobox input and selecting a value still works normally.

## Override

To restore previous behavior: `.widget-combobox { min-width: 0; }` in app CSS.

### DIFF
--- a/openspec/changes/fix-combobox-flex-collapse/proposal.md
+++ b/openspec/changes/fix-combobox-flex-collapse/proposal.md
@@ -1,0 +1,38 @@
+## Why
+
+Combobox shrinks to 1px when placed in a flex container with `align-items: center` (row). The widget becomes unusable — invisible to the user — without any CSS override from the app developer.
+
+## Root cause
+
+The actual DOM chain in the bug scenario:
+
+```
+.row-center  (user's CSS: display:flex, flex-flow:row, align-items:center)
+  └─ .form-group.no-columns  (Atlas: flex-direction:column, no explicit width)
+       └─ .widget-combobox  (flex-grow:1 — grows height in column container, not width)
+            └─ .widget-combobox-input-container.form-control  (Atlas: min-width:50px — too narrow)
+                 └─ .widget-combobox-selected-items  (min-width:0 — intentional for tag wrapping)
+                      └─ .widget-combobox-input  (max-width:0 unfocused, width:1px multiselect inactive)
+```
+
+The user created `.row-center`. Atlas injected `.form-group.no-columns` (column flex, no explicit width). Inside it, `flex-grow:1` on `.widget-combobox` distributes height, not width. Width is sized by content.
+
+Atlas has `min-width: 0` on `.widget-combobox` and `min-width: 50px` on `.form-control` (`.widget-combobox-input-container`). The `50px` floor propagates up — but it is too narrow: the down-arrow icon alone consumes ~53px, leaving zero space for text input.
+
+**Why native textbox is not affected:**
+
+Textbox renders `.form-control` directly inside `.form-group` — Atlas's `min-width: 50px` is wide enough for a plain input. Combobox `.form-control` must also contain the down-arrow icon, making `50px` insufficient.
+
+## What changes
+
+`packages/pluggableWidgets/combobox-web/src/ui/Combobox.scss` — `.widget-combobox-input-container` rule.
+
+Add `min-width: 15ch` — overrides Atlas's `50px` floor on `.form-control` with a value wide enough to be usable. Scoped to the combobox, not a global Atlas change. `ch` scales with font-size.
+
+## Impact
+
+Must not break:
+
+- Single-select combobox layout in all container types
+- Multiselect combobox layout (active and inactive states)
+- Atlas UI demo site rendering (already unaffected per ticket)
--- a/openspec/changes/fix-combobox-flex-collapse/.openspec.yaml
+++ b/openspec/changes/fix-combobox-flex-collapse/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-05-13
--- a/openspec/changes/fix-combobox-flex-collapse/tests.md
+++ b/openspec/changes/fix-combobox-flex-collapse/tests.md
@@ -1,0 +1,29 @@
+## Tests
+
+- [ ] **Combobox root element has min-width: 15ch in stylesheet**
+    - **Type:** unit
+    - **Given:** Combobox.scss source
+    - **When:** `.widget-combobox` rule is inspected
+    - **Then:** `min-width` is `15ch`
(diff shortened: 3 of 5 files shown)


packages/pluggableWidgets/combobox-web/src/__tests__/ComboboxStyles.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { readFileSync } from "fs";
+import { join } from "path";
+
+const scss = readFileSync(join(__dirname, "../ui/Combobox.scss"), "utf-8");
+
+describe("Combobox SCSS — flex collapse fix (WC-3409)", () => {
+    test(".widget-combobox has min-width: 15ch", () => {
```
r0b1n: Testing what is in the CSS file feels over the top.

openspec/changes/fix-combobox-flex-collapse/proposal.md
```diff
@@ -0,0 +1,40 @@
...
+```
+.row-center  (user's CSS: display:flex, flex-flow:row, align-items:center)
+  └─ .form-group.no-columns  (Mendix-injected: flex-direction:column, no explicit width)
+       └─ .widget-combobox  (flex-grow:1 — grows height in column container, not width)
+            └─ .widget-combobox-input-container  (flex-grow:1, no min-width)
+                 └─ .widget-combobox-selected-items  (min-width:0 — intentional for tag wrapping)
+                      └─ .widget-combobox-input  (max-width:0 unfocused, width:1px multiselect inactive)
+```
+
+The user created `.row-center`. Mendix injected `.form-group.no-columns` as an intermediate wrapper. That wrapper has no explicit width and becomes a column-direction flex container. Inside it, `flex-grow:1` on `.widget-combobox` distributes height, not width. The widget's width is sized by content, and `min-width: min-content` resolves to ~0 because `.widget-combobox-selected-items` has `min-width: 0` (needed for multiselect tag wrapping).
```
r0b1n: How other input widgets, like text box handle this? Maybe atlas has to be fixed instead of it has to be a combination of fixes in two places.

openspec/changes/fix-combobox-flex-collapse/proposal.md
```diff
@@ -0,0 +1,40 @@
+## Why
+
+Combobox shrinks to 1px when placed in a flex container with `align-items: center` (row). The widget becomes unusable — invisible to the user — without any CSS override from the app developer.
+
+## What changes
+
+`packages/pluggableWidgets/combobox-web/src/ui/Combobox.scss` — `.widget-combobox` rule (line 21).
+
+Replace `min-width: min-content` with `min-width: 15ch`.
```
r0b1n: It as `0`, not `min-content` if I see correctly.

openspec/changes/fix-combobox-flex-collapse/proposal.md
```diff
@@ -0,0 +1,40 @@
...
+  └─ .form-group.no-columns  (Mendix-injected: flex-direction:column, no explicit width)
+       └─ .widget-combobox  (flex-grow:1 — grows height in column container, not width)
+            └─ .widget-combobox-input-container  (flex-grow:1, no min-width)
+                 └─ .widget-combobox-selected-items  (min-width:0 — intentional for tag wrapping)
+                      └─ .widget-combobox-input  (max-width:0 unfocused, width:1px multiselect inactive)
+```
+
+The user created `.row-center`. Mendix injected `.form-group.no-columns` as an intermediate wrapper. That wrapper has no explicit width and becomes a column-direction flex container. Inside it, `flex-grow:1` on `.widget-combobox` distributes height, not width. The widget's width is sized by content, and `min-width: min-content` resolves to ~0 because `.widget-combobox-selected-items` has `min-width: 0` (needed for multiselect tag wrapping).
+
+`min-width: 15ch` is an explicit value — it does not ask children, so the children's suppressed minimums are irrelevant.
```
r0b1n: Why 15, and not 5 or 42?

yordan-st: A too small value like 5 or to high like 42 would make it too narrow or wide in most cases. Thought 15 is a general sweet spot, considering the average font size.

packages/pluggableWidgets/combobox-web/src/__tests__/ComboboxStyles.spec.ts
yordan-st: Agreed, redundant

openspec/changes/fix-combobox-flex-collapse/proposal.md
yordan-st: Investigated further. Atlas's min-width: 50px on `.form-control` does reach the combobox but 50px is too small. Therefore it seems collapsed.

What should the correct minimum be in Atlas? My widget-level fix used 15ch (~160px at default font) which felt reasonable for a text input. Happy to implement whatever value makes sense for atlas. just need a steer on the number.

r0b1n: I think we should fix it in the `.form-control` level, override this 50px with a value that is more applicable to combobox. So inputs behave consistently.

r0b1n (approved):

PR #2251, opened 2026-06-05
[WC-3410]: Color Picker: re-dispatch mouseup in capture phase to prevent stuck drag in popups

rahmanunver:
### Pull request type

---

### Description

**Problem:** When the Color Picker widget is placed inside a Mendix popup page, a single click on any drag surface (saturation gradient, hue slider, alpha slider) locks the picker into drag mode — color keeps changing on mouse move without the button held.

**Root cause:** \`react-color@2.19.3\` attaches its drag cleanup listener (\`unbindEventListeners\`) to \`window\` in the **bubble phase**. A Mendix dialog calls \`stopPropagation()\` on \`mouseup\` before it bubbles up to \`window\`, so react-color never receives the release event and the \`mousemove\` listener stays alive indefinitely.

**Fix:** Add a \`useEffect\` in \`components/ColorPicker.tsx\` that listens for \`mouseup\` on \`document\` in the **capture phase** (fires top-down, before any descendant can stop propagation) and re-dispatches it on \`window\`. This ensures react-color's cleanup always runs regardless of dialog interference. The listener is only attached while the picker is visible (\`hidden === false\`) and removed on hide/unmount.

Also adds a \`jest.config.js\` + \`jest.setup.ts\` to suppress pre-existing \`defaultProps\` deprecation warnings emitted by \`react-color\` itself (React 18.3 warnings, unrelated to this fix).

1. Place Color Picker widget inside a Data View on a **popup page**
2. Run app, open popup
3. Click once on saturation/hue/alpha area — release the mouse button
4. Move mouse → color must **not** change
5. Confirm inline Color Picker on a normal page still works correctly
6. Run \`pnpm run test\` in \`packages/pluggableWidgets/color-picker-web\` → 30/30 pass, no console errors

### DIFF
--- a/packages/pluggableWidgets/color-picker-web/src/components/ColorPicker.tsx
+++ b/packages/pluggableWidgets/color-picker-web/src/components/ColorPicker.tsx
@@ -83,12 +83,16 @@ export const ColorPicker = (props: ColorPickerProps): ReactElement => {
         [onColorChange, abortCompleteColorChange]
     );
 
-    const validateColor = (colorValue: string): void => {
-        const message = validateColorFormat(colorValue, format);
-        const validProps = validateProps(props);
-        const alertMessage = message ? invalidFormatMessage?.replaceAll(":colors:", message) : undefined;
-        setAlertMessage(validProps || alertMessage);
-    };
+    const validateColor = useCallback(
+        (colorValue: string): void => {
+            const message = validateColorFormat(colorValue, format);
+            const validProps = validateProps(props);
+            const alertMessage = message ? invalidFormatMessage?.replaceAll(":colors:", message) : undefined;
+            setAlertMessage(validProps || alertMessage);
+        },
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+        [format, invalidFormatMessage]
+    );
 
     const setColorPickerHidden = useCallback(
         (hide: boolean): void => {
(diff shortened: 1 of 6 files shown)


packages/pluggableWidgets/color-picker-web/src/components/ColorPicker.tsx
```diff
@@ -174,6 +178,24 @@ export const ColorPicker = (props: ColorPickerProps): ReactElement => {
...
+
+        // react-color binds its mouseup cleanup to window in the bubble phase.
+        // A Mendix dialog calls stopPropagation on mouseup, preventing it from
+        // reaching window — leaving the picker stuck in drag mode.
+        // Re-dispatching in the capture phase ensures react-color always sees
+        // the release event regardless of dialog interference.
+        const releaseDrag = (): void => {
+            window.dispatchEvent(new MouseEvent("mouseup"));
+        };
+        document.addEventListener("mouseup", releaseDrag, true);
```
iobuhov, line 195: The code here implies that we going to dispatch mouseup up until event is removed.

samuelreichert (approved):

packages/pluggableWidgets/color-picker-web/src/components/ColorPicker.tsx
rahmanunver: Analysis by Claude:

Correct that it re-dispatches on every document mouseup while the picker is open — that's intentional. The listener has to be on document (capture) because the drag release frequently happens outside the widget, and in a popup the dialog's stopPropagation() is exactly what stops the real event reaching window — which is the bug. The re-dispatch is idempotent when no drag is active: react-color's handleMouseUp only removes its mousemove/mouseup listeners, so a synthetic mouseup while idle is a no-op. No recursion either — an event dispatched on window propagates only over [window]; document is a descendant of window, so the capture listener on document is never on that path (verified in jsdom: fires once per real mouseup). The listener is attached only while the picker is visible and removed on hide/unmount, so no leak.

leonardomendix (approved):

leonardomendix (approved):

PR #2257, opened 2026-06-09
[WC-3428] DG2: remove margin from progress bar and pagination bar

yordan-st:
### Pull request type

---

### Description

We removed the 16px margin from both the Progress Bar and Pagination Bar inside Data Grid. The margin was originally on the progress bar, then moved to the pagination bar in a prior fix — but neither element should have this margin. Removing it fully fixes the visual regression caused by the unintended CSS class name match.

### DIFF
--- a/packages/modules/data-widgets/CHANGELOG.md
+++ b/packages/modules/data-widgets/CHANGELOG.md
@@ -6,12 +6,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
-### [Unreleased] Datagrid
-
-#### Fixed
-
-- We fixed an issue where a Progress Bar widget placed as Custom Content inside a Data Grid would not display correctly due to a CSS class name conflict.
-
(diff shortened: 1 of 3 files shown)


packages/modules/data-widgets/CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### [Unreleased] Datagrid
+
+#### Fixed
+
+- We fixed an issue where a Progress Bar widget placed as Custom Content inside a Data Grid would not display correctly due to a CSS class name conflict.
+
```
r0b1n: It should go into datagrid changelog or if you want to keep it here, it should be a changelog of the whole module, see under `## [3.1.0] DataWidgets - 2025-07-24` as an example.

r0b1n (dismissed):

samuelreichert (approved):

iobuhov (approved):

leonardomendix (approved):

PR #2258, opened 2026-06-09
[WC-3452] Add pusher module to the repo

r0b1n:
This adds pusher module to the repo, has old widget as an asses to copy into the module.


packages/modules/pusher/assets/Pusher_widget_legacy_1.2.0.mpk
iobuhov, line 1: Why we need legacy mpk?

r0b1n: The current plan is to let users migrate with less friction, having two widgets side by side, this widget also has `(legacy)` appended to it's name. But this plan might change toward the final stage of course.

gjulivan (dismissed):

iobuhov (dismissed):
LGTM

leonardomendix (approved):

gjulivan (approved):

iobuhov (approved):

PR #2270, opened 2026-06-17
[WC-3464] Rich text Tiptapeditor

gjulivan:
### Pull request type

---


yordan-st:
Would have a look at the high and medium flagged CC review comments

packages/pluggableWidgets/rich-text-web/src/components/toolbars/Toolbar.scss
```diff
@@ -0,0 +1,356 @@
...
+                border-radius: 4px 4px 0 0;
+            }
+
+            &:last-child {
+                border-radius: 0 0 4px 4px;
+            }
+        }
+    }
+
+    button {
```
iobuhov, line 172: Can we use different selector?

iobuhov (commented):
I just looked at Rich Text - looks fine to me, but I did half only. My only concern is drop down and some styling in toolbar.

rahmanunver:
What are we doing with the states? I see store is gone. Is tiptap handling it itself?

gjulivan:
> What are we doing with the states? I see store is gone. Is tiptap handling it itself?

editorcontext is still there, just not on folder

packages/pluggableWidgets/rich-text-web/src/components/toolbars/Toolbar.scss
gjulivan: this selector is `.tiptap-toolbar button`, so, it won't affect any other buttons.

leonardomendix (dismissed):

r0b1n (approved):

yordan-st (approved):

PR #2287, opened 2026-06-22
[WC-3441] DG2: export column enhancements

yordan-st:
### Pull request type

---

### Description

When export type is set to "Default" for attribute columns, the exported Excel cells now use the attribute's own formatting (number decimals/grouping, date pattern) instead of exporting raw values without formatting.

For numbers, the exported cell mirrors what the grid shows: Mendix Decimal attributes do not expose a fixed decimal precision at runtime (only whether digits are grouped), so the format uses up to 8 fractional digits with trailing zeros suppressed. This keeps `1234.56` as `1234.56`, `0.5` as `0.5`, and integers as integers — instead of collapsing to whole numbers.

Also hides the export type and format properties in Studio Pro for dynamic text columns, since they have no effect.

1. Attribute column with exportType "Default" and a Decimal attribute → exported cell is a numeric cell that mirrors the grid (e.g. `1234.56` stays `1234.56`, `0.5` stays `0.5`, integers stay integers; thousands grouping applied when the attribute uses it). Excel format: `#,##0.########` (grouped) or `0.########` (ungrouped).
2. Attribute column with exportType "Default" and a DateTime attribute with custom pattern `dd/MM/yyyy` → exported cell should have Excel format `dd/mm/yyyy`
3. Attribute column with exportType "Number" or "Date" (custom) → still uses the manually specified export format (unchanged behavior)
4. Dynamic text column → Studio Pro should NOT show export type, export number format, or export date format properties
5. Custom content columns → unchanged behavior, all export types still work
6. Existing configurations with no changes → no regression, everything exports as before

### DIFF
--- a/packages/pluggableWidgets/datagrid-web/src/features/data-export/__tests__/cell-readers.spec.ts
+++ b/packages/pluggableWidgets/datagrid-web/src/features/data-export/__tests__/cell-readers.spec.ts
@@ -126,6 +126,36 @@ describe("cell-readers", () => {
             expect(cell.z).toBe("0");
         });
 
+        it("mirrors the grid when number formatter config omits decimalPrecision (grouped)", () => {
+            // Real Mendix Decimal attributes only expose `groupDigits` at runtime.
+            const attr = listAttribute(() => new Big("1234.56")) as any;
+            attr.formatter = { type: "number", config: { groupDigits: true } };
+            const col = column("Amount", c => {
+                c.showContentAs = "attribute";
+                c.attribute = attr;
+                c.exportType = "default";
+            });
+            const cell = readSingleCell(col);
+            expect(cell.t).toBe("n");
+            expect(cell.v).toBe(1234.56);
+            expect(cell.z).toBe("#,##0.########");
+        });
+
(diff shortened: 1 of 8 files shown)


packages/pluggableWidgets/datagrid-web/src/features/data-export/__tests__/cell-readers.spec.ts
```diff
@@ -68,6 +68,63 @@ describe("cell-readers", () => {
...
+            const attr = listAttribute(() => testDate) as any;
+            attr.formatter = { type: "datetime", config: { type: "custom", pattern: "dd/MM/yyyy" } };
+            const col = column("Created", c => {
+                c.showContentAs = "attribute";
+                c.attribute = attr;
+                c.exportType = "default";
+            });
+            const cell = readSingleCell(col);
+            expect(cell.t).toBe("d");
+            expect(cell.v).toEqual(new Date(Date.UTC(2024, 5, 15)));
```
iobuhov, line 96: Let's use same syntax as on line 86 to avoid confusion

iobuhov (dismissed):

leonardomendix (approved):

rahmanunver (approved):

PR #2295, opened 2026-06-29
feat(rich-text): store Quill Delta alongside HTML

Raymond0212:
### Pull request type

---

### Description

This PR adds optional Quill Delta JSON persistence to the Rich Text web widget.

We have seen serious discrepancies when persisted HTML is rendered back into editable Rich Text content, especially for content that contains **tabs or formatting** that does not round-trip cleanly through semantic HTML. Customers also have **strong requirements for reliable diff** views and side-by-side content comparison.

To support those use cases, this change keeps the existing HTML persistence unchanged and adds an optional secondary storage path for the native Quill document model:

- Existing value attribute continues to store semantic HTML.
- New **Enable delta** setting enables Delta persistence (please refer to the attached screenshot).
- New **Delta attribute** stores `JSON.stringify(quill.getContents())` in a separate string attribute. (please refer to the attached screenshot).
- Delta attribute is hidden unless Delta persistence is enabled. (please refer to the attached screenshot).
- Studio Pro validation reports a configuration error when Delta persistence is enabled without selecting a Delta attribute.
- `onChange` still runs once per logical editor update.
- `onLeave` change detection uses Delta snapshots when Delta persistence is enabled, so formatting-only changes can be detected.
- Existing editor loading remains HTML-based for backward compatibility.

The stored Delta JSON is intended to **support future rendering** paths using Quill Delta directly, and allow users to write **custom JavaScript-based diff** and side-by-side comparison features.

- Verify existing HTML persistence still works when **Enable delta** is disabled.
- Verify **Delta attribute** is hidden by default.
- Enable **Enable delta** and verify **Delta attribute** becomes visible.
- Verify Studio Pro shows a configuration error when Delta persistence is enabled but no Delta attribute is selected.
- Select a string attribute for Delta persistence and edit content in the widget.
- Confirm the existing value attribute stores semantic HTML.
- Confirm the Delta attribute stores valid JSON with an `ops` array.
- Verify `onChangeType = onDataChange` still triggers once per debounced editor update.
- Verify `onChangeType = onLeave` detects formatting-only changes when Delta persistence is enabled.

Validation performed:

- `pnpm lint` in `packages/pluggableWidgets/rich-text-web`
- `pnpm test` in `packages/pluggableWidgets/rich-text-web`
- `pnpm build` in `packages/pluggableWidgets/rich-text-web`

E2E test perform on local app with correct parsable Quill delta json persisted in DB.

image

### DIFF
--- a/packages/pluggableWidgets/rich-text-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/rich-text-web/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Added
+
+- We added an optional Delta persistence setting that stores raw Quill Delta JSON in a separate string attribute while keeping the existing HTML value attribute unchanged.
+
 ## [4.12.0] - 2026-04-22
 
 ### Added
--- a/packages/pluggableWidgets/rich-text-web/src/RichText.editorConfig.ts
+++ b/packages/pluggableWidgets/rich-text-web/src/RichText.editorConfig.ts
@@ -1,4 +1,4 @@
-import { hidePropertiesIn, hidePropertyIn, Properties } from "@mendix/pluggable-widgets-tools";
+import { hidePropertiesIn, hidePropertyIn, Problem, Properties } from "@mendix/pluggable-widgets-tools";
 import {
     container,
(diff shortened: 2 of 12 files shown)


gjulivan:
We are not going this way because this will cause problems when engine changes.

Please check on
https://github.com/mendix/web-widgets/pull/2270

PR #2308, opened 2026-07-06
[CRAFTING] Add Page Object Model for datagrid-web's E2E specs

leonardomendix:
### Pull request type

---

### Description

This PR introduced a Page Object Model for datagrid-web's E2E specs, rewrote all six specs on it, and documented POM conventions in the e2e guidelines.


packages/pluggableWidgets/datagrid-web/e2e/pages/DataGridPage.js
```diff
@@ -0,0 +1,119 @@
...
+    constructor(page, name = "datagrid1") {
+        this.page = page;
+        this.name = name;
+        this.root = page.locator(`.mx-name-${name}`);
+    }
+
+    /** Navigate to a page; the fixture auto-waits for Mendix readiness. */
+    async open(path = "/") {
+        await this.page.goto(path);
+    }
```
r0b1n: I beleive this is overkill, I would rather wrap a datagrid, then we can use for example two datagrids on a page.

leonardomendix: Done! ✅

PR #2310, opened 2026-07-06
[WC-3348] Charts: restore highlighted JSON editor in playground

yordan-st:
### Pull request type

---

### Description

Charts v6.3.0 shipped the playground editor on CodeMirror, which threw a bundling error and broke the playground. As a stopgap (to unblock WC-3345), CodeMirror was replaced with a plain `<textarea>` in 6.3.1 — losing syntax highlighting and any feedback on invalid JSON.

This restores a real editor without reintroducing a CodeMirror-class bundling dependency. `CodeEditor` now uses `react-simple-code-editor` + `highlight.js` (JSON), adapting the lightweight pattern already used in `rich-text-web`. It adds inline invalid-JSON feedback (a sticky error banner above the editor) and keeps the existing prop contract (`value` / `onChange` / `readOnly` / `height`) so both playground panels work unchanged.

Scope is intentionally limited to the editor component. Leo's earlier review findings on the already-merged custom-chart / playground rework are tracked separately in **WC-3488**.

- Added deps: `react-simple-code-editor`, `highlight.js` (tree-shaken core + JSON language only)
- New unit tests for `CodeEditor` (highlighting, JSON lint, readOnly, height, no-CodeMirror guard)

Put a chart (e.g. Line chart) on a page, set **Show playground slot = Yes**, drop the **Chart playground** widget into the slot, configure a Series, and run the app (F5). Click **Toggle Editor** to open the sidebar.

1. **Syntax highlighting** — the JSON in the editable "Custom settings" panel is colored (keys/strings/numbers), not plain text.
2. **Invalid JSON feedback** — break the JSON (e.g. delete a `}`); a red error banner appears at the top of the editor. Fix it → the banner disappears.
3. **Live chart update** — with the dropdown on **Layout**, set `{ "title": { "text": "TEST" } }`; the chart title updates live.
4. **Read-only panel** — the lower "Settings from the Studio Pro" panel is not editable.
5. **Keyboard** — Tab inserts spaces inside the editor; Esc then Tab moves focus out.
6. **No regression** — no CodeMirror bundling error in the browser console; the playground loads.


iobuhov (changes requested):
- Let's move the opensepc to "charts" package
- Make sure we don't touch maps

iobuhov (dismissed):
LGTM

leonardomendix (dismissed):

gjulivan (approved):

leonardomendix (approved):

PR #2314, opened 2026-07-07
[WC-3488]: Charts rework & cleanup

yordan-st:
### Pull request type

---

### Description

**WC-3488** — cleanup of the unresolved code review (Leo) on the already-merged custom-chart / playground editor rework. All findings shipped unfixed and were still live on `main`; this PR resolves them.

**What & why**, per finding:

- **[Critical] `computed().get()` misuse** — `useCustomChart.ts` wrapped a plain object in `computed(() => ({...})).get()`, allocating a throwaway MobX atom every render with no caching or reactivity benefit. Replaced with a plain object literal; reactivity is handled by the existing `observer` wrapper on `CustomChart.tsx`.
- **[Critical] `containerStyle` dead code** — the hook returned a `containerStyle` never consumed (`CustomChart.tsx` computes its own via `constructWrapperStyle`). Removed the computation, the return field, and the unused import.
- **[Important] hardcoded `layoutOptions: {}` / `configOptions: {}`** — this silently blanked the playground's **Modeler Layout** and **Modeler Configuration** panels. Now passes the adapter's real `layout` / `config`. (User-visible → changelog entry added.)
- **[Important] `store.data as Data[]` unsound cast** — replaced the bare cast with a single named boundary helper `toPlotlyData()`. No runtime validation added: the store already guards on write (`setDataAt` parses, type-checks, warns), and Plotly's `Data` union is impractical to validate exhaustively — the helper just localizes and documents the unavoidable conversion.
- **[Important] `@types/jest` in `dependencies`** — moved to `devDependencies` and dropped `"types": ["jest"]` from `tsconfig.build.json` so jest types no longer ship as a runtime dep / pollute the production build.
- **[Important] deleted 402-line test, no replacement** — added a unit spec for `EditableChartStore` (8 cases: reset, `setDataAt` valid/out-of-range/invalid-JSON/non-object, `setLayout`/`setConfig` null-guard, JSON getters).
- **[Minor] undocumented `eslint-disable react-hooks/set-state-in-effect`** — documented why it's safe and the accepted risk.
- **[Minor] duplicate `prettifyJson`** — extracted to one shared helper, imported in both editor controllers.
- **[Minor] `onViewSelectChange` not memoized** — wrapped in `useCallback` in both controllers.
- **[Minor] `key` duplicated (React state + MobX box)** — collapsed to a single source of truth in `useV2EditorController` (kept the React state that drives render; the reaction now re-subscribes on the key via its effect deps with `fireImmediately`).

**Deliberately out of scope (deferred to WC-3348):**

- The silent empty-JSON `catch {}` blocks in the editor controllers (Leo's last Minor). WC-3488 marks this as WC-3348's scope; left untouched here.
- `CodeEditor.tsx` and the editor restore — that was WC-3348 (#2310).

**Notes for reviewers:** the OpenSpec change (`custom-chart-web/openspec/changes/fix-rework-cleanup/`) has the full proposal / test plan / task breakdown. Two test-infra additions (zero product impact): a `^src/` `moduleNameMapper` in `custom-chart-web/jest.config.js` (mirrors the tsconfig `baseUrl` the source already uses) and a `stubObjectURL` shim (the shared-charts barrel loads Plotly, which touches `URL.createObjectURL` on import in the test env).

---

**Automated (green locally):**

- `custom-chart-web` — 33 tests · `chart-playground-web` — 6 tests · `shared/charts` — 45 tests
- Lint clean on all changed files; `shared/charts` production build passes after the B5 tsconfig change.

**Manual QA — Modeler panels (the user-visible fix):**

1. Add a **Custom chart** widget to a page; give it static/sample data (e.g. `[{"type":"bar","x":[1,2],"y":[3,4]}]`) and enable the playground slot. Optionally set a Layout value like `{"title":"Test"}`.
2. Run the app, hard-reload the browser (widget JS is cached — Empty Cache and Hard Reload).
3. Open the chart playground → in the sidebar view selector pick **Layout**, then **Configuration**.
4. **Expected:** the **Modeler Layout** / **Modeler Configuration** panels show real JSON (width/height/autosize/font/margin, plus any layout you set) — **not** `{}`.

**Regression checks:**

- Custom chart still renders and click events still fire.
- In the playground, switch between Layout / a trace / Configuration — the editor content follows the selection with no stale/desynced text.
- Edit a value in the editor → chart updates; feed invalid JSON → chart doesn't crash (invalid input is ignored — this catch remains silent by WC-3348 scope decision).

### DIFF
--- a/packages/pluggableWidgets/chart-playground-web/src/helpers/__tests__/editorJson.spec.ts
+++ b/packages/pluggableWidgets/chart-playground-web/src/helpers/__tests__/editorJson.spec.ts
@@ -1,0 +1,11 @@
+import { prettifyJson } from "../editorJson";
+
+describe("prettifyJson", () => {
+    it("formats valid JSON with two-space indentation", () => {
+        expect(prettifyJson('{"a":1}')).toBe('{\n  "a": 1\n}');
+    });
+
+    it("returns an error object string for invalid JSON", () => {
+        expect(prettifyJson("{ not json")).toBe('{ "error": "invalid JSON" }');
+    });
+});
--- a/packages/pluggableWidgets/chart-playground-web/src/helpers/__tests__/stubObjectURL.ts
+++ b/packages/pluggableWidgets/chart-playground-web/src/helpers/__tests__/stubObjectURL.ts
@@ -1,0 +1,7 @@
+// shared-charts' main barrel imports plotly, which calls URL.createObjectURL at load time.
+// jsdom/happy-dom doesn't implement it. Import this module BEFORE the barrel to stub it.
+if (typeof window.URL.createObjectURL !== "function") {
+    window.URL.createObjectURL = () => "";
+}
+
+export {};
--- a/packages/pluggableWidgets/chart-playground-web/src/helpers/__tests__/useComposedEditorController.spec.ts
+++ b/packages/pluggableWidgets/chart-playground-web/src/helpers/__tests__/useComposedEditorController.spec.ts
@@ -1,7 +1,4 @@
-jest.mock("plotly.js-dist-min", () => ({}));
-jest.mock("react-plotly.js", () => jest.fn(() => null));
-jest.mock("react-plotly.js/factory", () => jest.fn(() => jest.fn(() => null)));
-
+import "./stubObjectURL";
 import { act, renderHook } from "@testing-library/react";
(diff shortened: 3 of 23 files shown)


iobuhov (dismissed):
LGTM

leonardomendix (dismissed):

iobuhov (approved):

rahmanunver (approved):

PR #2321, opened 2026-07-13
MobX crop store for Image Cropper

rahmanunver:
### Pull request type

---

### Description

Stacked on top of the Image Cropper PR (`feat/image-cropper-polish-rotate-bw`) — base is set to that branch, so this diff shows **only** the MobX refactor, not the base PR's changes.

**Why.** Review feedback flagged that the widget's state had become too coupled to React. `ImageCropperContainer` carried five state-mirroring refs (`committedCropRef`, `zoomRef`, `grayscaleRef`, `liveCropRef`, `zoomAnchorRef`) plus two callback mirrors (`markInternalRef`, `showPreviewRef`). They existed only because `applyCrop` was a stable-identity `useCallback` (kept stable for the 400ms auto-apply debouncer) and therefore couldn't read current state through its stale closure. A MobX store method reads `this.zoom` directly, so the refs are no longer needed.

**Moved into `ImageCropperStore` (MobX):** the crop-domain working state — `liveCrop`, `committedCrop`, `zoomAnchor` (`observable.ref`), `zoom`, `grayscale` (`observable`) — plus `aspect` (`computed`) and the crop operations as actions (`setZoom`, `commitCrop`, `toggleGrayscale`, `rotate`, `reset`, `initFromImageLoad`, `onImageChanged`). The store also owns the debounce + auto-apply gate (absorbing `useAutoApplyCrop`). Crop ops are now plain calls like `store.setZoom(v)` / `store.rotate(90)`.

**Stayed normal React state (imperative / DOM-bound):** the `imageRef` DOM node, `usePreviewSrc` (blob object-URL lifecycle), `useOriginalImage` (fetch + unmount cleanup + internal-change flag), and `CropArea`'s local DOM-measurement state. These are injected into the store via `setDeps`. Config props are read live through the mobx-kit gate, not copied into observables.

**Controlled/uncontrolled unchanged.** The only externally controlled value is `props.image` (Mendix `EditableValue`); zoom/rotation/grayscale/crop are internal. Inbound image changes sync through one `uri`-keyed effect calling `store.onImageChanged()`; outbound writes go `markInternalChange()` → `setValue()`, with the internal-change flag absorbing the resulting `uri` change so there's no feedback loop. No reactions used for prop syncing.

**Preview stays an async action, not a computed** — `cropImage`/`rotateImage` are async canvas operations reading a live DOM ``, so they can't be a pure synchronous computed.

**Removed:** the 5 mirror refs + 2 callback mirrors, and the `useAutoApplyCrop` / `useImageCropperState` hooks (subsumed by the store). `ImageCropperContainer` is now an `observer` wired via the shared `widget-plugin-mobx-kit` gate pattern (`GateProvider` + `useConst` + `useSetup`), matching `custom-chart-web`.

**Public API unchanged** — no XML/props changes; widget behavior is identical.


leonardomendix (approved):

packages/pluggableWidgets/image-cropper-web/src/components/ImageCropperContainer.tsx
```diff
@@ -293,41 +76,37 @@ export function ImageCropperContainer(props: ImageCropperContainerProps): ReactE
...
                 boundaryHeight={props.boundaryHeight}
-                onImageLoad={handleImageLoad}
-                zoom={state.zoom}
+                onImageLoad={(percentCrop, pixelCrop) => store.initFromImageLoad(percentCrop, pixelCrop)}
+                zoom={store.zoom}
                 minZoom={Number(props.minZoom)}
                 maxZoom={Number(props.maxZoom)}
-                setZoom={handleZoomChange}
-                zoomAnchor={zoomAnchor}
+                setZoom={next => store.setZoom(typeof next === "function" ? next(store.zoom) : next)}
```
r0b1n: Does this prop have to be so complicated? Maybe we can simplify it down the road

packages/pluggableWidgets/image-cropper-web/src/components/CropArea.tsx
r0b1n, line 65: Should those calculations also be part of mobx?

r0b1n, line 58: And those

packages/pluggableWidgets/image-cropper-web/src/components/ImageCropperContainer.tsx
```diff
@@ -338,12 +117,12 @@ export function ImageCropperContainer(props: ImageCropperContainerProps): ReactE
...
-                onZoomChange={handleZoomChange}
-                onRotateLeft={() => handleRotate(-90)}
-                onRotateRight={() => handleRotate(90)}
-                onToggleGrayscale={handleToggleGrayscale}
-                onReset={handleReset}
+                onZoomChange={zoom => store.setZoom(zoom)}
+                onRotateLeft={() => store.rotate(-90)}
+                onRotateRight={() => store.rotate(90)}
+                onToggleGrayscale={() => store.toggleGrayscale()}
+                onReset={() => store.reset()}
```
r0b1n, line 124: Maybe we can pass down the store instead of mapping every prop and method.

packages/pluggableWidgets/image-cropper-web/src/components/ImageCropperContainer.tsx
rahmanunver: Changed, moved the updater resolution into the store.

packages/pluggableWidgets/image-cropper-web/src/components/CropArea.tsx
rahmanunver: Intentionally view-local. displaySize comes from the loaded  and feeds the  sizing + the stray-click guard (isStrayCrop) before crops reach the store, so the store never needs it. CropArea is also rendered store-free by the editor preview, so this stays reusable without a store.

rahmanunver: same as previous

packages/pluggableWidgets/image-cropper-web/src/components/ImageCropperContainer.tsx
rahmanunver: Considered it. Doable for CropToolbar, but it'd couple the toolbar to the store (as an observer), labels still come from props. Container wiring is intentional, and CropArea can't take a store anyway (editor preview has none). I'm leaving the mapping.

r0b1n (approved):

PR #2332, opened 2026-07-16
[WC-3501]: fix(combobox-web): fix inverted editability logic

r0b1n:
## Summary

- Fixed a bug in `getReadonly()` where the `conditionally` editability mode had inverted logic: `expression.value = true` (meaning "editable") was being returned directly as the read-only flag, making the widget read-only when it should be editable, and vice versa.
- Fixed a related bug where the presence of an editable `targetAttribute` caused the function to return early with `false` (editable), bypassing the custom editability condition entirely. Now `targetAttribute` only short-circuits when it is explicitly read-only.
- Added unit tests covering all branches of `getReadonly`.

## Root cause

Two issues in `getReadonly`:
1. The original early-return on `targetAttribute` presence meant any editable attribute would always report the widget as editable, ignoring the `never` or `conditionally` settings.
2. The `conditionally` branch returned `customEditabilityExpression.value ?? true` directly, where the expression represents *editability* (true = editable), so the boolean was inverted.

### DIFF
--- a/packages/pluggableWidgets/combobox-web/CHANGELOG.md
+++ b/packages/pluggableWidgets/combobox-web/CHANGELOG.md
@@ -6,6 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+- We fixed an issue where the custom editability condition (`never` or `conditionally`) was ignored when the combobox attribute was present and editable.
+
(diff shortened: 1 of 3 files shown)


yordan-st (dismissed):
LGTM

leonardomendix (dismissed):

leonardomendix (approved):

yordan-st (approved):

PR #2333, opened 2026-07-16
feat(image-cropper-web): allow custom aspect ratio width/height as expressions

hunter-koppen:
Convert the customAspectWidth/customAspectHeight properties from static integers to Integer-returning expressions so mendix developers can bind an attribute (or any expression) to the ratio sides instead of a fixed value.

The store now reads each DynamicValue<Big> via ValueStatus.Available guard, resolveAspectRatio tolerates undefined, and the editor preview parses the expression text (numeric literals only) with a free-aspect fallback.


leonardomendix:
Hello maker! ☺️
  
  Thank you for opening a pull request to the \`web-widgets\` repository and helping us make our widgets better. 💙
  
  A small note from the maintainers:
  
  We know that you probably wish to see your cool new feature or fix ASAP, but bringing a new feature to the widget is a small journey. But don't be scared! We just need time to triage, test, and analyze this feature or fix. We may request XML, code, or other changes that may be needed to match our desired look and feel for the new feature. We, as maintainers, strive to make sure that every change made in the code doesn't break existing behavior.
  
  For now, here is what will happen next:
  
  1. Our TotW\* will quickly scan your code and, if all is well, trigger a PR check.
  2. We will discuss your new feature or fix within our team.
  3. If we decide to add it, we will do code review and testing. As mentioned, we may ask you to make some changes.
  4. If all goes well, we will publish this new feature or fix in the next minor or patch release.
  
  IMPORTANT: This repository uses code checks (linters, tests, etc.). Someone from our team will trigger these checks for your PR. In short, we will not proceed with your PR until all checks are green. It's the PR owner's goal to make checks green; maintainers are not going to do it for you.
  
  \* TotW — Tech of the Week

rahmanunver:
Hi @hunter-koppen, thanks for this! I'm taking it over to get it moving: handling some edge cases around your change, plus a related fix and some chores.

Pushing directly to this branch rather than opening a new PR so the contribution stays attributed to you.

packages/pluggableWidgets/image-cropper-web/openspec/changes/image-cropper-aspect-expression/design.md
```diff
@@ -0,0 +1,66 @@
...
+
+When the resolved ratio changes, rebuild the crop box in one step (`buildInitialCrop`) rather than letting `ReactCrop` interpolate. Guard the auto-apply gate so a ratio change alone does not push a wrong-ratio image to the bound attribute — seeding is programmatic and must remain "disarmed" (the store already distinguishes user-driven commits from programmatic ones via `userDragged` / `armed()`).
+
+### Decision 3: Editor preview parses numeric literals only
+
+The editor has no runtime data — only expression _text_. `toNumber` parses a numeric literal and falls back to `undefined` (free aspect) otherwise. This is display-only and already implemented in the PR; the spec pins it so it isn't regressed.
+
+### Decision 4: Raise `minimumMXVersion` to 11.12
+
+Expression-typed properties with `returnType Integer` bound to attributes are the supported baseline. Bump `marketplace.minimumMXVersion` from `10.21.0` to `11.12`.
```
r0b1n: It is already 11.12. If it is not set correctly it should be corrected. This is not part of the story.

packages/pluggableWidgets/image-cropper-web/openspec/changes/image-cropper-aspect-expression/specs/image-cropper/spec.md
```diff
@@ -0,0 +1,91 @@
...
+#### Scenario: Custom ratio
+
+- **WHEN** `aspectRatio` is `custom` and both `customAspectWidth` and `customAspectHeight` resolve to values greater than 0
+- **THEN** the crop selection SHALL be locked to `customAspectWidth / customAspectHeight`
+- **AND** if either resolved value is not greater than 0, the crop SHALL fall back to free-form
+
+#### Scenario: Custom ratio expression not yet resolved
+
+- **WHEN** `aspectRatio` is `custom` and either side is not yet Available
+- **THEN** the resolved ratio SHALL be treated as not yet known, which is distinct from free-form
```
r0b1n, line 29: I miss explicit scenario definition when properties never resolve to a real value, for example they are `unavailable` from the very first render and it stays this way.

packages/pluggableWidgets/image-cropper-web/src/stores/ImageCropperStore.ts
```diff
@@ -124,7 +136,29 @@ export class ImageCropperStore implements SetupComponent {
...
+     * which is why an unguarded seed picks free aspect and then visibly jumps once the
+     * expression resolves. Only custom mode can be pending; every preset is synchronous.
+     */
+    get aspectReady(): boolean {
+        if (this.props.aspectRatio !== "custom") {
+            return true;
+        }
+        const isAvailable = (p: DynamicValue<Big>): boolean => p.status === ValueStatus.Available;
+        return isAvailable(this.props.customAspectWidth) && isAvailable(this.props.customAspectHeight);
     }
```
r0b1n, line 174: I think we should make aspect return something more meaningful - `undefined` - not ready; `-1` - ready, free; positive number - ready, aspect set. Then this check doesn't have to rely on the props, it has to rely on simple fact that: not undefined means ready.

packages/pluggableWidgets/image-cropper-web/src/stores/ImageCropperStore.ts
```diff
@@ -124,7 +136,29 @@ export class ImageCropperStore implements SetupComponent {
...
 
     get aspect(): number | undefined {
-        return resolveAspectRatio(this.props.aspectRatio, this.props.customAspectWidth, this.props.customAspectHeight);
+        const toNumber = (p: DynamicValue<Big>): number | undefined =>
+            p.status === ValueStatus.Available && p.value ? p.value.toNumber() : undefined;
+        return resolveAspectRatio(
+            this.props.aspectRatio,
+            toNumber(this.props.customAspectWidth),
+            toNumber(this.props.customAspectHeight)
+        );
```
r0b1n: It is cleaner to keep reading the value even if it is not strictly availalbe, if value goes to loading it retain the old value for a while and then replaces with the new value. 
1. (Loading, undefined) // first render
1. (Available, 5) // loading done
1. (Loading, 5) // something changed under the hood, but old value retained
1. (Available, 10) // new value arrived

packages/pluggableWidgets/image-cropper-web/CHANGELOG.md
```diff
@@ -8,4 +8,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+
+### Fixed
+
+- The crop selection no longer briefly appears at the wrong ratio while a custom aspect ratio expression is still loading.
+- The crop selection is no longer cleared and the image no longer reloaded when unrelated properties on the page change.
+
+## [1.0.0] - 2026-07-16
+
+### Added
+
```
r0b1n: Version header should not be added

packages/pluggableWidgets/image-cropper-web/openspec/changes/image-cropper-aspect-expression/design.md
rahmanunver: It is set to 11.12 on the marketplace for the initial release, but it was shipped with a lower version in package.json, this aligns that in this PR.

packages/pluggableWidgets/image-cropper-web/CHANGELOG.md
rahmanunver: Like the version, the changelog was also not aligned, the v1 is released, shouldn't this be here?

packages/pluggableWidgets/image-cropper-web/openspec/changes/image-cropper-aspect-expression/design.md
r0b1n: This block implies that the bump was done in order to implement this story. It is wrong. The bump is a fix of an oversight, it should not be mentioned as a requirement for this change document.

packages/pluggableWidgets/image-cropper-web/CHANGELOG.md
r0b1n: Version should be a added in chanegelogs PR, not in this PR. And I see there is a version header in main already, so this PR have to be rebased most likely

leonardomendix (dismissed):

r0b1n (approved):

leonardomendix (approved):

PR #2354, opened 2026-07-23
[WC-3471]: fix(image-web): migrate image source properties to allowUpload

yordan-st:
### Pull request type

---

### Description
`imageObject` and `defaultImageDynamic` in the Image widget were declared as `type="image"` without `allowUpload`, which Mendix has deprecated — `DynamicValue<WebImage>` is removed in Mendix 12, and Studio Pro 11.8+ already shows a design-time deprecation warning for this pattern.

This adds `allowUpload="true"` to both properties and regenerates typings accordingly (`EditableImageValue<WebImage>`). The widget's read logic in `Image.tsx` is unchanged — `EditableImageValue` exposes the same `status`/`value.uri` shape the widget already consumes, so this is a display-only migration with no new upload UI.

`marketplace.minimumMXVersion` is bumped to `11.12.0` since the widget XML now requires `allowUpload` support, which older Studio Pro versions cannot parse (verified: Studio Pro <11.8 fails to load the widget package with "The 'allowUpload' attribute is not declared").

Manually verified in Studio Pro 11.12 that both Static and Dynamic image-source configuration remain available in the property editor after this change — existing apps using a static image are not broken.

Widget version is not bumped in this PR (bumped at release time per repo convention); this change will require a major bump (2.0.0) on release due to the `minimumMXVersion` increase.

- Existing Image widget configurations (static image, dynamic/attribute-bound image, icon, image URL datasources) continue to render identically.
- No upload/file-picker UI appears anywhere on the widget.
- Widget can be added to a page in Studio Pro 11.12+; "Image source" and "Default image" properties still offer both Static and Dynamic configuration.


packages/pluggableWidgets/image-web/src/__tests__/Image.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+        status: ValueStatus.Available,
+        value: { uri: "https://example.com/a.png", name: "a.png" },
+        readOnly: false,
+        validation: undefined,
+        setValidator: jest.fn(),
+        setValue: jest.fn(),
+        setThumbnailSize: jest.fn(),
+        ...overrides
+    } as ImageProp;
+}
```
r0b1n: We should extent existing test utils instead

r0b1n (dismissed):

PR #2372, opened 2026-08-06
[WC-3520]: Gallery recover from stale sort-order attribute id

yordan-st:
### Pull request type

---

### Description

Gallery crashed with an uncaught MobX reaction error — `Sort order item: invalid attribute id '<id>'` — when a sort order restored from `datasource.sortOrder` referenced an attribute id no longer valid in the current app build. Mendix attribute ids are per-build tokens regenerated on redeploy, so a sort order persisted per-user (personalization attribute / DB storage) can outlive the ids it references.

`QueryParamsService` now guards the forwarding of sort order into `ListValue.setSortOrder()`: on a runtime rejection it catches the error, falls back to the default (unsorted) order, and emits a `console.warn` so the reset is diagnosable. The guard sits on the shared forwarding path (the sort reaction), so it protects the widget whether or not a sort widget is configured — the `SortOrderStore` is not instantiated when no DropdownSort exists, which is why the fix is not there.

Ticket: WC-3520

1. Configure a Gallery with sort personalization stored in a per-user attribute (DB storage).
2. Sort the gallery, then redeploy the app so attribute ids regenerate (stale stored sort order).
3. Re-run and open the page as the same user.
   - Before fix: widget crashes with `invalid attribute id` uncaught reaction error.
   - After fix: widget renders, falls back to default sort order, `console.warn` logged.
4. Confirm the normal case (all-valid sort ids) still applies sort order unchanged, preserving order and direction.

### DIFF
--- a/packages/pluggableWidgets/gallery-web/openspec/changes/fix-stale-sort-order-attribute/specs/gallery-sort-order-validation/spec.md
+++ b/packages/pluggableWidgets/gallery-web/openspec/changes/fix-stale-sort-order-attribute/specs/gallery-sort-order-validation/spec.md
@@ -1,0 +1,31 @@
+## ADDED Requirements
+
+### Requirement: Applying a restored sort order never surfaces an uncaught reaction error
+
+When the Gallery widget forwards a restored (persisted or props-provided) sort order to the datasource, a rejection by the Mendix runtime — such as a sort instruction whose attribute id is no longer valid for the current app build — SHALL NOT escape as an uncaught MobX reaction exception. The widget SHALL catch the rejection and recover to the datasource's default sort order so it remains usable.
+
+This guard SHALL apply on the path every sort order takes into the datasource, regardless of whether a sort widget (DropdownSort) is configured — i.e. it SHALL NOT depend on a `SortOrderStore` being instantiated.
+
+#### Scenario: Restored sort order contains a stale attribute id
+
+- **WHEN** the widget is initialized with a `datasource.sortOrder` restored from a previous session/run that references an attribute id no longer present in the current app build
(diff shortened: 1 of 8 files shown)


iobuhov (approved):
LGTM

packages/pluggableWidgets/gallery-web/openspec/changes/fix-stale-sort-order-attribute/specs/gallery-sort-order-validation/spec.md
```diff
@@ -0,0 +1,31 @@
+## ADDED Requirements
+
+### Requirement: Applying a restored sort order never surfaces an uncaught reaction error
+
+When the Gallery widget forwards a restored (persisted or props-provided) sort order to the datasource, a rejection by the Mendix runtime — such as a sort instruction whose attribute id is no longer valid for the current app build — SHALL NOT escape as an uncaught MobX reaction exception. The widget SHALL catch the rejection and recover to the datasource's default sort order so it remains usable.
+
+This guard SHALL apply on the path every sort order takes into the datasource, regardless of whether a sort widget (DropdownSort) is configured — i.e. it SHALL NOT depend on a `SortOrderStore` being instantiated.
+
+#### Scenario: Restored sort order contains a stale attribute id
```
r0b1n, line 9: Is it really the case that framework is supplying us with outdated sorting information? We have to let the framework team know so they can fix it us well. Can you confirm this is the case and create a trivial reproduction code for it?

yordan-st: Confirmed. Stack trace shows the invalid attribute id comes from the framework's own datasource-property object. We need to create ticket and point the to the original ticket, it has a testproject included.

leonardomendix (approved):

PR #2381, opened 2026-08-13
[WC-3466] chore(datagrid-date-filter-web): migrate react-datepicker to v9 (WC-3466)

iobuhov:
## What & why

Upgrades the Date filter widget's calendar from `react-datepicker@6` to `react-datepicker@9` (and `date-fns@3 → 4`), resolving **WC-3466**. v9 is a 3-major jump: Popper.js is replaced by Floating UI, the calendar popup is now a `role="dialog" aria-modal="true"`, and day-name columns expose full weekday accessible names.

## Changes

- **`chore`: migrate react-datepicker to v9 with date-fns 4** — bumps `react-datepicker` and `date-fns`, drops the now-unnecessary `@types/react-datepicker` (v9 ships its own types), widens the `widget-plugin-filtering` `date-fns` peer to `^3.6.0 || ^4.1.0`, and adapts the component to v9's typings:
  - `DatePicker.tsx` — new discriminated `PickerChangeHandler`, `selectsRange` narrowed so the props union resolves to its range arm, drops no-op props.
  - `DatePickerController.ts` / `useSetup.ts` — retyped against v9's `DatePickerProps`.
  - Unit tests updated + new `DatePickerController.spec.ts`.
  - CHANGELOG entry for the day-name screen-reader improvement.
- **`test`: fix flaky `filter_init_condition` e2e** — the two "apply filter right after load" tests used a racy `goto("/#/…", {timeout:1000})` + `reload()` that intermittently landed on the home page under the shared worker session. Switched to the resolved path `goto("/p/filter_init_condition")` + `waitForMendixApp`, matching the text/number filter specs. (Pre-existing flake, not caused by the migration.)
- **`docs`: update migration tasks** — OpenSpec task bookkeeping.

## Testing

- `pnpm run test` — unit suite green (68 tests, snapshots unchanged).
- `pnpm turbo build` + `pnpm run verify` — pass.
- e2e against a local Studio Pro app: all non-screenshot tests pass reliably, serial and parallel (previously the init-condition tests failed consistently in the full suite).

## Notes / follow-ups

- The open-calendar screenshot baseline (`test.fixme`) and any updated closed-state baseline must be regenerated **in CI** (`chromium-linux`), not on macOS.
- v9 hard-codes the dialog's accessible name ("Choose Date") in English with no override prop — non-English apps get an untranslated dialog label. Worth an upstream issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/shared/widget-plugin-filtering/package.json
```diff
@@ -46,7 +46,7 @@
         "mobx-react-lite": "4.0.7"
     },
     "peerDependencies": {
-        "date-fns": "^3.6.0"
+        "date-fns": "^3.6.0 || ^4.1.0"
```
r0b1n, line 49: We probably don't need 3.6.0 anymore. I see it is gone from package-lock.

r0b1n (approved):

leonardomendix (approved):

PR #2392, opened 2026-08-19
[WC-3442] feat(datagrid-web): add onBeforeExport and onAfterExport event actions

r0b1n:
## Summary

- Adds two optional action properties — **On before export** and **On after export** — to Data Grid 2, giving developers a logging/auditing hook into the export lifecycle
- `onBeforeExport` fires fire-and-forget immediately before the first datasource page fetch; receives `gridName`, `columnTitles`, `chunkSize`, `fileName`, `sheetName`, and `startTime`
- `onAfterExport` fires after the export resolves (success or abort); additionally receives `exportedItemCount`, `status` (`"success"` | `"aborted"`), and `endTime`
- Both actions are optional and fully independent; the export flow is unchanged when neither is configured
- `fileName` and `sheetName` are passed by the external export caller via `exportData()` options and default to empty strings when not provided

## Changes

- `Datagrid.xml` — two new `<property type="action">` blocks with `<actionVariables>` in the Events group
- `DatagridProps.d.ts` — updated manually to match XML (will be regenerated on build)
- `DSExportRequest.ts` — exposed `get loaded()` and `get limit()` public getters (renamed private fields to `_loaded` / `_limit`)
- `ExportController.ts` — added `BeforeExportArgs` / `AfterExportArgs` types; `name` constructor param; `beforeexport` / `afterexport` events in `ControllerEvents`; public `on()` method (returns `Unsubscribe`); emits both events in `exportData()`
- `useDataExport.ts` — stores latest `ActionValue` props in refs; subscribes to `beforeexport` / `afterexport` once per controller lifetime via `useEffect([entry])`; effect cleanup unsubscribes automatically
- `ExportController.spec.ts` — 5 new unit tests covering all event scenarios

## Test plan

- [ ] All existing unit tests pass (`pnpm run test`)
- [ ] No lint errors
- [ ] In Studio Pro: configure `onBeforeExport` and `onAfterExport` on a Data Grid 2, trigger an export, verify both microflows/nanoflows are called with correct variable values
- [ ] Verify `status` is `"aborted"` when the user cancels mid-export
- [ ] Verify export works normally when neither action is configured

### DIFF
--- a/packages/pluggableWidgets/datagrid-web/src/features/data-export/useDataExport.ts
+++ b/packages/pluggableWidgets/datagrid-web/src/features/data-export/useDataExport.ts
@@ -50,7 +50,7 @@ export function useDataExport(
     }, [columnsStore.visibleColumns, entry]);
 
     useEffect(() => {
-        return entry?.controller.on("beforeexport", args => {
+        return entry.controller.on("beforeexport", args => {
             const action = onBeforeExportRef.current;
             if (action?.canExecute) {
                 action.execute({
@@ -66,7 +66,7 @@ export function useDataExport(
     }, [entry]);
 
     useEffect(() => {
-        return entry?.controller.on("afterexport", args => {
+        return entry.controller.on("afterexport", args => {
             const action = onAfterExportRef.current;
             if (action?.canExecute) {
                 action.execute({
--- a/packages/pluggableWidgets/datagrid-web/src/features/data-export/useDataExport.ts
+++ b/packages/pluggableWidgets/datagrid-web/src/features/data-export/useDataExport.ts
@@ -1,5 +1,5 @@
 import { Big } from "big.js";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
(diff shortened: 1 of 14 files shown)


packages/pluggableWidgets/datagrid-web/src/features/data-export/useDataExport.ts
```diff
@@ -44,13 +45,56 @@ export function useDataExport(
         );
     }, [columnsStore.visibleColumns, entry]);
 
+    useEffect(() => {
```
gjulivan, line 52: i felt like this is an antipattern of synchronizing props from useExportData to ExportController.

we already have NanoEvent style of communicatings between events.
and DSExportRequest is emitting loadstart and loadend.
can we subscribe to this somewhere?
seems not quite right to break the nanoevents pattern and sync using useEffect for this.
Am I wrong?

gjulivan (dismissed):

leonardomendix (approved):

gjulivan (approved):

PR #2420, opened 2026-09-10
[WC-3548]: honor Position of pagination for custom pagination

yordan-st:
### Pull request type

**Bug fix** (non-breaking change which fixes an issue)

### Description

Data Grid 2's "Position of pagination" property already controlled the built-in `Pagination` control but had no effect on custom pagination: `WidgetFooter` rendered the custom pagination widgets unconditionally and `WidgetTopBar` never rendered them at all, so "Above grid" silently did nothing. This mirrors the bug fixed for Gallery, which explicitly scoped Data Grid 2 out to keep that PR reviewable — this is that follow-up.

Custom pagination now honors `pagingPosition` in both runtime and editor preview: top bar for "Above grid", footer for "Below grid", and footer-only (not duplicated) for "Both" — duplicating the widgets placeholder would duplicate widget instances, DOM ids, and state. `pagingPosition` stays visible in Studio Pro when custom pagination is enabled (previously hidden as a stopgap), and a new design-time warning explains the single-render behavior for "Both".

- Add a Data Grid 2, enable Custom pagination, place a widget in the drop zone.
- Set "Position of pagination" to "Above grid" → widgets render above the grid, both at runtime and in the Studio Pro page editor preview.
- Set it to "Below grid" → widgets render below the grid (unchanged from before).
- Set it to "Both" → widgets render once, below the grid, and Studio Pro shows a warning on the "Position of pagination" property.
- Confirm "Position of pagination" stays visible/editable in the properties panel while Custom pagination is enabled.
- Regression: built-in pagination (Custom pagination off) still respects top/bottom/both as before.

No XML changes, no docs PR needed. Changelog entry included in this branch.

Ticket: WC-3548


iobuhov (approved):
LGTM

samuelreichert (approved):

leonardomendix (approved):

